PR #79, opened 2022-07-07
add git submodule example/mnist-classifier

kengz:


peterschmidt85:
Agreed not to use Git submodules.
Instead, we could organize a repository (e.g. `awesome-dstack`) with links to the content from the community, including examples. I'll provide more information soon!

PR #122, opened 2022-11-22
Fix aws permissions docs

crisp-snakey:
The documentation for the required AWS permissions was under specified/incorrect. This updates the documentation to correct these errors and includes the following fixes:

1. List specific actions for S3 to remove the warning in the AWS console about the S3 permissions being not specific enough. It includes the actions needed to create/read/update/delete objects in S3.
2. Adds the ec2 permission to list subnets.
3. Adds permissions to create and use the created `InstanceProfile` and `Role` also clarify the naming of the created role and instance profile and how dstack manipulates the bucket name.
4. Completes the permissions required to access the `secretsmanager` and create/update/read secrets by the user.
5. Gives the `DescribeLogGroup` action access to the wild card (`*`) resource otherwise it won't work.
6. Adds the `TagLogGroup` permissions that is required to start runs.
7. Groups the `ec2` actions that have a wildcard resource.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -88,7 +88,10 @@ Here's the full list of AWS permissions the `dstack` CLI needs:
         "secretsmanager:TagResource",
         "secretsmanager:DeleteSecret"
       ],
-      "Resource": "arn:*:secretsmanager:*:*:secret:/dstack/{bucket_name}/secrets/*"
+      "Resource": [
+        "arn:aws:secretsmanager:*:*:secret:/dstack/{bucket_name}/credentials/*",
+        "arn:aws:secretsmanager:*:*:secret:/dstack/{bucket_name}/secrets/*"
+      ]
     },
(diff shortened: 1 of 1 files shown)


peterschmidt85:
@crisp-snakey  Thank you a lot for the PR! I merged it, and then also fixed a few things. Please check [#a1cf53a](https://github.com/dstackai/dstack/commit/a1cf53ab6108ad9321b72d552e7460a318001aff).

crisp-snakey:
Oops, seems like I forgot a few things when translating the AWS permissions back into the documentation. Also looks like I forgot the permissions to list secrets.

peterschmidt85:
@crisp-snakey BTW, we just found another bug in the policy. Fixed now: https://github.com/dstackai/dstack/commit/8f9a441920d2e3241f108f2c9778c2f501556bc1

PR #201, opened 2023-02-19
- [CLI]: Show a message if a new version of dstack is available

peterschmidt85:


cli/dstack/cli/updates.py
```diff
@@ -0,0 +1,74 @@
...
+        f"/latest-version"
+        f"?anonymous_installation_id={anonymous_installation_id()}"
+    )
+    try:
+        return requests.get(latest_version_url).text.strip()
+    except Exception:
+        return None
+
+
+def _check_for_updates():
```
r4victor: The name starting with an underscore indicates that the name is private to a module. So `_check_for_updates()` should be `check_for_updates()` and other function names should start with an underscore (e.g. vice versa).

cli/dstack/cli/updates.py
```diff
@@ -0,0 +1,74 @@
...
+
+        def __is_last_check_time_outdated() -> bool:
+            modified_datetime = datetime.fromtimestamp(os.stat(last_check_path).st_mtime)
+            current_datetime = datetime.now()
+            return (
+                modified_datetime.timetuple().tm_yday != current_datetime.timetuple().tm_yday
+                or modified_datetime.year != current_datetime.year
+            )
+
+        def __check_version():
```
r4victor: "Flat is better than nested"

cli/dstack/cli/updates.py
```diff
@@ -0,0 +1,74 @@
...
+from typing import Optional
+
+import requests
+from packaging import version as pkg_version
+
+from dstack import version
+
+
+def anonymous_installation_id() -> str:
+    ai_id_path = Path.home() / ".cache" / ".dstack" / "anonymous_installation_id"
```
r4victor, line 14: Why storing metadata files in `.cache/.dstack` and not in `.dstack`? We already have our place in home dir.

PR #259, opened 2023-03-20
Add private docker registry support

### LINKED ISSUE #246: [Feature] Private Docker registry support
# Problem

Cannot use a private Docker registry to provide images to `dstack`

# Solution

## Implementation notes

Login details could be stored and saved using `dstack secrets`, perhaps.

I imagine it working a little like the below;

```
workflows:
  - name: my-awesome-workflow
    provider: docker
    image: my-awesome-registry.com/my-registry-name:latest
    provider_setup:
      - docker login $DOCKER_REGISTRY_HOST --username $DOCKER_REGISTRY_USERNAME --password 
$DOCKER_REGISTRY_PASSWORD
    setup:
      - pip install my-package
    commands:
      - echo "hooray"
```

# Benefits

This enables the use of private Docker images with secrets and/or proprietary code in `dstack` workflows.

Egor-S:
Closes #246

* Added `registry_auth` for `docker ` provider
* Variables interpolation by the runner
  * `$VAR_NAME`
  * `${VAR_NAME}` and `${ VAR_NAME }`
  * `$$` for escaped `$`
  * Variables come from secrets
  * Warning if a variable is missed
  * Only `registry_auth.username` and `registry_auth.password` at this moment
* Interpolation string validation by the CLI
* Tests for interpolation and validation
* Docs update


docs/usage/private-docker-registry.md
r4victor, line 1: This page fits nowhere in the docs structure. I would suggest to not create a separate page to describe Private Docker Registry. We can describe `registry_auth` property on the docker reference page as we do for other properties.

Screenshot 2023-03-20 at 14 05 16

cli/dstack/core/interpolation.py
r4victor, line 1: `interpolation.py` doesn't feel like it belongs to `core` directory because all other modules there are centered around data models. I'd move the interpolation logic to `job.py` or some other model file or to a separate file in `utils`.

r4victor (dismissed):

cli/dstack/core/interpolation.py
Egor-S: Will move to utils

Egor-S:
Reworked interpolation syntax:

* `${{ secrets.VAR_NAME }}` used for secrets
  * `$$` escapes `$`, but it is really needed only before `{{`
* `JinjaMock` prevents client-side interpolation
* Updated tests
* Client-side validation is handled by jinja2 template rendering
* Fixed docs

cli/dstack/utils/jinjamock.py
```diff
@@ -0,0 +1,27 @@
...
+    Example:
+        value = "${{ secrets.any_key }}"
+        assert value == Template(value, variable_start_string="${{").render(secrets=JinjaMock("secrets"))
+    """
+
+    def __init__(self, namespace: str):
+        self.__n__ = namespace
+
+    def __getattr__(self, item: str) -> JinjaMockValue:
+        namespace = self.__dict__["__n__"]
```
r4victor: Seem like some weird magic. Why can't we just store `self.namespace = namespace`? Why access via self.__dict__? Why return `JinjaMockValue`? Can't we just return a string? Why name variables like __s__ – i.e. double underscores on both sides usually means it has special meaning to the interpreter.

Not critical, just reads weird.

r4victor (approved):

cli/dstack/utils/jinjamock.py
Egor-S: I tried to invent something unlikely to be used as a secret name

`${{ secrets.namespace }}` is more likely than `${{ secrets.__n__ }}` in my opinion, but brokes `JinjaMock`

`JinjaMockValue` is designed to raise errors in indexing and chaining. `str` would allow indexing and break `JinjaMock`: `${{ secrets.password[1] }}` and `${{ secrets.password.hash }}`

Probably I could avoid using `__dict__` there

Egor-S: `JinjaMock` is removed

cli/dstack/utils/interpolator.py
```diff
@@ -0,0 +1,68 @@
...
+        start = 0
+        while start < len(s):
+            dollar = s.find("$", start)
+            if dollar == -1 or dollar == len(s) - 1:
+                tokens.append(s[start:])
+                break
+            if s[dollar + 1] == "$":  # escaped $$
+                tokens.append("$")
+                start += 2
+                continue
```
r4victor: Seems like the code doesn't handle escaped $$ in the middle of the string. E.g. `self.assertEqual("echo ${{ENV}}", get_interpolator().interpolate("echo $${{ENV}}"))` fails.

PR #260, opened 2023-03-20
Execute background commands

### LINKED ISSUE #257: [Bug]: Doesn't allow running bash commands in background
### dstack version

0.2

### Python version

3.9.1

### Host OS

macOS 13.1

### Host Arch

ARM

### What happened?

I want to run several bash commands inside my workflow. One of the commands must be run in background.

Here's an example:

```yaml
workflows:
  - name: test
    provider: bash
    commands:
      - echo 1 &
      - echo 2
```

The `echo 1` command must run background.

However, when I run the workflow via `dstack`, it fails:

```
bash: syntax error near unexpected token `&&'
```

### Relevant log output

_No response_

### Additional Information

Background commands can be quite useful, for example, if we want to run `tensorboard` in parallel while the workflow is running.

Egor-S:
Closes #257

* Allow background commands
  * Wraps command ending with `&` in curly brackets (`sleep 5 & && echo 123` --> `{ sleep 5 & } && echo 123`)
* Add tests for `engine.ShellCommands`

### DIFF
--- a/docs/reference/providers/bash.md
+++ b/docs/reference/providers/bash.md
@@ -110,3 +110,20 @@ workflows:
 ```
 
 </div>
+
+### Run task in a background
+
+<div editor-title=".dstack/workflows/ping-background.yaml">
+
+```yaml
+workflows:
+  - name: ping-background
+    provider: bash
+    commands:
+      - apt update && apt -y install iputils-ping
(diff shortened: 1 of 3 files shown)


peterschmidt85 (approved):
LGTM

PR #297, opened 2023-04-03
Workflow cache

### LINKED ISSUE #284: [Feature] Allow to specify paths to the folders that must be cached (for the current user, repo and workflow)

Egor-S:
Closes #284 

* Add workflow option `cache`
* If cache exists, it would be copied from storage before the run
* Cache is saved after the run to storage
* Cache is unique per-repo & per-user & per-workflow
  * Location `cache/{repo/address}/{user_email}/{workflow_name}/*`
* Add `dstack prune cache` command
  * Add `Backend.delete_workflow_cache`
  * Add `Storage.delete_prefix` to delete trees

Limitations:

* `delete_prefix` is not implemented for hub backend
* `delete_prefix` could be slow for large trees (aws, gcp) — need to use batch operations

Example workflow:

```yaml
workflows:
  - name: cache-demo
    provider: bash
    commands:
      - "du -hd2 /root/.cache"
      - "pip install numpy pandas matplotlib"
      - "du -hd2 /root/.cache"
    cache:
      - "/root/.cache/pip"
```

**Edit:**

Allowed syntax:

```yaml
cache:
  - "/my/cache/foo"
  - path: "my/cache/bar"
```


peterschmidt85 (approved):
Let's add support for Hub and merge. Okay?

peterschmidt85:
For `bash`, `code`, `lab`, and `notebook` providers, let's allow the CLI to automatically replace the `~/` prefix with `/root` (the same place where we remove the `./` prefix). To avoid hardcoding, we can allow the provider to specify whether it is aware of the user's home directory path.
For all others, let's show an error telling that this is not allowed.
This will make defining paths inside the user's home directory much easier.

Egor-S:
* `hub` backend supports cache pruning
* `aws` backend works correctly with a cache
* `artifacts` and `cache` support `~` (home directory) expansion for dstack-managed providers

peterschmidt85 (approved):

PR #307, opened 2023-04-10
Refactor backend repo

Egor-S:
1. Store a diff filename in a job file instead of a diff itself
2. Remove unnecessary `__str__` formatters for models
3. Add `Repo` model (as preparation for git-less mode)
4. Use custom `Repo.name` instead of `repo_address` as a key
5. Use custom `Repo.username` instead of `local_repo_user_name` and `local_repo_user_email`
6. Lazy loading for `repo_user_config` via `ConfigManager` with caching
7. Store `Repo` in backend and secrets manager instances
8. Remove `repo_address` from the majority of methods
9. Replace internal `repo_address` with `repo_name`

ToDo

* [x] Test with AWS
* [x] Test with GCP
* [x] Test with Hub


cli/dstack/backend/gcp/__init__.py
```diff
@@ -33,8 +33,16 @@
 
 
 class GCPBackend(CloudBackend):
-    def __init__(self, repo: Optional[Repo], backend_config: Optional[GCPConfig] = None):
-        super().__init__(backend_config=backend_config, repo=repo)
```
peterschmidt85: I'd change the order: first backend_config, and then repo.

cli/dstack/backend/gcp/__init__.py
```diff
@@ -33,8 +33,16 @@
...
+    def __init__(
+        self,
+        repo: Optional[Repo],
+        backend_config: Optional[GCPConfig] = None,
+        credentials: Optional[RepoCredentials] = None,
+        auto_init: bool = False,
```
peterschmidt85, line 41: IMO, I'd set auto_init by default to true. Easier to use. If the user knows what to di, then the user would set it to false.

cli/dstack/core/dependents.py
```diff
@@ -1,16 +1,7 @@
 from pydantic import BaseModel
 
-from dstack.core.repo import RepoAddress
-
 
 class DepSpec(BaseModel):
-    repo_address: RepoAddress
+    repo_id: str
```
peterschmidt85: Why not RepoRef?

cli/dstack/core/repo.py
```diff
@@ -57,92 +19,155 @@ class RepoProtocol(Enum):
 
 class RepoCredentials(BaseModel):
```
peterschmidt85: I'd rename to RemoteRepoCredentials

cli/dstack/backend/gcp/__init__.py
Egor-S: I suggest leaving it as it is now to keep the behavior. We could change it with the SDK release

cli/dstack/core/repo.py
```diff
@@ -57,92 +19,155 @@ class RepoProtocol(Enum):
...
+class RemoteRepoData(RepoData):
+    repo_type: Literal["remote"] = "remote"
+    repo_host_name: str
+    repo_port: Optional[int]
+    repo_user_name: str
+    repo_name: str
+    repo_branch: Optional[str] = None
+    repo_hash: Optional[str] = None
+    repo_diff: Optional[str] = None
+    repo_protocol: Optional[str] = None
```
peterschmidt85: I've thought repo_protocol is only used as a part of credentials.

Egor-S: Without `repo_protocol` we can't create credentials from `RemoteRepoData`. One of the constructors for `RemoteRepo` accepts only `RemoteRepoData` — in this case, there is no git URL or local repo to find the protocol

cli/dstack/backend/base/cache.py
```diff
@@ -1,16 +1,13 @@
 import os.path
 
 from dstack.backend.base.storage import Storage
-from dstack.core.repo import RepoAddress
 
 
-def delete_workflow_cache(
-    storage: Storage, repo_address: RepoAddress, username: str, workflow: str
-):
+def delete_workflow_cache(storage: Storage, repo_name, username: str, workflow: str):
```
r4victor: Should be `repo_id` instead of `repo_name`?

cli/dstack/cli/config.py
```diff
@@ -1,21 +1,44 @@
...
+        """
+        repo_dir = Path.cwd() if repo_dir is None else Path(repo_dir).resolve()
+        return self.repos / f"{'.'.join(repo_dir.parts[1:])}.yaml"
+
+    @property
+    def repo_user_config(self) -> RepoUserConfig:
+        try:
+            return self._cached_read(self.repo_user_config_path(), RepoUserConfig)
+        except FileNotFoundError:
+            raise InvalidGitRepositoryError()
```
r4victor: Should define our own error instead of raising an error from a library.

cli/dstack/cli/commands/init/__init__.py
```diff
@@ -56,25 +59,29 @@ def register(self):
...
+        )
+        test_repo_credentials(repo.repo_data, repo_credentials)
 
-        config = BaseConfig()
-        repo_user_config = RepoUserConfig(ssh_key_path=get_ssh_keypair(args.ssh_identity_file))
-        config.write(
-            config.repos / f"{local_repo_data.path(delimiter='.')}.yaml",
-            repo_user_config,
-            mkdir=True,
+        config.repo_user_config = RepoUserConfig(
```
r4victor: I'd not implement writing to a file via setter – kinda magic

cli/dstack/api/repos.py
```diff
@@ -0,0 +1,80 @@
...
+    if os.path.exists(gh_config_path):
+        with open(gh_config_path, "r") as f:
+            gh_hosts = yaml.load(f, Loader=yaml.FullLoader)
+        oauth_token = gh_hosts.get(hostname, {}).get("oauth_token")
+        if oauth_token:
+            return RemoteRepoCredentials(
+                protocol=RepoProtocol.HTTPS, oauth_token=oauth_token, private_key=None
+            )
+    # default user key
+    if os.path.exists(default_ssh_key):
```
r4victor, line 64: If I work with a https public repo, I don't need ssh key, but this line makes it picked up by default. init won't work if my ssh key is not set up in github. I suggest we don't require it for public repos.

PR #333, opened 2023-04-18
Get rid of sudo in tests

r4victor:


r4victor:
Requires more consideration: #335

PR #380, opened 2023-05-04
Optimize cache synchronization for GCP and AWS

### LINKED ISSUE #330: [Feature] When uploading large files (e.g. larger than 1MB), skip if they were uploaded already (e.g. previously)
Motivation:

When we use cache files, we upload them once the job is finished or stopped, even if they didn't change. If we cache pre-trained models from HuggingFace, these files are always pretty big (e.g. 7GB and more), and we have to wait significant additional time on each run.

Egor-S:
Closes #330 

* Set mtime with a value from GCS or S3
* Do not upload if both size and modification time haven't changed
* Delete object if the local file doesn't exist anymore
* Add `GetCache` to backend interface

Further work:
* ~~Add support for AWS and Local backends~~
* Optimize memory usage for large trees


Egor-S:
* Set mtime from AWS S3
* Cloud agnostic sync upload implementation

PR #383, opened 2023-05-06
Refactor assets file structure

kamilkrzyskow:
Hi, 👋 
the assets were scattered around, so I standardized it all into one directory. Additionally, I changed files with a underscore `_` to use a dash `-` instead, and I made some other minor name adjustments. There are some files that are redundant, aren't used anywhere like `cats-in-hats.png` or `new.svg`, there is also a missing file `dstack-stable-diffusion-code.png`. 

The structure could be improved more, for example splitting assets into more subdirectories, also I'm not sure if adding `dstack` to every "in-house" image is the best approach. 

In our docs we keep every static / shared across languages asset in the `overrides/assets` directory. This breaks IDE relative path lookup inside the `docs` directory, but allows to keep everything inside one `assets` directory. The relative paths work after the site build.

Anyways, things that I noticed when refactoring:
- The custom css makes it quite hard to read the left navigation using Firefox, there is some sort of white overlay over the text.
- The website styling crashes on localhost, when using Chromium, some typescript error not being able to get some palette property. It didn't happen in Firefox, and the online version works with Chromium 🤷.

I don't like to work with CSS, so I didn't investigate further. The issues above also occurred before my changes ✌️ 

I also noticed that after my last PR the whole https://github.com/dstackai/dstack/actions/runs/4900569673 action run. 6,5 minutes seems rather wasteful when I only changed a line in the docs. I believe you can limit the trigger to include/exclude certain file paths.

---

I see an `svg` wrapper over `png` for the first time, I doubt I will use that new knowledge, but thanks for the new experience. 😵


kamilkrzyskow:
Looking at the errors
https://github.com/dstackai/dstack/actions/runs/4901189610/jobs/8752338871?pr=383#step:7:1200
https://github.com/dstackai/dstack/actions/runs/4901189610/jobs/8752332789?pr=383#step:5:11
this doesn't seem to be related to my changes 🤔 I only changed the `README.md` file to use different links to assets. They are currently 404 since the site isn't updated, but the errors ask for credentials.

Also another case of CI redundancy, this run will basically be duplicated after the merge.

peterschmidt85:
@kamilkrzyskow Thank you so much for refactoring.

Here are the action points for the future:

- [ ] Fix the bugs in Firefox and Chromium.
- [ ] Ensure that certain GitHub Actions jobs only run if there have been changes in the corresponding source roots.
- [ ] Do not run certain GitHub Actions jobs on pull requests from external repos as the required secrets are not configured there

PR #410, opened 2023-05-17
Let users to define ports

### LINKED ISSUE #400: [Usability]: Let users to set apps' ports explicitly
### Problem

To access any application inside the container you have to:

1. Query the number of ports
2. Use a value of the env variable `PORT_N` to configure your app

It's inconvenient 

### Solution

Define ports instead of their count. Do not choose a port number randomly.

### Benefit

A more intuitive approach to apps configuration

### Alternatives

Detect all listened ports inside the container and forward them

### Would you like to help contributing this feature?

Yes

Egor-S:
Closes #400 

* Add `ports: list[int]` to all providers schema
* Add `-p/--ports int,int,int` to all providers args parser
* Serialize ports as part of app specs
* Reserve ports range `2000-2999` for dstack needs (code, lab, notebook, ssh server, ws logs)
* Remove PortsManager from the runner
* Use the closest port for local mapping
* Implement OpenSSH server via ProviderExtension interface

TODO
* [x] Implement flexible ports remapping for local runner (docker without host mode)
* [x] Let user to define custom ports mapping
* [x] Test with remote backends


Egor-S:
* `--port` in args and `ports:` in schema support mappings `listen:map_to`
* Disable `host` mode for local backend
* Implement flexible mapping for runner and cli
  * Fail if user-defined mapping couldn't be satisfied
  * Try to use the same port if possible
  * Choose the closest free port otherwise

docs/docs/reference/providers/bash.md
```diff
@@ -98,20 +98,24 @@ The list of directories to cache between runs
...
 
-If you'd like your workflow to expose ports, you have to specify the `ports` property with the number
-of ports to expose. Actual ports will be assigned on startup and passed to the workflow via the environment
-variables `PORT_<number>`.
+If you'd like your workflow to expose ports, you have to specify the `ports` property with the list
+of ports to expose. You could specify a mapping `APP_PORT:LOCAL_PORT` or just `APP_PORT` — in this
+case dstack will choose available `LOCAL_PORT` for you.
+
+!!! info "NOTE:"
+    Ports range `2000–2999` is reserved for dstack needs. However, you could remap them to different `LOCAL_PORT`s.
```
peterschmidt85: 1. Do we show an error if the user chooses a port within `2000-2999`?
2. Is it difficult to use `10000-10999` (to reduce the chances of an overlap)?

cli/dstack/core/job.py
```diff
@@ -352,7 +349,8 @@ def unserialize(job_data: dict):
         app_specs = (
             [
                 AppSpec(
-                    port_index=a["port_index"],
+                    port=a["port"],
```
peterschmidt85: Will it with nicely with the old runs (serialized differently?)

docs/docs/reference/providers/bash.md
Egor-S: 1. I don't see how we could reliably differentiate remapping and port requests. For example, I want to run code on a different port: `-p 2001:3001`
2. I believe we can do that

cli/dstack/core/job.py
Egor-S: I can make it safer, but job resubmitting with ports would not be successful

PR #437, opened 2023-05-26
Sync local changes with remote run

### LINKED ISSUE #402: [Usability] Allow to sync local changes with the running workflow

Egor-S:
Closes #402 

* Add ~~`--sync`~~ `--reload` option for dstack run
* Watch local changes with `watchfiles` in a separate thread
  * Collect changes in the queue from the beginning
* Apply changes with `Copier` in a separate thread
  * `LocalCopier` for local backend, using `shutil` and `pathlib`
  * `SSHCopier` for remote backends, using `subprocess` with ssh and scp to the host
* Do not sync gitignored files and directories

todo
* [x] Prevent copying conflict for directory and file
* [x] Test with hot reloading app


peterschmidt85:
I suggest that we call the option `--reload`, this is also how it's named in both Gradio and FastAPI.

PR #459, opened 2023-06-07
Show run plan before run execution

### LINKED ISSUE #339: [Usability] Make the `dstack run` command show the plan and prompt the user to confirm
1. The `dstack run` command should show the jobs (incl. the target project and instance names) the run is going to create and ask the user to confirm.
2. The user should be able to use the `--yes` or `-y` flag to auto-confirm the plan.

Motivation:

- Often, the user wants to see which instance types the run will use before the run is actually submitted. This reduces the risk of using expensive instances wrongly.

Important notes:

- For now, it's not important that the actual execution will be 100% the same as the shown plan. For now, it's a lot more important that the user may see the project and the instance type that is going to be used.

r4victor:
Closes #339


peterschmidt85:
Does it also fix the issue for GCP/Azure that the CLI doesn't show the progress for long time (while silently provisioning the run)?

r4victor:
@peterschmidt85, it does.

PR #465, opened 2023-06-08
Support T4 (`g4dn.*`) A10G (`g5.*`) gpu types for AWS

### LINKED ISSUE #445: [Feature] Support T4 (`g4dn.*`) A10G (`g5.*`) gpu types for AWS
Most importantly, check that the CUDA driver works there well.

peterschmidt85:


peterschmidt85:
@r4victor,
Now that we add support for `T4` and `A10G` for AWS, shall we also support `T4` (and possibly `L4` - need to check if it works with CUDA 11.4) for GCP? Shall I create a separate issue?

PR #470, opened 2023-06-09
Auto-detect VS Code version for dev environment

Egor-S:
* Rely on `code` in the PATH
* Do not install `vscode-server` if can't find `code` binary

### DIFF
--- a/cli/dstack/_internal/cli/commands/run/configurations.py
+++ b/cli/dstack/_internal/cli/commands/run/configurations.py
@@ -38,9 +38,9 @@ def _parse_dev_environment_configuration_data(
                 "ms-toolsai.jupyter",
             ],
         )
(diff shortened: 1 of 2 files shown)


cli/dstack/_internal/cli/commands/run/configurations.py
```diff
@@ -29,14 +30,17 @@ def _parse_dev_environment_configuration_data(
...
-    provider_data["setup"].append("pip install -q --no-cache-dir ipykernel")
+    try:
+        VSCodeDesktopServer.patch_setup(
+            provider_data["setup"],
+            vscode_extensions=[
+                "ms-python.python",
+                "ms-toolsai.jupyter",
+            ],
+        )
+        provider_data["setup"].append("pip install -q --no-cache-dir ipykernel")
```
peterschmidt85: `ipykernel` should be still installed even if VS Code is not installed

PR #477, opened 2023-06-14
Export/import prebuild image via overlay2 internals

### LINKED ISSUE #468: [Bug]: Prebuild takes too long in cloud environment
### dstack version

master

### Python version

any

### Host OS

macOS 13.4

### Host Arch

ARM

### What happened?

`docker save dstackai/miniforge` takes an average of 4 minutes for the first run on AWS. It is probably caused by lazy AMI loading (EBS has to fetch data from underlying S3). The current implementation relies on `save` in prebuild and use stages.

### Relevant log output

_No response_

### Additional Information

Possible solutions:

1. Read and write docker cache directly on Linux, rely on local docker cache on Mac OS and Windows
2. Use a cloud-provided registry

Egor-S:
Closes #468 

* On the local backend use daemon's cache for storing prebuild images
* On cloud backends (i.e. linux-only) export the latest diff layer from the internal overlay2 structure
  * Archive as tar
  * Store in cloud storage

TODO

* [x] Test with cloud backends
* [x] Run benchmarks
* [ ] Consider prebuilding with cheaper instances


Egor-S:
Side improvement: manual AMI build now use the runner version from the same branch

PR #515, opened 2023-06-22
Updated fields in run details page

### LINKED ISSUE #505: Update run page UI for new configuration format
After moving from workflow/provider concept to run configurations, we updated the run list UI, but not the run details page. We should update it as well. The page should include the same fields present in the list UI and some additional fields. More specifically: Configuration, User, Instance, Spot, Status, Submitted, Error.

olgenn:
Implemented: The run details page include same fields as the run list page


olgenn:
Error field is visible if server response will have error_code field and that hasn't null

PR #518, opened 2023-06-26
Warn user about long git diff

### LINKED ISSUE #456: Show a warning if the Git diff is taking more than X seconds

Egor-S:
Closes #456 

* Warn user if git diff takes longer than 5 seconds
  * Suggest updating .gitignore
* Gracefully exit from git diff on Ctrl+C
* Do not serialize repo_diff
  * It accelerates get_run_plan for large diffs

### DIFF
--- a/cli/dstack/_internal/core/repo/remote.py
+++ b/cli/dstack/_internal/core/repo/remote.py
@@ -1,11 +1,14 @@
+import io
 import os
+import subprocess
 import tempfile
+import time
 from pathlib import Path
 from typing import Any, BinaryIO, Dict, Optional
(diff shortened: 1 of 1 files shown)


Egor-S:
Ctrl+C may be ignored during subprocess deletion. However, it is not very likely and doesn't affect the consistency

```
Exception ignored in: <function Git.AutoInterrupt.__del__ at 0x10e8988c0>
  File "/Users/egors/dstack/envs/p37/lib/python3.7/site-packages/git/cmd.py", line 567, in __del__
    self._terminate()
```

Issue #672, opened 2023-08-23
Configure gateways from project settings page


Egor-S:
Core features:

1. Show the list of existing gateways in the project
2. Delete gateway
3. Create a gateway
4. Edit gateway
    - Assign wildcard domain
    - Mark as default

Creation form:
- `backend` - select, required
- `region` - select, null value allowed

Update form:
- `default` - checkbox, null value allowed
- `wildcard_domain` - text, null value allowed

peterschmidt85:
Also, allow to pre-select a region

peterschmidt85:
Also, when creating a gateway, should be possible to select a primary backend

PR #676, opened 2023-08-25
Issue 672 gateway configure api

olgenn:


Egor-S:
Closes #672

- Add default project gateway
- Add UI for configuring gateways
- Add gateway region and wildcard domain
- Add auto-generated subdomains for service jobs
- Remove `gateway` option from configuration
- Service job can use only the default gateway
- Update docs

PR #748, opened 2023-10-29
Multiple attachment improvements #747

### LINKED ISSUE #747: Multiple attachment improvements
- [ ] Allow to invoke `run.attach` on a run obtained via `client.runs.get` and have attached logs
- [ ] Support `-a` (`--attach`) in `dstack logs`
- [ ] Automatically detach from the run on the program's exit
- [ ] Allow to invoke `run.attach` without errors when the tunnel is already established by another process

peterschmidt85:
- Allow to invoke `run.attach` without errors when the tunnel is already established by another process
- Allow to invoke `run.attach` on a run obtained via `client.runs.get` and have attached logs
- Support `-a` (`--attach`) in `dstack logs`
- Automatically detach from the run on the program's exit

### DIFF
--- a/src/dstack/_internal/cli/commands/logs.py
+++ b/src/dstack/_internal/cli/commands/logs.py
@@ -1,9 +1,10 @@
 import argparse
 import sys
+from pathlib import Path
 
 from dstack._internal.cli.commands import APIBaseCommand
-from dstack._internal.cli.utils.common import confirm_ask
 from dstack._internal.core.errors import CLIError
+from dstack._internal.core.services.ssh.ports import PortUsedError
 
 
 class LogsCommand(APIBaseCommand):
@@ -13,6 +14,19 @@ class LogsCommand(APIBaseCommand):
     def _register(self):
         super()._register()
         self._parser.add_argument("-d", "--diagnose", action="store_true")
+        self._parser.add_argument(
+            "-a",
+            "--attach",
+            action="store_true",
+            help="Set up an SSH tunnel, and print logs as they follow.",
+        )
+        self._parser.add_argument(
+            "--ssh-identity",
+            metavar="SSH_PRIVATE_KEY",
+            help="A path to the private SSH key file for SSH tunneling",
+            type=Path,
+            dest="ssh_identity_file",
+        )
         self._parser.add_argument("run_name")
 
     def _command(self, args: argparse.Namespace):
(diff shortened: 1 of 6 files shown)


src/dstack/_internal/cli/commands/logs.py
```diff
@@ -13,14 +14,38 @@ class LogsCommand(APIBaseCommand):
...
         super()._command(args)
         run = self.api.runs.get(args.run_name)
         if run is None:
             raise CLIError(f"Run {args.run_name} not found")
+        if not args.diagnose and args.attach:
+            if run.status.is_finished():
+                raise CLIError(f"Run {args.run_name} is finished")
+            else:
+                try:
+                    run.attach(args.ssh_identity_file)
```
Egor-S: We should use an ssh key from repo config by default

src/dstack/_internal/cli/commands/logs.py
```diff
@@ -13,14 +14,38 @@ class LogsCommand(APIBaseCommand):
...
+                except PortUsedError:
+                    pass
```
Egor-S: If the exception was raised, PortsLock won't assign any ports automatically (`{10999: 0}`)

src/dstack/_internal/core/services/ssh/tunnel.py
```diff
@@ -134,12 +134,20 @@ class ClientTunnel(SSHTunnel):
...
-    def __init__(self, host: str, ports: Dict[int, int], id_rsa_path: PathLike):
-        self.temp_dir = tempfile.TemporaryDirectory()
+    def __init__(
+        self,
+        host: str,
+        ports: Dict[int, int],
+        id_rsa_path: PathLike,
+        control_sock_path: Optional[str] = None,
+    ):
+        self.temp_dir = tempfile.TemporaryDirectory() if not control_sock_path else None
```
Egor-S, line 144: I wouldn't mix dir and path in a single variable

src/dstack/_internal/core/services/ssh/attach.py
```diff
@@ -1,14 +1,38 @@
...
 from dstack._internal.core.services.ssh.ports import PortsLock
 from dstack._internal.core.services.ssh.tunnel import ClientTunnel
 from dstack._internal.utils.path import PathLike
-from dstack._internal.utils.ssh import include_ssh_config, update_ssh_config
+from dstack._internal.utils.ssh import get_ssh_config, include_ssh_config, update_ssh_config
 
 
 class SSHAttach:
+    @staticmethod
+    def reuse_control_sock_path_and_port_locks(run_name: str) -> Optional[Tuple[str, PortsLock]]:
```
Egor-S, line 17: I see many edge cases in this approach.

What happens if we reuse the connection and exit early? I suppose we will close the tunnel with atexit and the original process will fail.

We could reuse the control path socket (for example, under `~/.dstack/sock/<job-name>`), but forward only a single port to the runner logs to a free local port.

PR #804, opened 2023-12-17
Allow to configure disk volume and type via resources

### LINKED ISSUE #135: Allow to configure disk volume  and type via `resources`
Currently, `dstack` always uses a  `100GB` `gp2` disk. 

The user should be able to change the defaults via the `resources` property of the YAML file.

Also, this must be reflected in the documentation (incl. the default).

peterschmidt85:
Initial support for disk size.

Now `Resources` has `disk` with `size_mib`. It's set based on `CatalogItem.disk_size` if set or `Requirements.resources.disk.size` otherwise.

Limitations:
 - If `CatalogItem.disk_size` is not set, `price` doesn't include the disk price.


src/dstack/_internal/core/models/runs.py
```diff
@@ -87,12 +92,18 @@ def pretty_format(self, resources_only: bool = False):
...
+            resources.update(
+                disk_size=self.disk.size_mib,
+            )
         res = pretty_resources(**resources)
         if not resources_only:
             if self.spot is not None:
                 res += f", {'spot' if self.spot else 'on-demand'}"
             if self.max_price is not None:
                 res += f" under ${self.max_price:g} per hour"
+            if self.disk is not None:
```
Egor-S: `pretty_resources` is also adding the disk size. Do we need to do that here?

src/dstack/_internal/core/backends/aws/compute.py
```diff
@@ -133,9 +133,14 @@ def run_job(
             {"Key": "dstack_user", "Value": run.user},
         ]
         try:
+            disk_size_mib = (
+                instance_offer.instance.resources.disk.size_mib
+                or job.job_spec.requirements.disk.size_mib
+            )
+            disk_size = round(disk_size_mib / 1024) if disk_size_mib else 100
```
Egor-S: AWS will fail if disk size is less than AMI size (100GB for now)

Why do we add a default value here and not in Profile, similar to CPU and RAM?

Egor-S (approved):

PR #829, opened 2024-01-09
Make gateway active

### LINKED ISSUE #799: Make gateway active
Currently, the job connects to the gateway with a reverse SSH tunnel. This approach makes it harder to detect service failure, on gateway failure all connected services become irreversibly unavailable.

This issue suggests establishing an SSH tunnel from the gateway.

Egor-S:
This PR changes the communication scheme between the server, job instance, and gateway. 

Closes #799 

* The gateway now runs the FastAPI app as a systemd service
* The gateway now establishes an SSH tunnel to the job instance (contrary to the previous behavior)
* The server registers and unregisters a service via secured HTTP requests to the gateway
* The gateway features an OpenAI interface
    * OpenAI requests are converted to TGI format and sent to the job instance
    * OpenAI interface supports /models query (scoped by the project)
* If run from source, the server picks the latest build version related to the HEAD

Steps to use an OpenAI interface:

1. Create a new default gateway
2. Redirect the DNS record to the gateway
3. Add the following to the service configuration (TGI)
    ```
    model:
      type: chat
      name: TheBloke/Mistral-7B-OpenOrca-AWQ
      format: tgi
      # optionally override `chat_template:` and `eos_token:` here
    ```
4. Run the service using `dstack run`
5. Make a request to `gateway.<your domain>` using `openai` python package

TODO before merge:
* [x] Update gateway app on server start up
* [x] Build gateway app in release workflow


src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -193,6 +209,48 @@ async def _process_job(job_id: UUID):
...
+            and job_model.status == JobStatus.RUNNING
+            and run.run_spec.configuration.type == "service"
+        ):
+            res = await session.execute(
+                select(GatewayModel).where(
+                    GatewayModel.project_id == project.id,
+                    GatewayModel.name == job.job_spec.gateway.gateway_name,
+                )
+            )
+            gateway = res.scalar_one()
```
r4victor: I think GatewayModel can be deleted while the service is running and in this case res.scalar_one() will throw an error. We should either disallow deleting gateways with running services or handle the case when the gateway was deleted (e.g. fail the service).

src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -254,26 +260,70 @@ async def register_service_jobs(
...
+):
+    try:
+        with gateway_tunnel_client(domain, gateway_ssh_private_key) as client:
+            client.preflight(project, domain, project_ssh_private_key, options)
+    except requests.RequestException:
+        raise GatewayError(f"Gateway is not working")
+
+
+async def update_gateways(session: AsyncSession):
+    if os.environ.get("DSTACK_SKIP_GATEWAY_UPDATE", None):
```
r4victor: Put DSTACK_SKIP_GATEWAY_UPDATE in server settings so that envs are not scattered around.

PR #852, opened 2024-01-27
error in documentation: use --url instead of --server

promsoft:
There must be error in documentation: no such option `--server`
```
$dstack --version
0.14.0

$dstack config --help
Usage: dstack config [-h] [--project PROJECT] [--url URL] [--token TOKEN]
                     [--default] [--remove] [--no-default]

Options:
  -h, --help         Show this help message and exit
  --project PROJECT  The name of the project to configure
  --url URL          Server url
  --token TOKEN      User token
  --default          Set the project as default. It will be used when
                     --project is omitted in commands.
  --remove           Delete project configuration
  --no-default       Do not prompt to set the project as default
```

### DIFF
--- a/docs/docs/installation/index.md
+++ b/docs/docs/installation/index.md
@@ -423,7 +423,7 @@ To configure the client on a different machine or for other projects, use [`dsta
 <div class="termy">
 
 ```shell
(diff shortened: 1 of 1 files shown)


peterschmidt85:
@promsoft Thank you for the fix! ☺️

PR #922, opened 2024-02-22
Fix #921

### LINKED ISSUE #921: Fix instance statuses when using the vast.ai
The status is always in `starting` state.

TheBits:
* Fix instance status
* Reorder columns like dstack ps
* Stop vastai/k8s instance immediately
* Skip vastai/k8s offers in dstack pool


src/dstack/_internal/cli/commands/pool.py
```diff
@@ -360,8 +363,8 @@ def print_instance_table(instances: Sequence[Instance]) -> Table:
             instance.region,
             instance.instance_type.resources.pretty_format(),
             "yes" if instance.instance_type.resources.spot else "no",
-            f"[{style}]{instance.status.value}[/]",
             f"${instance.price:.4}",
+            f"[{style}]{instance.status.value}[/]",
```
Egor-S: I remember, we wanted to remove any styling for the status

src/dstack/_internal/core/models/instances.py
```diff
@@ -85,6 +85,11 @@ class LaunchedInstanceInfo(BaseModel):
     backend_data: Optional[str] = Field(default=None)  # backend-specific data in json
 
 
+class InstanceRuntime(Enum):
```
Egor-S, line 88: It would be great if we could someday replace `dockerized` with this enum

src/dstack/_internal/server/background/tasks/process_finished_jobs.py
```diff
@@ -89,30 +89,41 @@ async def _process_job(job_id):
                 except Exception as e:
                     logger.warning("failed to unregister service: %s", e)
 
+        job_provisioning_data = None
+        if job_model.job_provisioning_data is not None:
+            job_provisioning_data = parse_raw_as(
```
Egor-S, line 94: Note, that for plain types it's simpler to use `BaseModel.parse_raw`

src/dstack/_internal/cli/commands/pool.py
TheBits: Fixed

Issue #933, opened 2024-02-26
[Bug]: The SSH tunnel fails because of a messy `~/.ssh/config`


peterschmidt85:
Steps to reproduce:

1. Have a messy `~/.ssh/config` (containing some problematic settings)
2. Run `dstack server`
3. Define `.dstack.yml`
4. Run `dstack run .`

The server fails to establish the SSH tunnel:

```
DEBUG 2024-02-23T08:59:20.314 dstack._internal.core.services.ssh.tunnel SSH tunnel failed: b'ssh: connect to host california-a.tensordockmarketplace.com port 20499: Connection refused\r\n'
```

A possible reason:

After a long debugging session with @spott, we figured out that the problem goes away if the server adds `-F <home directory>/.dstack/ssh/config` to all the `ssh` commands aiming to establish a connection. This solution appears to resolve the issue because, in that case, `dstack` doesn't utilize the user's default SSH config, where some parts may cause the issue.

spott:
[here](https://github.com/dstackai/dstack/pull/934) is a pull request to implement some of these changes.

peterschmidt85:
Gonna be fixed in 0.16.1

Issue #927, opened 2024-02-23
[Bug]: The SSH tunnel fails with `Too many authentication failures`


spott:
Steps to reproduce:

1. Have multiple identity files under `~/.ssh` (e.g. more than 30)
2. Run `dstack server`
3. Define `.dstack.yml`
4. Run `dstack run .`

The server fails with 

```
Warning: Permanently added '192.18.137.190' (ED25519) to the list of known hosts.\r\nReceived disconnect from 192.18.137.190 port 22:2: Too many authentication failures\r\nDisconnected from 192.18.137.190 port 22\r\n
```

A possible reason:

The server, when [establishing an SSH tunnel](https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/core/services/ssh/tunnel.py) doesn't specify the identity file. Because of that, `ssh` tries all identity files in sequence [resulting in the error](https://superuser.com/questions/268776/how-do-i-configure-ssh-so-it-doesnt-try-all-the-identity-files-automatically).

peterschmidt85:
Gonna be fixed in 0.16.1

PR #934, opened 2024-02-26
Fix the SSH issues related to the default SSH config and too many identity files

spott:
This PR addresses the following issues:

- #927
- #933
- #937

Implementation notes:

1. For issue 927, use `IdentitiesOnly` in the client SSH config
2. For issue 933, use `-F <client dstack SSH config path>` on the client, and `-F none` in other places when invoking the `ssh` command to exclude the default SSH config settings
3. For issue 937, display a clear error message

### DIFF
--- a/gateway/src/dstack/gateway/services/tunnel.py
+++ b/gateway/src/dstack/gateway/services/tunnel.py
@@ -8,12 +8,9 @@ from typing import List, Optional
 
 from pydantic import BaseModel
 
-from dstack._internal.core.services.configs import ConfigManager
 from dstack.gateway.errors import SSHError
 
 logger = logging.getLogger(__name__)
-
-config = ConfigManager()
 
 
 class SSHTunnel(BaseModel):
@@ -45,7 +42,7 @@ class SSHTunnel(BaseModel):
         cmd = [
             "ssh",
             "-F",
-            str(config.dstack_ssh_config_path),
+            "none",
             "-i",
             id_rsa_path,
             "-M",
@@ -61,7 +58,7 @@ class SSHTunnel(BaseModel):
             proxy = [
                 "ssh",
                 "-F",
-                str(config.dstack_ssh_config_path),
+                "none",
                 "-i",
                 id_rsa_path,
                 "-W",
--- a/gateway/src/dstack/gateway/services/tunnel.py
+++ b/gateway/src/dstack/gateway/services/tunnel.py
@@ -8,9 +8,12 @@ from typing import List, Optional
 
 from pydantic import BaseModel
 
+from dstack._internal.core.services.configs import ConfigManager
 from dstack.gateway.errors import SSHError
 
 logger = logging.getLogger(__name__)
+
+config = ConfigManager()
 
 
 class SSHTunnel(BaseModel):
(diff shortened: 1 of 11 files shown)


gateway/src/dstack/gateway/services/tunnel.py
Egor-S, line 1: You should not change this file. It is not a part of a CLI (`src/dstack/_internal/cli`). `dstack-gateway` is a separate package deployed on a gateway instance. There is no custom SSH config

runner/internal/gateway/ssh.go
Egor-S, line 1: This code is not used anymore. `dstack-gateway` opens a connection to the `dstack-runner`

src/dstack/_internal/server/services/ssh/__init__.py
Egor-S, line 1: It's a server code and config manager is supposed to be used by CLI. I guess it's better to use `-F none` here

Egor-S (changes requested):

gateway/src/dstack/gateway/services/tunnel.py
spott: Should we change it to `-F none` if there is no custom SSH config?

src/dstack/_internal/server/services/ssh/__init__.py
spott: Done.

gateway/src/dstack/gateway/services/tunnel.py
Egor-S: We can, but actually there isn't any config — it's a dedicated instance fully managed by dstack

runner/internal/gateway/ssh.go
spott: I'm not sure what I should do here... if it is unused, should I delete it?

gateway/src/dstack/gateway/services/tunnel.py
spott: I changed it to `-F none`.  Let me know if you think something else should be done.

peterschmidt85:
@spott I've pushed some changes. Please let me know if all looks good to you.
cc @Egor-S

src/dstack/_internal/utils/ssh.py
```diff
@@ -56,8 +56,15 @@ def include_ssh_config(path: PathLike, ssh_config_path: PathLike = default_ssh_c
             with open(ssh_config_path, "r") as f:
                 content = f.read()
         if include not in content:
-            with open(ssh_config_path, "w") as f:
-                f.write(include + content)
+            try:
```
peterschmidt85, line 62: @spott 
I created the corresponding issue: #937

peterschmidt85 (changes requested):
Left the comment about handling the permission error. Please let me know what you think.

src/dstack/_internal/utils/ssh.py
peterschmidt85: Also, it's important to update the `update_ssh_config` function too.

src/dstack/_internal/utils/ssh.py
```diff
@@ -56,8 +56,15 @@ def include_ssh_config(path: PathLike, ssh_config_path: PathLike = default_ssh_c
...
+                with open(ssh_config_path, "w") as f:
+                    f.write(include + content)
+            except PermissionError:
+                print(f"Couldn't write to {ssh_config_path} due to a permissions problem")
```
Egor-S: @peterschmidt85 
We've discussed that a logger is a better option there. But the default logging level is CRITICAL, so we need to change it in `src/dstack/_internal/cli/utils/common.py : configure_logging`

Egor-S (approved):

PR #944, opened 2024-02-28
Refactor runs and jobs processing

### LINKED ISSUE #914: Separate run and jobs processing
We need to introduce separate `process_runs` for future multi-nodes and replicas.

1. `run.status` is a dedicated field, not directly computed from jobs' statuses
2. Jobs are resubmitted by `process_runs`, not `process_jobs`
3. In the future, `process_runs` should manage
    * auto-scaling
    * interconnected clusters
    * service register and unregister (gateway)

Egor-S:
Closes #914

This PR features essential changes for replicas and multi-node support in dstack.

* Introduce `process_runs` background task
* Add `JobModel.replica_num`
* Add `RunStatus`
* Add `RunTerminationReason` and rename `JobErrorCode` to `JobTerminationReason`
* Runs and jobs have to pass `TERMINATING` before getting final status. `termination_reason` defines the final status
    * Drop `removed` flag from JobModel
* Add `RunModel.gateway` for services

2024/02/29
* Add `RunModel.service_spec` to store URL and model details

Known TODO before the merge:
* [x] Clean up all TODOs in the code
* [x] Check against race conditions and deadlocks
* [x] Squash and rebase migrations
* [x] Add services support
* [x] Test services
* [x] Test with VastAI
* [x] Force container termination in the dstack-shim
* [x] Add documentation to the architecture section


src/dstack/_internal/server/services/jobs/__init__.py
```diff
@@ -169,19 +118,138 @@ def _get_job_configurator(run_spec: RunSpec) -> JobConfigurator:
...
+        # it's too early to terminate the instance
+        return
+
+    res = await session.execute(
+        sa.select(InstanceModel)
+        .where(
+            InstanceModel.project_id == job_model.project_id,
+            InstanceModel.job_id == job_model.id,
+        )
+        .options(sa.orm.joinedload(InstanceModel.project))
```
r4victor: "orm" is not a known member of module "sqlalchemy" Pylance

src/dstack/_internal/server/services/jobs/__init__.py
```diff
@@ -169,19 +118,138 @@ def _get_job_configurator(run_spec: RunSpec) -> JobConfigurator:
...
+                "%s: instance '%s' has been released, new status is %s",
+                fmt(job_model),
+                instance.name,
+                instance.status.name,
+            )
+            # TODO(egor-s): unregister service replica
+        finally:
+            PROCESSING_POOL_IDS.remove(instance.id)
+
+    job_model.status = job_termination_reason_to_status(job_model.termination_reason)
```
r4victor: job_model.termination_reason can be None. I understand the code assumes that if the status is Terminating, then job_model.termination_reason != None, but this is fragile. Maybe handle the None case explicitly, e.g. having default termination reason.

src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -257,57 +271,102 @@ async def generate_gateway_name(session: AsyncSession, project: ProjectModel) ->
...
         raise ServerClientError("Gateway tunnel is not working")
     except httpx.RequestError as e:
         raise GatewayError(f"Gateway is not working: {e}")
 
 
+async def register_replica(
+    gateway: GatewayModel, run: Run, job_provisioning_data: JobProvisioningData
+):
+    if gateway is None:
+        raise GatewayError("Gateway is not found")
```
r4victor: Make gateway Optional since the function checks for None.

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -0,0 +1,283 @@
...
+    async with get_session_ctx() as session:
+        async with PROCESSING_RUNS_LOCK:
+            res = await session.execute(
+                sa.select(RunModel).where(
+                    RunModel.status.not_in(RunStatus.finished_statuses()),
+                    RunModel.last_processed_at < get_current_datetime() - PROCESSING_INTERVAL,
+                    RunModel.id.not_in(PROCESSING_RUNS_IDS),
+                )
+            )
+            runs: List[RunModel] = res.scalars().all()
```
r4victor: No need for List[RunModel]. Moreover, it's a wrong type.

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -0,0 +1,283 @@
...
+            PROCESSING_RUNS_IDS.update(run.id for run in runs)
+
+    futures = [process_single_run(run.id, [job.id for job in run.jobs]) for run in runs]
+    for future in asyncio.as_completed(futures):
+        try:
+            run_id = await future
+            PROCESSING_RUNS_IDS.remove(run_id)  # unlock job processing as soon as possible
+        except Exception as e:
```
r4victor: We should't catch Exception. E.g. it would include asyncio's CancelledError, preventing the server stopping.

Should catch only expected exceptions. And if something goes wrong, the run should be failed. PROCESSING_RUNS_IDS should be cleared in finally.

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -0,0 +1,283 @@
...
+    await wait_unlock(SUBMITTED_PROCESSING_JOBS_LOCK, SUBMITTED_PROCESSING_JOBS_IDS, jobs_ids_set)
+    await wait_unlock(RUNNING_PROCESSING_JOBS_LOCK, RUNNING_PROCESSING_JOBS_IDS, jobs_ids_set)
+    await wait_unlock(
+        TERMINATING_PROCESSING_JOBS_LOCK, TERMINATING_PROCESSING_JOBS_IDS, jobs_ids_set
+    )
+
+    async with get_session_ctx() as session:
+        run = await session.get(RunModel, run_id)
+        if run is None:
+            raise ValueError(f"Run {run_id} not found")
```
r4victor: Do not raise general-purpose ValueError. Handle the error case appropriately, e.g. ignore/log/mark run as failed. If needed to propagate to the caller, raise a custom error.

src/dstack/_internal/server/models.py
```diff
@@ -188,15 +203,17 @@ class JobModel(BaseModel):
...
+    )
     job_spec_data: Mapped[str] = mapped_column(String(4000))
     job_provisioning_data: Mapped[Optional[str]] = mapped_column(String(4000))
     runner_timestamp: Mapped[Optional[int]] = mapped_column(Integer)
     # `removed` is used to ensure that the instance is killed after the job is finished
-    removed: Mapped[bool] = mapped_column(Boolean, default=False)
     remove_at: Mapped[Optional[datetime]] = mapped_column(DateTime)
     instance: Mapped[Optional["InstanceModel"]] = relationship(back_populates="job")
     used_instance_id: Mapped[Optional[uuid.UUID]] = mapped_column(UUIDType(binary=False))
+    replica_num: Mapped[int] = mapped_column(Integer)
```
r4victor, line 216: Maybe define replica_num closer to the top for better readability.

r4victor (approved):

src/dstack/_internal/server/background/tasks/process_runs.py
Egor-S: Why is it the wrong type?

I will remove it. In the first place, I've added it to help PyCharm with type hinting. Does VS Code accurately compute sqlalchemy return types?

src/dstack/_internal/server/models.py
Egor-S: I suggest reordering all fields as a separate issue. I don't see "the only best" solution here right now.

src/dstack/_internal/server/background/tasks/process_runs.py
r4victor: The defined return type is sequence: `def all(self) -> Sequence[_R]:`. Casting it to List raises Pylance warning.

Issue #958, opened 2024-03-03
[Bug]: `vastai` doesn't show any offers since `0.16.0`


peterschmidt85:
Steps to reproduce:

1. Configure a `vastai` backend in `~/.dstack/server/config.yml`
2. Define `.dstack.yml` (with any or no `resources`)
3. `dstack run .`

No offers are shown. 

If I proceed, I get:

```
All provisioning attempts failed. This is likely due to cloud providers not having enough capacity. Check CLI and server logs for more details.
```

The workaround is to downgrade to `0.15.1` (note this requires deleting `~/.dstack/server/data/`).

peterschmidt85:
Is fixed in `0.16.1`

PR #959, opened 2024-03-03
Fix vast ai no offers. Issue #958

iRohith:
Added vast.ai to BACKENDS_WITH_CREATE_INSTANCE_SUPPORT list

### DIFF
--- a/src/dstack/_internal/server/services/runs.py
+++ b/src/dstack/_internal/server/services/runs.py
@@ -89,8 +89,7 @@ BACKENDS_WITH_CREATE_INSTANCE_SUPPORT = [
     BackendType.DATACRUNCH,
     BackendType.GCP,
     BackendType.LAMBDA,
-    BackendType.TENSORDOCK,
-    BackendType.VASTAI
+    BackendType.TENSORDOCK
 ]
 
 logger = get_logger(__name__)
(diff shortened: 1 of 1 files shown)


TheBits:
Hi, Rohith!
Thanks for paying attention to dstack!

This list shouldn't include Vastai. The `create_instance` method is not supported by Vastai.

Based on your fix, I found that we shouldn't filter the backends list in the `async def get_offers_by_requirements`. Can you please remove the filter with `BACKENDS_WITH_CREATE_INSTANCE_SUPPORT` from `async def get_offers_by_requirements`?

iRohith:
Updated and removed filter with BACKENDS_WITH_CREATE_INSTANCE_SUPPORT from async def get_offers_by_requirements

src/dstack/_internal/server/services/runs.py
```diff
@@ -89,7 +89,7 @@
     BackendType.DATACRUNCH,
     BackendType.GCP,
     BackendType.LAMBDA,
-    BackendType.TENSORDOCK,
+    BackendType.TENSORDOCK
```
TheBits: It need a comma here
```suggestion
    BackendType.TENSORDOCK,
```

PR #967, opened 2024-03-05
Inherit core models from DualBaseModel

### LINKED ISSUE #957: Allow to introduce new pydantic models fields without breaking clients
As of now, introducing new fields to common pydantic models is backward-incompatible since models have `extra = Extra.forbid` set. We can remove this setting if pydantic models used to parse user input (configuration, profile) are decoupled.

See: https://github.com/pydantic/pydantic/discussions/2652

r4victor:
Closes #957


r4victor:
It turned out `pydantic-duality` does not support Python 3.8 and 3.9.
Opened an issue: https://github.com/zmievsa/pydantic-duality/issues/8

PR #977, opened 2024-03-07
Fix InstanceStatus migration. Added a test to check migrations

### LINKED ISSUE #976: Fix migration
The right database migration was not completed after the InstanceStatus enum were updated.

TheBits:


r4victor (approved):

Egor-S:
The migrations test works only from the repo root. I would suggest replacing the relative path with the absolute one, built from the `__file__`

PR #986, opened 2024-03-08
Implement replicated services

### LINKED ISSUE #966: [Feature]: Support running multiple replicas for a service
dstack must support multiple replicas for services.
Replicas are independent of each other and dstack-gateway balances requests between them.

Egor-S:
Closes #966 

* New field in services configuration `replicas` defaults to 1
* Gateway refactoring
  * **Incompatible** changes in internal state representation: all services have to be recreated manually after deployment
  * Operates with run_id as service_id and job_id as replica_id
  * Paired API methods for registration and registration of services and replicas
  * Return 503 if no replicas
  * Rollback broken nginx configurations
* Add process_runs tests for replicas
* Show job statuses while `dstack run` attaches to logs
* Show replicas in `dstack ps` on separate lines
* Add JobSpec.replica_num
  * `Run.jobs` should not be indexed with `JobSpec.job_num` anymore
* Drop JobSpec.pool_name

Minimal configuration:

```yaml
type: service
port: 8000
commands:
  - uname -n > uname.html
  - python3 -m http.server 8000
replicas: 2
auth: false
```

1. Run and open the link
2. Navigate to `uname.html` and reload it (be aware of caching, use Ctrl+Shift+F5/Cmd+Shift+R)
3. See how content changes from time to time
4. Kill any instance with the cloud provider console
5. See how the service works with a single replica and later replaces the second one


gateway/src/dstack/gateway/core/store.py
```diff
@@ -0,0 +1,316 @@
...
+                ),
+                # Unregister from nginx
+                self.nginx.unregister_domain(service.domain),
+                # Notify subscribers
+                *(
+                    subscriber.on_unregister(project, service.id)
+                    for subscriber in self._subscribers
+                ),
+                return_exceptions=True,
+            )
```
r4victor: Since return_exceptions=True, all exceptions raised inside coroutines will not be reraised and, thus, silently ignored. I think we should at least log unexpected exceptions with error level.

gateway/src/dstack/gateway/core/store.py
```diff
@@ -0,0 +1,316 @@
...
+                service.domain,
+            )
+
+            await asyncio.gather(
+                # Terminate SSH tunnel
+                run_async(replica.ssh_tunnel.stop),
+                # Remove from nginx
+                self.nginx.remove_upstream(service.domain, replica.id),
+                return_exceptions=True,
+            )
```
r4victor: Same – ignored exceptions

src/dstack/_internal/server/services/runs.py
```diff
@@ -342,18 +341,25 @@ async def submit_run(
     )
     session.add(run_model)
 
+    replicas = 1
     if run_spec.configuration.type == "service":
+        replicas = run_spec.configuration.replicas.min
```
r4victor, line 346: Can't run_spec.configuration.replicas.min be None, e.g. if I specify `replicas: ..3`?

gateway/src/dstack/gateway/registry/routes.py
```diff
@@ -2,60 +2,77 @@
...
+
+@router.post("/services/{run_id}/replicas/{job_id}/unregister")
+async def post_unregister_replica(
+    project: str, run_id: str, job_id: str, store: Annotated[Store, Depends(get_store)]
+) -> OkResponse:
+    await store.unregister_replica(project, run_id, job_id)
+    return OkResponse()
+
+
+@router.post("/entrypoints/register")
```
r4victor, line 71: What's the point of this endpoint? I don't see it to be used

src/dstack/_internal/server/background/tasks/process_submitted_jobs.py
```diff
@@ -106,7 +107,15 @@ async def _process_submitted_job(session: AsyncSession, job_model: JobModel):
     )
 
     run = run_model_to_run(run_model)
-    job = run.jobs[job_model.job_num]
+    job = find_job(run.jobs, job_model.replica_num, job_model.job_num)
```
r4victor, line 110: It seems there is no need in find_job returning Optional. Maybe it should return value or error? Also, it seems like getting a job should always succeed, so there is no need to handle the None case explicitly.

r4victor (approved):

gateway/src/dstack/gateway/registry/routes.py
Egor-S: My mistake, thanks for pointing it out. I will test OpenAI integration, which uses this endpoint

PR #1003, opened 2024-03-11
Allow to specify required env variables (and allow their automatic capture from the user environment)

spott:
added support for grabbing an environment variable from the environment, also should throw an error when you add an env variable twice

I'm still working on testing this.


spott:
I was able show this works with the following `.dstack.yml`:

```
type: task

python: "3.10"

env:
  - WANDB="yes"
  - HOMEBREW_REPOSITORY
commands:
  - echo "${HOMEBREW_REPOSITORY}"
  - echo "${WANDB}"
  - sleep 100
```

where `HOMEBREW_REPOSITORY` is just some random local env variable I had.

The output is:

```
/opt/homebrew
"yes"
```

I'm marking this ready for review now.  (I'm not sure how to encapsulate this test into your test suite, if someone wanted to help with that).

r4victor:
@spott, I'm afraid capturing environment variables inside the BaseConfiguration validator is a bad idea since the validator will be also triggered on the server side, which would allow retrieving any server environment variable.

The better place to put this logic is BaseRunConfigurator: https://github.com/dstackai/dstack/blob/0244dab3a075d385059210995addc6fb466503f8/src/dstack/_internal/cli/services/configurators/run.py#L52

This way, the environment variables will be captured only on the client side.

spott:
Ok, I updated the pull request and re-tested it, let me know if there is anything else that needs to be done.

r4victor:
Merged with minor tweaks. Thanks!

PR #1005, opened 2024-03-12
Clarify gateway development docs

jvstme:
Apparently one should use the gateway key, the project key does not work


Egor-S (approved):
Everything is correct. Thanks

PR #1022, opened 2024-03-18
Allow configuring CA for Gateway

### LINKED ISSUE #1004: [Bug]: Can't run a service on dstack Sky because of LetsEncrypt's `too many certificates already issued for "dstack.ai"`
`dstack` is using LetsEncrypt for creating SSL certificates when deploying services. LetsEncrypt has a [rate limit](https://letsencrypt.org/docs/rate-limits/).

Because dstack Sky is using the same domain for all users, it may exceed the limit, resulting in a failure to deploy a service:

Screenshot_2024-03-12_at_2 16 04_PM

## Solution

Add optional support for [ZeroSSL](https://zerossl.com/) (which can be enabled via an environment variable, e.g. `ZEROSSL_API_KEY`) and use it instead of LetsEncrypt when enabled.

## Workarounds:

1. Wait until the rate limit is reset (can be any time later today)
2. [Allow the user to specify a custom domain](https://github.com/dstackai/dstack/issues/997) (not supported yet)

jvstme:
This PR allows configuring CAs other than Let's Encrypt for issuing TLS certificates at the Gateway. Any CA that supports ACME with reusable EAB credentials can be used.

Implementation:
- When the Server starts, it reads new environment variables `DSTACK_ACME_*`
- When the Server creates or initializes a Gateway, it sends these ACME settings to the Gateway through the new `/api/config` method
- The Gateway saves the settings in persistent storage
- The Gateway passes the settings to Certbot when requesting new certificates

Closes #1004


src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -451,6 +463,34 @@ async def _update_gateway(connection: GatewayConnection, build: str):
...
+            logger.debug(
+                "Failed attempt %s at configuring gateway %s: %s",
+                attempt + 1,
+                connection.ip_address,
+                e,
+            )
+            await asyncio.sleep(GATEWAY_CONFIGURE_DELAY)
+    else:
+        async with connection.client() as client:
+            await client.submit_gateway_config()
```
r4victor, line 489: Should work ok, but I wouldn't call submit_gateway_config in two different places – it's a bit redundant and also the error handling would work differently on last attempt.

Something like this is more readable:

```python
    for attempt in range(GATEWAY_CONFIGURE_ATTEMPTS):
        try:
            async with connection.client() as client:
                await client.submit_gateway_config()
        except httpx.NetworkError as e:
            logger.debug(
                "Failed attempt %s at configuring gateway %s: %s",
                attempt + 1,
                connection.ip_address,
                e,
            )
            await asyncio.sleep(GATEWAY_CONFIGURE_DELAY)
        else:
            logger.info("Gateway %s configured", connection.ip_address)
            return

```

src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -54,6 +54,8 @@
 
 GATEWAY_CONNECT_ATTEMPTS = 9
 GATEWAY_CONNECT_DELAY = 10
+GATEWAY_CONFIGURE_ATTEMPTS = 40
```
r4victor, line 57: Any reasons to have this many attempts?

r4victor (approved):

src/dstack/_internal/server/services/gateways/__init__.py
jvstme: > the error handling would work differently on last attempt

That was intentional. If the last attempt fails we give up and let the caller handle the error. Here the callers log a different message with WARNING or ERROR level instead of DEBUG.

I agree this retry logic is not very readable. I considered using [tenacity](https://tenacity.readthedocs.io), but that would lead to less detailed log messages or more code to get detailed messages, so I gave up on it for now

src/dstack/_internal/server/services/gateways/__init__.py
jvstme: In my testing, there were about 50 seconds from the moment the SSH tunnel was up to the moment the Gateway's HTTP server was ready to handle requests. I decided to add some margin, so here the max total delay is about 120 seconds (40 attempts with 3-second delays). We could use fewer attempts with longer delays, but then the user would have to wait longer.

I don't like 3-second delays either. Ideally, the Gateway could notify the Server when it's ready so that the Server could start interacting immediately. But that seems less trivial, so I didn't choose that path

src/dstack/_internal/server/services/gateways/__init__.py
r4victor: Got it

src/dstack/_internal/server/services/gateways/__init__.py
r4victor: > the Gateway could notify the Server when it's ready so that the Server could start interacting immediately

Yeah, that's an option, but you're right – definitely out of scope of this PR.

PR #1024, opened 2024-03-18
Set default values for the retry_policy column for existing data

TheBits:


src/dstack/_internal/server/migrations/versions/555138b1f77f_change_instancemodel_for_asynchronous_.py
```diff
@@ -34,6 +34,7 @@ def upgrade() -> None:
         )
 
     # ### end Alembic commands ###
+    op.execute(sa.sql.text("UPDATE instances SET retry_policy = FALSE"))
```
jvstme, line 37: Does not seem to help, I still get the same error, as it happens before this line.
```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) NOT NULL constraint failed: _alembic_tmp_instances.retry_policy
[SQL: INSERT INTO _alembic_tmp_instances (id, name, created_at, deleted, deleted_at, project_id, pool_id, status, started_at, finished_at, termination_policy, termination_idle_time, backend, backend_data, region, price, job_provisioning_data, offer, job_id, last_job_processed_at, termination_deadline, termination_reason, health_status) SELECT instances.id, instances.name, instances.created_at, instances.deleted, instances.deleted_at, instances.project_id, instances.pool_id, instances.status, instances.started_at, instances.finished_at, instances.termination_policy, instances.termination_idle_time, instances.backend, instances.backend_data, instances.region, instances.price, instances.job_provisioning_data, instances.offer, instances.job_id, instances.last_job_processed_at, instances.termination_deadline, instances.termination_reason, instances.health_status 
FROM instances]
```
Can you reproduce it if you run this migration when the `instances` table already has records?

Maybe add `default` instead? I'm not sure why Alembic didn't do this automatically.
```python
batch_op.add_column(sa.Column("retry_policy", sa.Boolean(), nullable=False, default=False))
```

jvstme (changes requested):

jvstme (approved):

PR #1055, opened 2024-03-27
Fixed the YAML schema validation for `replicas`

peterschmidt85:
Previously, the YAML schema showed errors if `replicas` was set to a `str` or `int`

### DIFF
--- a/src/dstack/_internal/core/models/configurations.py
+++ b/src/dstack/_internal/core/models/configurations.py
@@ -1,6 +1,6 @@
 import re
 from enum import Enum
-from typing import Dict, List, Mapping, Optional, Union
+from typing import Any, Dict, List, Mapping, Optional, Union
 
 from pydantic import BaseModel, Field, ValidationError, conint, constr, root_validator, validator
 from typing_extensions import Annotated, Literal
@@ -233,7 +233,10 @@ class ServiceConfiguration(BaseConfiguration):
         Field(description="Mapping of the model for the OpenAI-compatible endpoint"),
     ] = None
     auth: Annotated[bool, Field(description="Enable the authorization")] = True
-    replicas: Annotated[Range[int], Field(description="The range ")] = Range[int](min=1, max=1)
+    replicas: Annotated[
+        Union[conint(ge=1), constr(regex=r"^[0-9]+..[1-9][0-9]*$"), Range[int]],
+        Field(description="The range "),
+    ] = Range[int](min=1, max=1)
     scaling: Annotated[
         Optional[ScalingSpec], Field(description="The auto-scaling configuration")
     ] = None
@@ -247,15 +250,20 @@ class ServiceConfiguration(BaseConfiguration):
         return v
 
     @validator("replicas")
-    def convert_replicas(cls, v: Range[int]) -> Range[int]:
+    def convert_replicas(cls, v: Any) -> Range[int]:
+        if isinstance(v, str) and ".." in v:
+            min, max = v.replace(" ", "").split("..")
+            v = Range(min=min or 0, max=max or None)
+        elif isinstance(v, (int, float)):
+            v = Range(min=v, max=v)
         if v.max is None:
-            raise ValueError("The maximum number of replicas must be set")
-        if v.max < 1:
-            raise ValueError("The maximum number of replicas must be greater than 0")
-        if v.min is None:
-            v.min = 0
-        elif v.min < 0:
-            raise ValueError("The minimum number of replicas must be greater or equal than 0")
+            raise ValueError("The maximum number of replicas is required")
+        if v.min < 0:
+            raise ValueError("The minimum number of replicas must be greater than or equal to 0")
+        if v.max < v.min:
+            raise ValueError(
+                "The maximum number of replicas must be greater than the minium number of replicas"
+            )
         return v
 
     @root_validator()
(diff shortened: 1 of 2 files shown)


src/dstack/_internal/core/models/configurations.py
```diff
@@ -233,7 +233,10 @@ class ServiceConfiguration(BaseConfiguration):
         Field(description="Mapping of the model for the OpenAI-compatible endpoint"),
     ] = None
     auth: Annotated[bool, Field(description="Enable the authorization")] = True
-    replicas: Annotated[Range[int], Field(description="The range ")] = Range[int](min=1, max=1)
+    replicas: Annotated[
+        Union[conint(ge=1), constr(regex=r"^[0-9]+..[1-9][0-9]*$"), Range[int]],
+        Field(description="The range "),
```
jvstme, line 238: (off-topic) Not very descriptive

src/dstack/_internal/core/models/configurations.py
```diff
@@ -247,25 +250,30 @@ def convert_port(cls, v) -> PortMapping:
...
-        if v.min is None:
-            v.min = 0
-        elif v.min < 0:
-            raise ValueError("The minimum number of replicas must be greater or equal than 0")
+            raise ValueError("The maximum number of replicas is required")
+        if v.min < 0:
+            raise ValueError("The minimum number of replicas must be greater than or equal to 0")
+        if v.max < v.min:
+            raise ValueError(
+                "The maximum number of replicas must be greater than the minium number of replicas"
```
jvstme: (nit) Technically, greater than or equal

src/dstack/_internal/core/models/configurations.py
```diff
@@ -233,7 +233,10 @@ class ServiceConfiguration(BaseConfiguration):
         Field(description="Mapping of the model for the OpenAI-compatible endpoint"),
     ] = None
     auth: Annotated[bool, Field(description="Enable the authorization")] = True
-    replicas: Annotated[Range[int], Field(description="The range ")] = Range[int](min=1, max=1)
+    replicas: Annotated[
+        Union[conint(ge=1), constr(regex=r"^[0-9]+..[1-9][0-9]*$"), Range[int]],
```
jvstme, line 237: I'm afraid this breaks type checking. The type checker sees `replicas` as `Union[int, str, Range[int]]`, but we use it specifically as `Range[int]` ([1](https://github.com/dstackai/dstack/blob/04a62ba630fd775c05bbdeb0f9a7519a155bde46/src/dstack/_internal/server/services/gateways/autoscalers.py#L92), [2](https://github.com/dstackai/dstack/blob/04a62ba630fd775c05bbdeb0f9a7519a155bde46/src/dstack/_internal/server/background/tasks/process_runs.py#L133), [3](https://github.com/dstackai/dstack/blob/04a62ba630fd775c05bbdeb0f9a7519a155bde46/src/dstack/_internal/server/services/runs.py#L422), etc)

src/dstack/_internal/core/models/configurations.py
```diff
@@ -247,25 +250,30 @@ def convert_port(cls, v) -> PortMapping:
         return v
 
     @validator("replicas")
-    def convert_replicas(cls, v: Range[int]) -> Range[int]:
+    def convert_replicas(cls, v: Any) -> Range[int]:
+        if isinstance(v, str) and ".." in v:
```
jvstme, line 254: It's not perfect that we duplicate range parsing logic. It is already implemented in [`Range`](https://github.com/dstackai/dstack/blob/04a62ba630fd775c05bbdeb0f9a7519a155bde46/src/dstack/_internal/core/models/resources.py#L26)

jvstme (commented):
Apparently, to avoid the type checking and code duplication issues (see comments below), we manipulate the JSON schema directly in other places where `Range` is used, see [`ResourcesSpec`](https://github.com/dstackai/dstack/blob/04a62ba630fd775c05bbdeb0f9a7519a155bde46/src/dstack/_internal/core/models/resources.py#L202). I'd suggest we try the same thing here with `replicas`. Or we can merge as is and come back to it later

peterschmidt85:
> Apparently, to avoid the type checking and code duplication issues (see comments below), we manipulate the JSON schema directly in other places where `Range` is used, see [`ResourcesSpec`](https://github.com/dstackai/dstack/blob/04a62ba630fd775c05bbdeb0f9a7519a155bde46/src/dstack/_internal/core/models/resources.py#L202). I'd suggest we try the same thing here with `replicas`. Or we can merge as is and come back to it later

Yeah, I was also thinking of it. The problem is we have a mess there already. Look at `ResourcesSpec` and `ResourcesSpecSchema`. I'd perhaps right now add a test, and leave it as is. And later, come back and fix the mess.

jvstme (approved):

peterschmidt85:
@jvstme I've added the test. Let me know if it's OK to merge

src/tests/_internal/core/models/test_configurations.py
```diff
@@ -0,0 +1,52 @@
...
+                "0..10",
+                {
+                    "metric": "rps",
+                    "target": 10,
+                },
+            )
+        ).replicas == Range(min=0, max=10)
+        with pytest.raises(
+            ConfigurationError,
+            match="When you set `replicas` to a range, ensure to specify `scaling`",
```
jvstme, line 42: (nit) I think the primary error here is `unexpected value; permitted: 'rps'`. The `ensure to specify scaling` error also appears, but as a consequence of the primary error

jvstme (approved):

jvstme:
@peterschmidt85, thank you! I think it's ok to merge

PR #1061, opened 2024-03-29
Return an erorr information from shim to dstack cli

### LINKED ISSUE #1060: [Bug]: The `dstack run` should show the error from runner or shim
### Steps to reproduce

1. Run a job with an error in the docker image name
2. The pulling image process will be interrupted by an error in the shim
3. The CLI shows to me the error `jobErrorCode.INTERRUPTED_BY_NO_CAPACITY` without any details.

### Expected behaviour

dstack run should show additional details about the error that occurred in a shim or runner. 

### Actual behaviour

I only see the error code:
```
lazy-lion-1 provisioning completed (failed)
Run failed with error code JobErrorCode.INTERRUPTED_BY_NO_CAPACITY. Check CLI and server logs for more details.
```

### dstack version

0.16.5

### Server logs

_No response_

### Additional information

There are a lot of errors during job initialization on shim:
- git clone and git authentication errors
- an error in the image name
- invalid credentials for docker registry
- timeout when docker pulling an image

TheBits:
Fixes #1060


src/dstack/_internal/server/migrations/versions/99b4c8c954ea_add_termination_reason_message.py
jvstme, line 1: Maybe there should also be a migration of `jobs.termination_reason` adding two new enum variants? Like [here](https://github.com/dstackai/dstack/blob/7c6b52733d3436f91ec407d127318c981333d3a4/src/dstack/_internal/server/migrations/versions/4b4319398164_introduce_runs_processing.py#L68).

I guess it is not necessary for SQLite. Apparently, for SQLite SQLAlchemy stores enums as `VARCHAR(max_variant_length)` and the new variants do not exceed the current length constraint, which is 34. But it can become important if we ever support another RDBMS

src/dstack/_internal/core/models/runs.py
```diff
@@ -96,6 +96,8 @@ class JobTerminationReason(str, Enum):
     # Set by the runner
     CONTAINER_EXITED_WITH_ERROR = "container_exited_with_error"
     PORTS_BINDING_FAILED = "ports_binding_failed"
+    CREATING_CONTAINER_ERROR = "creating_container_error"
```
jvstme, line 100: Should these two be in the `Set by the runner` section? It looks like they are set by the server

src/dstack/_internal/cli/commands/run.py
```diff
@@ -226,12 +226,24 @@ def _command(self, args: argparse.Namespace):
...
     termination_reason = _get_run_termination_reason(run)
     message = "Run failed due to unknown reason. Check CLI and server logs."
-    if _get_run_termination_reason(run) == JobTerminationReason.FAILED_TO_START_DUE_TO_NO_CAPACITY:
+    if termination_reason == JobTerminationReason.FAILED_TO_START_DUE_TO_NO_CAPACITY:
         message = (
             "All provisioning attempts failed. "
             "This is likely due to cloud providers not having enough capacity. "
             "Check CLI and server logs for more details."
         )
+    elif termination_reason == JobTerminationReason.CONTAINER_EXITED_WITH_ERROR:
```
jvstme: Isn't it `CREATING_CONTAINER_ERROR` here?

jvstme: Actually, it might be less error-prone to not have different messages for different termination reasons, but rather something like
```python
if termination_reason is not None:
    termination_reason_message = run._run.jobs[0].job_submissions[0].termination_reason_message

    message = f"Run failed with error code {termination_reason}"
    if termination_reason_message:
        message += f". Error: {termination_reason_message}"
```

src/dstack/_internal/cli/commands/run.py
```diff
@@ -226,12 +226,24 @@ def _command(self, args: argparse.Namespace):
...
+        message = (
+            "Cannot create container. "
+            f"Error: {run._run.jobs[0].job_submissions[0].termination_reason_message} "
+            "Check CLI and server logs for more details."
+        )
+    elif termination_reason == JobTerminationReason.EXECUTOR_ERROR:
+        message = (
+            "There is an error with git auth/clone or with command execution "
```
jvstme, line 243:
```suggestion
            "There is an error with git auth/clone or with command execution. "
```

src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -194,15 +194,18 @@ async def _process_job(job_id: UUID):
                     run_model,
                     job_model,
                 )
+                if not success:
+                    job_model.termination_reason = JobTerminationReason.INTERRUPTED_BY_NO_CAPACITY
 
             if not success:  # kill the job
                 logger.warning(
-                    "%s: failed because runner is not available, age=%s",
+                    "%s: failed because runner is not available or return an error,  age=%s",
```
jvstme, line 201:
```suggestion
                    "%s: failed because runner is not available or returned an error,  age=%s",
```

src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -194,15 +194,18 @@ async def _process_job(job_id: UUID):
                     run_model,
                     job_model,
                 )
+                if not success:
+                    job_model.termination_reason = JobTerminationReason.INTERRUPTED_BY_NO_CAPACITY
```
jvstme, line 197: Isn't it redundant, considering the same thing will happen on line 208?

runner/internal/runner/api/http.go
```diff
@@ -82,8 +82,8 @@ func (s *Server) runPostHandler(w http.ResponseWriter, r *http.Request) (interfa
 	var runCtx context.Context
 	runCtx, s.cancelRun = context.WithCancel(context.Background())
 	go func() {
-		_ = s.executor.Run(runCtx) // todo handle error
-		s.jobBarrierCh <- nil      // notify server that job finished
+		s.executor.Run(runCtx)
```
jvstme: Note the linter error.
Btw, I'd suggest we make the runner-test workflow fail is there are any errors

PR #1063, opened 2024-03-29
Add runpod backend

Bihan:
Add runpod backend First Version

### DIFF
--- a/src/dstack/_internal/core/backends/runpod/api_client.py
+++ b/src/dstack/_internal/core/backends/runpod/api_client.py
@@ -2,7 +2,7 @@ from typing import Any, List, Optional
 
 import requests
 
-from dstack._internal.core.errors import BackendInvalidCredentialsError
+from dstack._internal.core.errors import BackendError, BackendInvalidCredentialsError
 
 API_URL = "https://api.runpod.io/graphql"
 
@@ -97,6 +97,10 @@ class RunpodApiClient:
                 json=data,
             )
             response.raise_for_status()
+            if "errors" in response.json():
+                if "podTerminate" in response.json()["errors"][0]["path"]:
+                    raise BackendError("Instance Not Found")
+                raise BackendError(response.json()["errors"][0]["message"])
             return response
         except requests.HTTPError as e:
             if e.response.status_code in (requests.codes.forbidden, requests.codes.unauthorized):
--- a/src/dstack/_internal/core/backends/runpod/api_client.py
+++ b/src/dstack/_internal/core/backends/runpod/api_client.py
@@ -1,0 +1,276 @@
+from typing import Any, List, Optional
+
+import requests
+
+from dstack._internal.core.errors import BackendInvalidCredentialsError
+
+API_URL = "https://api.runpod.io/graphql"
+
+
+class RunpodApiClient:
+    def __init__(self, api_key: str):
+        self.api_key = api_key
+
+    def validate_api_key(self) -> bool:
+        try:
+            self.get_user_details()
+        except BackendInvalidCredentialsError:
+            return False
+        return True
+
+    def get_user_details(self):
+        resp = self._make_request({"query": user_details_query, "variable": {}})
+        if resp.ok:
+            data = resp.json()
+            return data
+
+    def create_pod(
+        self,
+        name: str,
+        image_name: str,
+        gpu_type_id: str,
+        cloud_type: str = "ALL",
+        support_public_ip: bool = True,
+        start_ssh: bool = True,
+        data_center_id: Optional[str] = None,
+        country_code: Optional[str] = None,
+        gpu_count: int = 1,
+        volume_in_gb: int = 0,
+        container_disk_in_gb: Optional[int] = None,
+        min_vcpu_count: int = 1,
+        min_memory_in_gb: int = 1,
+        docker_args: str = "",
+        ports: Optional[str] = None,
+        volume_mount_path: str = "/runpod-volume",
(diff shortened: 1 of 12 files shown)


src/dstack/_internal/server/services/backends/configurators/runpod.py
```diff
@@ -0,0 +1,91 @@
...
+    def create_backend(
+        self, project: ProjectModel, config: RunpodConfigInfoWithCreds
+    ) -> BackendModel:
+        if config.regions is None:
+            config.regions = REGIONS
+        return BackendModel(
+            type=self.TYPE.value,
+            config=RunpodStoredConfig(**RunpodConfigInfo.parse_obj(config).dict()).json(),
+            auth=RunpodCreds.parse_obj(config.creds).json(),
+        )
```
TheBits: The `project_id` is required argument. These are the reasons why the model is not created on a fresh installation

```suggestion
        return BackendModel(
            project_id=project.id,
            type=self.TYPE.value,
            config=RunpodStoredConfig(**RunpodConfigInfo.parse_obj(config).dict()).json(),
            auth=RunpodCreds.parse_obj(config.creds).json(),
        )
```

src/dstack/_internal/core/backends/runpod/compute.py
```diff
@@ -0,0 +1,150 @@
...
+                response_code = details.get("code")
+            except ValueError:
+                raise BackendError(e)
+
+            # code 3: There are no hosts available for your specified virtual machine.
+            if response_code == 3:
+                raise NoCapacityError(details.get("message"))
+            # code 9: Vm cannot be assigned ip from network. Network full
+            if response_code == 9:
+                raise BackendError(details.get("message"))
```
TheBits: This part from cudo. Please remove it.

src/dstack/_internal/core/backends/runpod/compute.py
```diff
@@ -0,0 +1,150 @@
...
+
+        instance_id = resp["id"]
```
TheBits, line 90: The `create_pod` returns `None` and this line raise TypeError.

For example, if my balance is zero, the 'create_pod' retrun status_code=200 and message will be given.:
```
{"errors":[{"message":"Your account balance is too low to rent a pod. Please add funds to your account.","path":["podFindAndDeployOnDemand"],"extensions":{"code":"RUNPOD"}}],"data":{"podFindAndDeployOnDemand":null}}
```

TheBits (changes requested):

src/dstack/_internal/core/backends/runpod/compute.py
Bihan: Thanks for pointing it out. I will check for `error` field and` raise BackendError` with messages from runpod.

r4victor:
In #967 we migrated pydantic models to pydantic-duality. Now the models should be inherited from `CoreModel` instead of `BaseModel` or `ForbidExtra`. And in configurators, it's recommended to use `MyModel.__response__.parse_obj` for parsing data from the db instead of `MyModel.parse_obj`. You can check out other backends/configurators to see how to migrate.

https://github.com/dstackai/dstack/blob/37ea54a8cb158ba54ae8203be04c883b3a6647e6/src/dstack/_internal/core/models/common.py#L7-L14

src/dstack/_internal/core/backends/runpod/compute.py
Bihan: I have updated the PR with changes in Exceptional Handling. I have added exception handling in `terminate_instance` for runpod.

TheBits (approved):

src/dstack/_internal/core/backends/runpod/api_client.py
```diff
@@ -0,0 +1,280 @@
...
+                    network_volume_id,
+                    allowed_cuda_versions,
+                )
+            }
+        )
+        if resp.ok:
+            data = resp.json()
+            return data["data"]["podFindAndDeployOnDemand"]
+
+    def get_pod(self, pod_id: str):
```
r4victor, line 80: @Bihan, The return types should be annotated. The functions should be written so that they return on all code paths.

r4victor: You do `response.raise_for_status()` in `_make_request()`, so it seems there is no need to check for `if resp.ok`.

Bihan: @r4victor Thank you for point out the issue. Yes `_make_request()` is the sole source as well as it handles the exceptions. Also I will annotate the return types and ensure functions return on all code paths. I will make changes and PR for the Fix.

src/dstack/_internal/core/backends/runpod/compute.py
```diff
@@ -0,0 +1,146 @@
...
+            docker_args=get_docker_args(authorized_keys),
+            ports="22/tcp",
+        )
+
+        instance_id = resp["id"]
+        # Wait until VM State is Active. This is necessary to get the ip_address.
+        pod = self.api_client.get_pod(instance_id)
+        while pod["runtime"] is None:
+            pod = self.api_client.get_pod(resp["id"])
+            time.sleep(1)
```
r4victor, line 95: Potentially an infinite loop. There should be a limit on the number of attempts.

Bihan: Ok I will check the average response time runpod takes,  then add a margin to set number of attempts.

PR #1010, opened 2024-03-13
Support `dstack gateway` for the `cudo` provider

Bihan:

PR #1069, opened 2024-04-01
Add gateway to cudo

Bihan:


peterschmidt85:
How is it different from #1010?

Bihan:
> How is it different from #1010?


@peterschmidt85 #1010 's required changes is already included in the main branch. I think we can delete this PR.
cc @TheBits

TheBits:
Duplicate #1010

PR #1085, opened 2024-04-03
Add json log formatter

### LINKED ISSUE #1062: [Feature]: Add JSON formatter for logging
### Problem

Dstack server outputs unstructured logs. Filtering unstructured logs by fields is not efficient in the CloudWatch. Multi-line unstructured logs (stacktrace for example) makes logs hard to understand.

### Solution

Let's add JSON formatter to dstack server. We will be able to filter easily. And structure logs with additional fields.

### Workaround

_No response_

Yes

TheBits:
Fixes #1062

### DIFF
--- a/src/dstack/_internal/server/background/tasks/process_submitted_jobs.py
+++ b/src/dstack/_internal/server/background/tasks/process_submitted_jobs.py
@@ -128,6 +128,15 @@ async def _process_submitted_job(session: AsyncSession, job_model: JobModel):
             instance = sorted_instances[0]
             instance.status = InstanceStatus.BUSY
             instance.job = job_model
+            logger.info(
+                "The job %s switched instance %s status to BUSY",
+                job_model.job_name,
+                instance.name,
+                extra={
+                    "instance_name": instance.name,
+                    "instance_status": InstanceStatus.BUSY.value,
+                },
+            )
 
             logger.info("%s: now is provisioning on '%s'", fmt(job_model), instance.name)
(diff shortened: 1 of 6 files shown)


src/dstack/_internal/server/background/tasks/process_submitted_jobs.py
```diff
@@ -128,6 +128,15 @@ async def _process_submitted_job(session: AsyncSession, job_model: JobModel):
...
             instance.job = job_model
+            logger.info(
+                "The job %s switched instance %s status to BUSY",
+                job_model.job_name,
+                instance.name,
+                extra={
+                    "instance_name": instance.name,
+                    "instance_status": InstanceStatus.BUSY.value,
+                },
+            )
```
r4victor, line 139: What's the choice for fields passed as extra? I suggest we pass all relevant context there. But we need to support extra for non-json logger as well – otherwise we'll duplicate the context in both message and extra.

r4victor (approved):

src/dstack/_internal/server/background/tasks/process_submitted_jobs.py
TheBits: > What's the choice for fields passed as extra?

To see the lifecycle of an instance, it is easier to read when logs are formatted as a table. I added `instance_name` and `instance_status` as a columns to this "table". Each log with extra fields records a completed action or fact, not a context.

PR #1104, opened 2024-04-09
Make `commands` optional in run configurations

### LINKED ISSUE #938: [UX] Make `commands` optional for tasks and services (and let use the default Docker image entrypoint)
Currently, the `commands` property is required for task and service configurations.

Often, the user may need to use a Docker image that already has the default entry point configured.

## Workaround

1. Ensure you pass `commands`
2. Ensure you set the `working_dir` properly. By default, `dstack` sets it to `/workflow` where it mounts the repos' code (if any).

jvstme:
Now, if the user specifies a custom `image` in
their configuration, they can omit `commands` so
that the image's default entrypoint+cmd and 
default working directory are used.

We always override the image's entrypoint, so it
is necessary to find out what the default
entrypoint was and pass it to the runner. This is
done by the dstack server by requesting the
image's manifest from its container registry.

Closes #938


jvstme:
Sorry for a large PR :( Many of the changed lines are copy-pasted test data and functions becoming async

r4victor (approved):

PR #1115, opened 2024-04-11
Added the ability of adding own instances

### LINKED ISSUE #1100: Ability to add your own instances to pools/dstack
After we finish [pool support](https://github.com/dstackai/dstack/issues/790), we will be ready to add our own servers to the pool. This is necessary for a particular set of users that have their own bare metal servers.

To add a server, we need to use the `dstack pool add` command in the same way as we add cloud instances with additional arguments:
```
dstack pool add \
    --remote \
    --remote-host 192.168.88.131 \
    --remote-port 22 \
    --ssh-user=root \
    --ssh-identity ~/ssh_key
```

I will explain the new arguments:
`--remote` - Adding a running server
`--remote-host 192.168.88.131` and `--remote-port 22` - Connection detils
`--ssh-user=root` and `--ssh-identity ~/ssh_key.pub` - Credentials for ssh

 To add the server, we need to go through the following steps.
 - connect via ssh to the server
 - install `dstack-shim`
 - add dstack-shim.service to the systemd
 
After install, `dstack-shim` should retrieve server resources and save them to the InstanceModel on the dstack server.
This information about server resources will be provided to the user via "dstack pool ps" command.

------

## SSH semantics
It is necessary to change the name the command to `dstack pool add-ssh` to add a remote instance. The add remote instance command must have the same semantics as the ssh command.

Add support for the next cli arguments:
- [x] Support the destination format `[user@]host` or `ssh://[user@]host[:port]`
- [x] `-l login_name` specifies the user
- [x] `-i key` specifies the private key-containing file.
- [x] `-p port` the connection port

## Improvements
- [x] After running the command, it should output a message that includes the `instance name` of the added remote host.
- [x] Support the argument for the `--region`
- [x] Support the argument for the `--pool`
- [x] Give the newly added remote instance a random name.
- [x] Make a draft documentation. 
- [x] Verify the installation of nvidia-docker.
- [x] Use the `dstack/base` image in `docker run` to verify the output of the `nvidia-smi` command.

## multinode support
- [x] Create two instances on the same network, check that they connect to each other.
- [ ] I need to create a new issue with support for remote instances as a backend for multinode.

## Improvements
- [x] The `add_remote` (`process_instances.py`) function blocks the server. For example, Ctrl-C does not work.
- [ ] Add the `-w` flag. Enabling it should show the process of provisioning the remote instance to the idle state. As an example, it should work as `dstack ps -w`.
- [x] Log a warning that the remote instance failed to provisioning
- [ ] Add multionde documentation
- [x] I can add the same instance two times

TheBits:
The first part of #1100


r4victor (commented):
`add_remote()` seems unnecessary long. I suggest this function to be refactored into multiple functions. One quick heuristic to see what can be refactored is to replace all comments of the form "Do that" (like # Prepare connection key) with functions that do the corresponding step.

A good practice is write functions so that they perform actions of the same level and use subfunctions to hide the unnecessary details. This greatly improves readability and maintainability as developers do not need to delve into all the details of what functions do but acquire a level of understanding sufficient for the task at hand.

src/dstack/_internal/server/background/tasks/process_instances.py
```diff
@@ -88,6 +117,136 @@ async def process_instances() -> None:
...
+                    working_dir=DSTACK_WORKING_DIR,
+                    dev=settings.DSTACK_VERSION is None,
+                )
+
+                # Get host info
+                host_info = get_host_info(client, DSTACK_WORKING_DIR)
+                logger.debug("Received a host_info %s", host_info)
+
+        except ProvisioningError as e:
+            logger.exception("Failed remote instance deploy %s", e)
```
r4victor: I don't think it's a good idea to call logger.exception() on any error. For example, it would cause server exceptions being logged on dstack sky when dstack fails to connect to users' instances but users won't see any errors.

For expected errors such as when dstack fails to connect to instances, there should be no exceptions logged. Instead, the failed instance should have an associated error status / message, so that users can see them via CLI/UI.

src/dstack/_internal/server/background/tasks/process_instances.py
```diff
@@ -88,6 +117,136 @@ async def process_instances() -> None:
...
+            )
+
+            # Prepare connection key
+            try:
+                private_string = [
+                    sk.private for sk in remote_details.ssh_keys if sk.private is not None
+                ][0]
+            except IndexError:
+                logger.error("There are no ssh private key")
+                raise BackendError()
```
r4victor: BackendError() is there to be raised by backends. Weird to see it raised in process_instances.py.

src/dstack/_internal/utils/remote.py
```diff
@@ -0,0 +1,164 @@
...
+    sftp_upload(client, tmp_file_path, dot_env)
+    try:
+        _, stdout, _ = client.exec_command(
+            f"sudo mv {tmp_file_path} {working_dir}/{DSTACK_SHIM_ENV_FILE}", timeout=10
+        )
+        output = stdout.read()
+        if output:
+            logger.warning("Cannot move %s: %s", tmp_file_path, output.decode())
+    except (paramiko.SSHException, OSError) as e:
+        raise ProvisioningError(e)
```
r4victor: Don't wrap catched exceptions with custom exceptions to avoid hiding original `__cause__`, `__context__`, and `__traceback__`. To raise custom exceptions with retaining original context use `raise ProvisioningError() from e`

src/dstack/_internal/server/background/tasks/process_instances.py
TheBits: Fixed

src/dstack/_internal/utils/remote.py
r4victor, line 1: src/dstack/_internal/utils/ may not be the best place for remote backend functionality. utils should be imported from all other modules, so they should not import anything from core, for example.
 
How about moving all remote backend functionality to src/dstack/_internal/core/backends/remote?

src/dstack/_internal/utils/remote.py
TheBits: Fixed!
I want to keep the `paramiko` and `socket` exceptions hidden from `process_instances`. So, I wrapped exceptions in ProvisioningError.

src/dstack/_internal/utils/remote.py
TheBits: Good idea. Move `remote.py` to `src/dstack/_internal/core/backends/remote/provisioning.py`

src/dstack/_internal/server/background/tasks/process_instances.py
TheBits: I deleted the logger.exception. There is currently no field to save the error message. I'll do it later.

PR #1125, opened 2024-04-12
Fix isinstance check for core models

### LINKED ISSUE #1124: [Bug]: BackendAuthError when running AWS, Azure and GCP with non-default creds
### Steps to reproduce

Set up a AWS, Azure or GCP backend with non-default creds. Ensure there are no default creds on the machine. Try submitting a run. You'll see:

```
✗ dstack run . -b azure
Unexpected dstack server error
```

This is because the isinstance check does not work when `creds` is parsed via `MyModel.__response__`:

https://github.com/dstackai/dstack/blob/dcb42762ed318f690d4873f5dce64e23500bab91/src/dstack/_internal/core/backends/azure/auth.py#L24

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### dstack version

master

### Server logs

```shell
File "/Users/r4victor/Projects/dstack/dstack/src/dstack/_internal/server/services/backends/__init__.py", line 256, in get_project_backends_with_models
    backend = await run_async(configurator.get_backend, backend_model)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/r4victor/Projects/dstack/dstack/src/dstack/_internal/server/utils/common.py", line 24, in run_async
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/r4victor/.pyenv/versions/3.11.2/lib/python3.11/concurrent/futures/thread.py", line 58, in run
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/r4victor/Projects/dstack/dstack/src/dstack/_internal/server/services/backends/configurators/azure.py", line 194, in get_backend
    return AzureBackend(config=config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/r4victor/Projects/dstack/dstack/src/dstack/_internal/core/backends/azure/__init__.py", line 13, in __init__
    self.credential, _ = auth.authenticate(self.config.creds)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/r4victor/Projects/dstack/dstack/src/dstack/_internal/core/backends/azure/auth.py", line 19, in authenticate
    check_credential(credential)
  File "/Users/r4victor/Projects/dstack/dstack/src/dstack/_internal/core/backends/azure/auth.py", line 39, in check_credential
    raise BackendAuthError()
dstack._internal.core.errors.BackendAuthError
```

### Additional information

_No response_

r4victor:
Fixes #1124

The PR replace isinstance() check for core models with is_core_model_instance() that checks against `__reponse__` explicitly. Changing isinstance() behaviour via metaclass doesn't work because pydantic-duality may depend on isinstance/subclass checks.

TODO: open a pydantic-duality discussion on this issue.

Tested:

- [x] AWS default creds
- [x] AWS access key creds
- [x] Azure default creds
- [x] Azure client creds
- [x] GCP default creds
- [x] GCP service account creds

### DIFF
--- a/src/dstack/_internal/core/models/common.py
+++ b/src/dstack/_internal/core/models/common.py
@@ -20,7 +20,7 @@ class CoreModelMeta(DualBaseModelMeta):
         return (
             type.__subclasscheck__(cls, subclass)
             or issubclass(subclass, cls.__request__)
-            or issubclass(subclass, cls.__request__)
+            or issubclass(subclass, cls.__response__)
         )
 
 
--- a/src/dstack/_internal/core/models/common.py
+++ b/src/dstack/_internal/core/models/common.py
@@ -49,7 +49,7 @@ class Duration(int):
 
 def is_core_model_instance(instance: CoreModel, class_: Type[CoreModel]) -> bool:
     """
-    Implements isintance check for CoreModel such that
+    Implements isinstance check for CoreModel such that
     models parsed with MyModel.__response__ pass the check against MyModel.
     See https://github.com/dstackai/dstack/issues/1124
     """
--- a/src/dstack/_internal/core/models/common.py
+++ b/src/dstack/_internal/core/models/common.py
@@ -1,7 +1,27 @@
 import re
 from typing import Union
 
-from pydantic_duality import DualBaseModel
+from pydantic_duality import DualBaseModel, DualBaseModelMeta
+
+
+# Change DualBaseModelMeta so that model parsed with MyModel.__response__
+# would pass the check for isinstance(instance, MyModel).
+class CoreModelMeta(DualBaseModelMeta):
+    """Metaclass used for custom types."""
+
+    def __instancecheck__(cls, instance) -> bool:
+        return (
+            type.__instancecheck__(cls, instance)
+            or isinstance(instance, cls.__request__)
+            or isinstance(instance, cls.__response__)
+        )
+
+    def __subclasscheck__(cls, subclass: type):
+        return (
+            type.__subclasscheck__(cls, subclass)
+            or issubclass(subclass, cls.__request__)
+            or issubclass(subclass, cls.__request__)
+        )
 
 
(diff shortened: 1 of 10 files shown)


src/dstack/_internal/core/models/common.py
```diff
@@ -45,3 +45,12 @@ def parse(cls, v: Union[int, str]) -> "Duration":
             }[unit]
             return cls(amount * multiplier)
         raise ValueError(f"Cannot parse the duration {v}")
+
+
+def is_core_model_instance(instance: CoreModel, class_: Type[CoreModel]) -> bool:
+    """
```
TheBits, line 51: I propose renaming the variables `instance` and `class_` to correspond with [original `isinstance`](https://docs.python.org/3/library/functions.html#isinstance) (`object` and `classinfo`).

jvstme (approved):
(nit) Consider using `is_core_model_instance` [here](https://github.com/dstackai/dstack/blob/abc5318b5fbcd1a280a1390b1859622cc6ad25f8/src/dstack/_internal/cli/services/configurators/run.py#L66) for consistency and to prevent similar bugs

src/dstack/_internal/core/models/common.py
r4victor: Since it's a positional-only C func, doesn't make a difference. `object` can't be used.

BTW, the pylance signature:
```
def isinstance(__obj: object, __class_or_tuple: _ClassInfo) -> bool: ...
```

TheBits: Thank you!

PR #1140, opened 2024-04-16
Add request timeouts to prevent code freezes

jvstme:
- Enable the ruff rule to prevent using requests without a timeout. The rule doesn't catch all the cases, but it's a best effort.
- Add missing timeouts where I am pretty confident about them.
- Leave TODOs where it would take me too much time to test and decide what the adequate timeout is.

### DIFF
--- a/src/dstack/_internal/server/services/runner/client.py
+++ b/src/dstack/_internal/server/services/runner/client.py
@@ -63,7 +63,7 @@ class RunnerClient:
         resp.raise_for_status()
 
     def upload_code(self, file: Union[BinaryIO, bytes]):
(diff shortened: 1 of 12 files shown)


src/dstack/_internal/server/services/runner/client.py
```diff
@@ -17,6 +17,7 @@
 
 REMOTE_SHIM_PORT = 10998
 REMOTE_RUNNER_PORT = 10999
+REQUEST_TIMEOUT = 5
```
TheBits: I recommend increasing it to 15 seconds.
```suggestion
REQUEST_TIMEOUT = 15
```

TheBits (approved):

PR #1169, opened 2024-04-24
Fix `KeyError: 'IpPermissions'` when using AWS

### LINKED ISSUE #1168: [Bug]: `KeyError: 'IpPermissions'` when creating instance on AWS
### Steps to reproduce

Not sure, I found this exception in sentry issue DSTACK-CLOUD-25, but could not reproduce it. However, the exception suggests an easy fix regardless of reproduction steps.

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### dstack version

dstack Sky production, April 20th

### Server logs

```shell
KeyError: 'IpPermissions'
  File "dstack/_internal/server/background/tasks/process_submitted_jobs.py", line 286, in _run_job_on_new_instance
    launched_instance_info: LaunchedInstanceInfo = await run_async(
  File "dstack/_internal/server/utils/common.py", line 24, in run_async
  File "concurrent/futures/thread.py", line 58, in run
  File "dstack_cloud/services/backends/dstack/compute.py", line 70, in run_job
    launched_instance_info = backend.compute().run_job(
  File "dstack/_internal/core/backends/aws/compute.py", line 186, in run_job
    launched_instance_info = self.create_instance(instance_offer, instance_config)
  File "dstack/_internal/core/backends/aws/compute.py", line 137, in create_instance
    security_group_id=aws_resources.create_security_group(
  File "dstack/_internal/core/backends/aws/resources.py", line 156, in create_security_group
    _add_ingress_security_group_rule_if_missing(
  File "dstack/_internal/core/backends/aws/resources.py", line 351, in _add_ingress_security_group_rule_if_missing
    if _rule_exists(rule, security_group["IpPermissions"]):
```

### Additional information

_No response_

jvstme:
AWS docs say `IpPermissions` and
`IpPermissionsEgress` can be missing.
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SecurityGroup.html

It is not mentioned whether a missing property is
equivalent to an empty list, but I assume it is.
Even if my assumption is wrong, we might at least
get a clearer error next time somebody manages to
reproduce this.

Fixes #1168

### DIFF
--- a/src/dstack/_internal/core/backends/aws/resources.py
+++ b/src/dstack/_internal/core/backends/aws/resources.py
@@ -348,7 +348,7 @@ def _add_ingress_security_group_rule_if_missing(
     security_group_id: str,
     rule: Dict,
 ) -> bool:
-    if _rule_exists(rule, security_group["IpPermissions"]):
+    if _rule_exists(rule, security_group.get("IpPermissions", [])):
         return False
     ec2_client.authorize_security_group_ingress(
         GroupId=security_group_id,
@@ -363,7 +363,7 @@ def _add_egress_security_group_rule_if_missing(
     security_group_id: str,
     rule: Dict,
 ) -> bool:
-    if _rule_exists(rule, security_group["IpPermissionsEgress"]):
+    if _rule_exists(rule, security_group.get("IpPermissionsEgress", [])):
         return False
     ec2_client.authorize_security_group_egress(
         GroupId=security_group_id,


r4victor (approved):
I think the issue is that ec2_client.describe_security_groups() returns "IpPermissions" but ec2_client.create_security_group() does not.

Since a new group has no rules, this fix should work fine for now.

jvstme:
> but ec2_client.create_security_group() does not

I could not reproduce this, even when creating a security group from scratch. So I'm still not sure why this happened in production. Anyways, I think this fix can only make better, not worse

PR #1183, opened 2024-04-29
Improve the retries in the get_host_info and get_shim_healthcheck

### LINKED ISSUE #1100: Ability to add your own instances to pools/dstack

TheBits:


src/dstack/_internal/core/backends/remote/provisioning.py
```diff
@@ -125,51 +125,66 @@ def run_shim_as_systemd_service(client: paramiko.SSHClient, working_dir: str, de
...
-        err = stderr.read().strip().decode()
-        if err:
-            raise ProvisioningError(
-                f"The command 'get_shim_healthcheck' didn't work. stdout: {out}, stderr: {err}"
+    retries = 20
+    iter_delay = 3
+    for _ in range(retries):
+        try:
+            _, stdout, stderr = client.exec_command(
+                "curl -s http://localhost:10998/api/healthcheck", timeout=15
```
r4victor, line 173: Why calling the shim api like this and not using python http client (e.g. shim client)?

TheBits: My two main points are:
1. The `check_instance` will use `ShimClient` to make the request.
2. If the `runner_ssh_tunnel` gets an SSH or requests.get error, it returns `False`. It is difficult to get `HealthcheckResponse` and `HealthStatus`. For an example, look at the function `instance_healthcheck`.

r4victor: @TheBits But why using client.exec_command here? Seems error-prone

Issue #1194, opened 2024-05-03
[Feature]: Oracle Cloud Infrastructure backend


jvstme:
Add support for Oracle Cloud Infrastructure. The main priority is on-demand GPU instances. Support for other types of instances may be added later.

## Tasks

Essential, cannot release the feature without these:
- [x] backend configuration
- [x] instances catalog in gpuhunt
- [x] getting offers in dstack
- [x] packer templates
- [x] running images from the marketplace
- [x] running and terminating instances
- [x] controlling disk size
- [x] creating network resources
- [x] creating compartments
- [x] client credentials without file paths
- [x] docs

Non-essential, may be added after the release:
- [x] document restrictive permissions
- [x] GitHub actions for publishing packer images
- [x] multi-node tasks
- [x] bare metal instances

May be implemented later:
- preemptible (spot) instances #1375 
- creating dstack gateways #1376
- flexible instances - medium priority; flexible instances are CPU-only, but for newer OCI regions (e.g. Chicago) they are the only CPU VM offerings 
- always free instances - lower priority; always free instances are very small CPU instances, but they could be useful for simple experiments

r4victor:
@jvstme, should we close it as completed?

PR #1197, opened 2024-05-06
Allow configuring OCI backend

jvstme:
This is the first part of the Oracle Cloud Infrastructure backend implementation (#1194). Behind the `OCI_BACKEND` feature flag.

### DIFF
--- a/src/dstack/_internal/core/backends/oci/region.py
+++ b/src/dstack/_internal/core/backends/oci/region.py
@@ -7,7 +7,7 @@ from dstack._internal.core.backends.oci.auth import get_client_config
 from dstack._internal.core.models.backends.oci import AnyOCICreds
 
 
-class OCIRegion:
+class OCIRegionClient:
     """
     A structure for region-specific objects, including region-bound API clients
     """
@@ -30,7 +30,7 @@ class OCIRegion:
 
 def get_subscribed_region_names(creds: AnyOCICreds) -> List[str]:
     config = get_client_config(creds)
-    region = OCIRegion(config)
+    region = OCIRegionClient(config)
 
     subscriptions: List[oci.identity.models.RegionSubscription] = (
         region.identity_client.list_region_subscriptions(config["tenancy"]).data
--- a/src/dstack/_internal/core/backends/oci/region.py
+++ b/src/dstack/_internal/core/backends/oci/region.py
@@ -1,0 +1,37 @@
+from functools import cached_property
+
+import oci
+from typing_extensions import Any, List, Mapping
+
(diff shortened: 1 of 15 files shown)


jvstme:
`generate-json-schema` CI job failed. I think the job might need fixing. Will do and then reopen this PR

upd: fixed the current PR with 445d5b0 instead

src/dstack/_internal/core/models/backends/oci.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def to_client_config(self) -> Mapping[str, Any]:
+        return self.dict(exclude={"type"})
+
+
+class OCIDefaultCreds(CoreModel):
+    type: Annotated[Literal["default"], Field(description="The type of credentials")] = "default"
+
+    def to_client_config(self) -> Mapping[str, Any]:
+        return oci.config.from_file()
```
r4victor: I recommend you removing any logic from models. Any logic on top of models can be implemented as functions in core/backends/oci. E.g. https://github.com/dstackai/dstack/blob/64056913d694ed439d0d59cebfe3598683ab56fc/src/dstack/_internal/core/backends/gcp/auth.py#L19

src/dstack/_internal/server/services/backends/configurators/oci.py
```diff
@@ -0,0 +1,123 @@
...
+        try:
+            available_regions = get_subscribed_region_names(config.creds)
+        except any_oci_exception:
+            raise_invalid_credentials_error(fields=[["creds"]])
+
+        if config.regions is None:
+            config.regions = available_regions
+        elif unsubscribed_regions := set(config.regions) - set(available_regions):
+            msg = f"Regions {unsubscribed_regions} are configured but not subscribed to in OCI"
+            raise ServerClientError(msg, fields=["regions"])
```
r4victor: Fields specifies paths to bad fields, e.g. `fields=[["field1", "subfield1"], ["field2", "subfield2"]]`. So it should be `fields=[["regions"]]`.

src/dstack/_internal/core/backends/oci/region.py
```diff
@@ -0,0 +1,38 @@
+from functools import cached_property
+
+import oci
+from typing_extensions import Any, List, Mapping
+
+from dstack._internal.core.backends.oci.auth import get_client_config
+from dstack._internal.core.models.backends.oci import AnyOCICreds
+
+
+class OCIRegion:
```
r4victor: I suggest calling this class something other than just OCIRegion (e.g. OCIRegionClient) to avoid confusing it with variables for region names.

r4victor (approved):

PR #1198, opened 2024-05-06
[Internal] Install all deps in JSON schema CI job

jvstme:
It can be handy to use backend-specific
dependencies in the pydantic models used for
schema generation. See PR #1197 for an example of
the `generate-json-schema` CI job failing without
these dependencies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -156,7 +156,7 @@ jobs:
       - name: Install AWS
         run: pip install awscli
       - name: Install dstack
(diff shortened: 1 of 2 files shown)


r4victor:
There should be no backend-specific dependencies in src/dstack/_internal/core/models. At minimum, dstack won't work if those dependencies are not installed.

Users that install dstack without extra dependencies should be able to load the dstack core models. So I don't think it's necessary to install all deps for json schema generation.

PR #1215, opened 2024-05-13
Implement getting OCI offers

jvstme:
This is another part of the Oracle Cloud Infrastructure backend implementation (#1194). Behind the OCI_BACKEND feature flag.

### DIFF
--- a/src/dstack/_internal/core/backends/oci/resources.py
+++ b/src/dstack/_internal/core/backends/oci/resources.py
@@ -1,0 +1,140 @@
+from concurrent.futures import Executor, as_completed
+from itertools import islice
+from typing import Dict, Iterable, List, Mapping, Set
+
+import oci
+
+from dstack._internal.core.backends.oci.region import OCIRegionClient
+from dstack._internal.core.models.instances import InstanceOffer
+
+LIST_SHAPES_MAX_LIMIT = 100
+CAPACITY_REPORT_MAX_SHAPES = 10  # undocumented, found by experiment
+
+
+def list_shapes(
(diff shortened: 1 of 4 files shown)


src/dstack/_internal/core/backends/oci/resources.py
```diff
@@ -0,0 +1,140 @@
...
+
+    future_to_region_name = {}
+    for region_name, region_client in regions.items():
+        future = executor.submit(list_shapes, region_client.compute_client, compartment_id)
+        future_to_region_name[future] = region_name
+
+    result = {}
+    for future in as_completed(future_to_region_name):
+        region_name = future_to_region_name[future]
+        shape_names = {shape.shape for shape in future.result()}
```
TheBits, line 48: `future.result()` can throw exceptions. I suggest adding exception handling here.

r4victor (approved):

src/dstack/_internal/core/backends/oci/resources.py
jvstme: Apparently we don't do exception handling in `backends/<backend>/resources.py` for any of the backends, so it is expected that these functions can raise backend-specific exceptions and the caller is responsible for handling them.

I don't see any backend-specific exception handling in `<Backend>Compute.get_offers` either. All exceptions raised when getting offers are handled and logged by `get_offers`' caller [here](https://github.com/dstackai/dstack/blob/8b01c98c2289e26bf908cce9fcefc9277ee8eee1/src/dstack/_internal/server/services/backends/__init__.py#L337).

PR #1231, opened 2024-05-16
Implement running OCI jobs and stopping instances

jvstme:
This is part of the OCI backend Implementation (#1194). Behind the OCI_BACKEND feature flag.

- Implement `run_job`, `create_instance`, `terminate_instance`
- Temporarily allow configuring existing cloud resources with env variables. Later dstack will create or discover these resources automatically
- Check shapes quota per availability domain, not per region. Some shapes can be available in only one domain within a region

### DIFF
--- a/src/dstack/_internal/core/backends/oci/compute.py
+++ b/src/dstack/_internal/core/backends/oci/compute.py
@@ -1,17 +1,22 @@
+import json
+import os
 from concurrent.futures import ThreadPoolExecutor
-from typing import List, Optional
+from dataclasses import dataclass
+from typing import Dict, List, Optional, Set
 
-from dstack._internal.core.backends.base.compute import Compute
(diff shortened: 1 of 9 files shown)


src/dstack/_internal/core/backends/oci/compute.py
```diff
@@ -73,12 +106,77 @@ def run_job(
...
+            user=run.user,
+        )
+        return self.create_instance(instance_offer, instance_config)
 
     def terminate_instance(
         self, instance_id: str, region: str, backend_data: Optional[str] = None
     ) -> None:
-        raise NotImplementedError
+        region_client = self.regions[region]
+        region_client.compute_client.terminate_instance(
```
r4victor: Can this raise an error if instance_id not found? dstack may terminate an already deleted instance, and terminate_instance() should not raise errors in such case.

r4victor (approved):

PR #1258, opened 2024-05-21
Changed `dstack pool remove` to `rm` [Issue #1256]

### LINKED ISSUE #1256: Rename `dstack pool remove` to `dstack pool rm`

muddi900:
fixes #1256

I couldn't run any tests because I don't know the dep management situation. But the arg has been renamed.


peterschmidt85:
Thank you @muddi900! Shall we add a command alias for users that are still using the old command?

jvstme (approved):
There was some unrelated reformatting, which probably means the file was formatted with a tool other than `ruff format`, which is what we use. I reverted the reformatting just so that the diff is more concise.

Other than that, everything looks good to me. Thanks, @muddi900!

PR #1263, opened 2024-05-22
Improved network address parsing

### LINKED ISSUE #1262: [Feature]: Allow to use IP interface in the `--network` argument
### Problem

It became clear that users enter both the **IP address** and the network mask. This results in an error occurring during the parsing of the network's IP address.

The IP address of the network must be computed automatically.

### Solution

_No response_

### Workaround

_No response_

Yes

TheBits:
Fixes #1262

### DIFF
--- a/src/dstack/_internal/cli/commands/pool.py
+++ b/src/dstack/_internal/cli/commands/pool.py
@@ -296,10 +296,10 @@ class PoolCommand(APIBaseCommand):
         # validate network
         if args.network is not None:
             try:
-                network = ipaddress.ip_network(args.network)
+                network = ipaddress.IPv4Interface(args.network).network
             except ValueError as e:
                 console.print(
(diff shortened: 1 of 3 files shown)


src/dstack/_internal/cli/commands/pool.py
```diff
@@ -298,10 +298,14 @@ def _add_ssh(self, args: argparse.Namespace) -> None:
             try:
                 network = ipaddress.ip_network(args.network)
```
jvstme: The nested try-except-try statements look redundant. Consider removing this outer try.
```python
try:
    network = ipaddress.IPv4Interface(args.network).network
except ValueError as e:
    console.print(...)
    return
```

jvstme (approved):

src/dstack/_internal/cli/commands/pool.py
TheBits: Many thanks! That made code better!

PR #1276, opened 2024-05-24
Implement creating OCI resources

jvstme:
This enables dstack to automatically create OCI compartments and network resources on server launch. This is part of the OCI backend implementation (#1194). Behind the OCI_BACKEND feature flag.

### DIFF
--- a/src/dstack/_internal/core/backends/oci/compute.py
+++ b/src/dstack/_internal/core/backends/oci/compute.py
@@ -38,21 +38,17 @@ class PreConfiguredResources:
     # TODO(#1194): remove this class and teach dstack to create or discover all
     # necessary resources automatically
 
-    compartment_id: str
-    subnet_ids: Dict[str, str]
     standard_image_ids: Dict[str, str]
     cuda_image_ids: Dict[str, str]
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


r4victor (commented):
I don't see where compartment_id is validated. How is it handled if it does not exist?

r4victor (approved):

jvstme:
@r4victor, compartment_id is optional in all models, but required in OCIStoredConfig. If it is missing, a default `dstack-{project_name}` compartment will be discovered or created before instantiating OCIStoredConfig
https://github.com/dstackai/dstack/blob/92e8b3daf93610097950df649db98bdadf6ba4f1/src/dstack/_internal/server/services/backends/configurators/oci.py#L101

jvstme:
Oh, maybe I misunderstood the question. If you were asking what would happen if you specify a nonexistent comparment_id in server config — the configurator will hang for about 3 minutes and then fail with an error message [here](https://github.com/dstackai/dstack/blob/92e8b3daf93610097950df649db98bdadf6ba4f1/src/dstack/_internal/core/backends/oci/resources.py#L337). Unfortunately, I could not find a way to distinguish a nonexistent compartment from a compartment being created, so this wait is required, but specifying a wrong compartment_id should be a rare case.

PR #1281, opened 2024-05-28
Support IPv6 in OCI

jvstme:
Part of the OCI backend implementation (#1194). Behind the OCI_BACKEND feature flag.


jvstme:
We decided not to support IPv6, as it is not supported in other backends and may cause additional maintenance load. Additionally, this PR was supposed to fix slow network connections that we are experiencing in OCI, but apparently this issue is not caused but missing or misconfigured IPv6

PR #1287, opened 2024-05-30
Add TPU support in gcp

Bihan:
TPU VM provisioning is implemented on top of the existing `gcp` code. Please review the changes and provide any feedback or suggestions.

**Installation**

- In addition to Service Account User & Compute Admin roles, **TPU Admin** role should also be granted.

**Running** 

- To run a **On-Demand** TPU Device 
`dstack run . -b gcp --gpu tpu-v2-8`

- To run a **On-Spot** TPU Device
`dstack run . -b gcp --gpu tpu-v2-8 --spot`

- To run a **On-Demand** TPU Pod
`dstack run . -b gcp --gpu tpu-v2-32` # For TPU versions v2, v3 no of cores > 8 is a TPU POD. TPU versions v4, v5 are all pods.

**TPU VM & Persistent disk**
1. A TPU VM includes a 100GB **boot disk**, which is **not resizable**.
2. Additional persistent disk can be attached to TPU VM. This is already implemented in this PR. Eg
`dstack run . -b gcp --disk 120GB --gpu tpu-v2-8  ` # This attaches a persistent disk of size 120GB to TPU VM.
3. GCP Does not provide clear instruction on how to configure auto deletion of persistent disk after
termination of the associated TPU VM. Due to this there are two issues in the current implementation. The first one is solved, and for the second reviewer's suggestion is required.

   3.1 During **spot provisioning** when submission is not successful.
This case is solved by catching the exception in GCPCompute.create_instance() method and deleting the associated persistent disk.

   3.2 During **spot provisioning** when submission is successful, but provisioning fails.
Is there a method (callback, lifecycle hook) which is invoked when run status changes to `FAILED`? Or any other suggestion?

4. To prevent two or more TPU VMs from a Pod from writing to a Persistent Disk at once, all Persistent Disks attached to a TPU VM in a Pod must be configured as read-only. So in our implementation if TPU VM is a device, then it is read-write and if it is a pod, then it is read-only.


r4victor:
> In addition to Service Account User & Compute Admin roles, TPU Admin role should also be granted.

dstack no longer uses coarse roles but only the most granular cloud permissions:

https://github.com/dstackai/dstack/blob/7a2d58c01b1b76055b8a3505a0db1d63337c1001/docs/docs/reference/server/config.yml.md#L322-L341

@Bihan, please document minimal required TPU-related permissions.

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -69,11 +73,12 @@ def get_offers(
...
             if _has_gpu_quota(quotas[region], offer.instance.resources):
                 availability = InstanceAvailability.UNKNOWN
-            # todo quotas: cpu, memory, global gpu
+            # todo quotas: cpu, memory, global gpu, tpu
             offers_with_availability.append(
                 InstanceOfferWithAvailability(**offer.dict(), availability=availability)
             )
-            offers_with_availability[-1].region = region
+            if not _contains_tpu(requirements):
+                offers_with_availability[-1].region = region
```
r4victor: Why do we show availability zone as the TPU region instead of region?

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -116,7 +136,68 @@ def create_instance(
...
+        startup_script = "#! /bin/bash\n" + startup_script
+        instance_id = f"tpu-{instance_config.instance_name}"
+        if "tpu" in instance_offer.instance.resources.gpus[0].name:
+            tpu_node = gcp_resources.create_tpu_node_struct(
+                instance_name=instance_offer.instance.name,
+                startup_script=startup_script,
+                authorized_keys=authorized_keys,
+                spot=instance_offer.instance.resources.spot,
+                project_id=self.config.project_id,
+                zone=instance_offer.region,
```
r4victor: For regular VMs, we try provisioning in all zones in the region. Why trying only one zone for TPUs?

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -116,7 +136,68 @@ def create_instance(
             project_id=self.config.project_id,
         )
         disk_size = round(instance_offer.instance.resources.disk.size_mib / 1024)
+        commands = get_shim_commands(authorized_keys=authorized_keys)
+        startup_script = " ".join([" && ".join(commands)])
+        startup_script = "#! /bin/bash\n" + startup_script
+        instance_id = f"tpu-{instance_config.instance_name}"
+        if "tpu" in instance_offer.instance.resources.gpus[0].name:
+            tpu_node = gcp_resources.create_tpu_node_struct(
```
r4victor: create_tpu_node_struct() name is misleading since it not only creates a struct but also does a network call to create a persistent disk. So it may raise GCP errors (which are not handled). I suggest creating persisting disk outside of create_tpu_node_struct() and passing the disk reference there.

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -116,7 +136,68 @@ def create_instance(
...
+                gcp_resources.wait_for_operation(operation, verbose_name="tpu instance creation")
+            except (
+                google.api_core.exceptions.ServiceUnavailable,
+                google.api_core.exceptions.NotFound,
+                google.api_core.exceptions.ResourceExhausted,
+            ):
+                # There is no clear instruction on how to configure auto delete of persistent disk on
+                # termination of associated TPU VM. GCP does not provide documentation on
+                # how to get the vm instance name of the TPU VM. It only provides node name.
+                if gcp_resources.is_disk_present(
```
r4victor: Seems there is no reason to check for disk existence. Just delete it and ignore if it does not exist. Saves a network call.

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -87,9 +92,24 @@ def terminate_instance(
...
-            self.instances_client.delete(
-                project=self.config.project_id, zone=zone, instance=instance_id
-            )
+            if "tpu" in instance_id:
+                name = f"projects/{self.project_id}/locations/{zone}/nodes/{instance_id}"
+                delete_request = tpu_v2.DeleteNodeRequest(
+                    name=name,
+                )
+                operation = self.tpu_client.delete_node(request=delete_request)
+                while not operation.done():
```
r4victor: Why not gcp_resources.wait_for_operation()?

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -87,9 +92,24 @@ def terminate_instance(
             backend_data_dict = json.loads(backend_data)
             zone = backend_data_dict["zone"]
         try:
-            self.instances_client.delete(
-                project=self.config.project_id, zone=zone, instance=instance_id
-            )
+            if "tpu" in instance_id:
```
r4victor: If my instance name has "tpu" in the name but it's not tpu, it will fail. You can store something in backend_data to indicate that the instance is TPU.

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -116,7 +136,68 @@ def create_instance(
...
+                internal_ip=None,
+                region=instance_offer.region,
+                price=instance_offer.price,
+                ssh_port=22,
+                username="ubuntu",
+                ssh_proxy=None,
+                dockerized=True,
+                backend_data=None,
+            )
         labels = {
```
r4victor: TPUs should be labeled as well. I believe they support labels, right?

r4victor:
@Bihan,

> 3.2 During spot provisioning when submission is successful, but provisioning fails.

I'm not sure I understood the problem. Could you please elaborate on this case? What do you mean by successful submission and failed provisioning?

src/dstack/_internal/core/backends/gcp/compute.py
Bihan: Ok. Now I got the concept of zone and region. Will fix it accordingly.

src/dstack/_internal/core/backends/gcp/compute.py
Bihan: Same as above. I will fix it.

src/dstack/_internal/core/backends/gcp/compute.py
Bihan: I will use gcp_resources.wait_for_operation(). This will make it consistent.

src/dstack/_internal/core/backends/gcp/compute.py
Bihan: Yes I found out TPUs does support labels. Will fix it.

Bihan:
> @Bihan,
> 
> > 3.2 During spot provisioning when submission is successful, but provisioning fails.
> 
> I'm not sure I understood the problem. Could you please elaborate on this case? What do you mean by successful submission and failed provisioning?

@r4victor I mean sometimes spot TPU VM can be interrupted by GCP during our provisioning process. You can reproduce the issue with following steps
1.   `dstack run . -b gcp --gpu tpu-v2-8`
2.    `Continue? [y/n]: y`
3. During below stage, which is after submitted stage. Manually delete TPU VM instance from console. 
```
⠏ Launching ordinary-gecko-1 (provisioning)
  - ordinary-gecko-1-0-0 (provisioning)
```
4. The run ends with below information
`Run failed with error code INTERRUPTED_BY_NO_CAPACITY. Check CLI and server logs for more details.`

I wanted to ask how can I  delete the attached disk after above failed state.

r4victor:
I got it. This case is not special to TPUs with attached disks. An interrupted job will be marked as failed and release the instance. The server will then fail to ping the instance and after some period (20m) terminate it by calling `backend.compute().terminate_instance`. So the disk won't be left hanging if you delete it in `terminate_instance`.

See:
https://github.com/dstackai/dstack/blob/a23e29c7db16a6c04ce9578d8698253a4085242a/src/dstack/_internal/server/background/tasks/process_instances.py#L702

r4victor:
@Bihan, I don't quite understand how users are supposed to work with attached disks. Have you tested them? To use the disks, they should be mounted in each TPU. It seems like dstack doesn't do this. Moreover, the disks should be formatted first and this is something that cannot be done if the disk is read-only. Even if dstack implements all of this, persistent disks are only suitable as read-only storage for TPU Pods. If you need more storage for writing, you'll need to use [other options](https://cloud.google.com/tpu/docs/storage-options).

It seems that proper storage options for TPU Pods are either Cloud Storage or Filestore depending on the use case. My proposal is to support only bootable 100GB disks for TPUs for now (to avoid complicated attached disks setup), and support Cloud Storage and/or Filestore later.

Bihan:
> @Bihan, I don't quite understand how users are supposed to work with attached disks. Have you tested them? To use the disks, they should be mounted in each TPU. It seems like dstack doesn't do this. Moreover, the disks should be formatted first and this is something that cannot be done if the disk is read-only. Even if dstack implements all of this, persistent disks are only suitable as read-only storage for TPU Pods. If you need more storage for writing, you'll need to use [other options](https://cloud.google.com/tpu/docs/storage-options).
> 
> It seems that proper storage options for TPU Pods are either Cloud Storage or Filestore depending on the use case. My proposal is to support only bootable 100GB disks for TPUs for now (to avoid complicated attached disks setup), and support Cloud Storage and/or Filestore later.

@r4victor Yes to use they should be mounted in each TPU manually by the user. I understood your point on the usage of storage for TPU Pods. For now I shall modify the implementation to support only the bootable 100GB disk.

src/dstack/_internal/core/backends/gcp/resources.py
```diff
@@ -226,3 +228,32 @@ def generate_random_resource_name() -> str:
...
+    node.accelerator_type = instance_name
+    node.runtime_version = "tpu-ubuntu2204-base"
+    node.network_config = tpu_v2.NetworkConfig(enable_external_ips=True)
+    ssh_keys = "\n".join(f"ubuntu:{key}" for key in authorized_keys)
+    node.metadata = {"ssh-keys": ssh_keys, "startup-script": startup_script}
+    node.labels = labels
+    return node
+
+
+def wait_for_operation(operation: Operation, verbose_name: str = "operation", timeout: int = 300):
```
r4victor: @Bihan, Any reason to define new wait_for_operation() vs using existing wait_for_extended_operation()? I suppose I mentioned wait_for_operation() in the review comment but I actually meant using existing wait_for_extended_operation().

As of now, wait_for_operation() seems incorrect since it will return result in case of an error, which won't be handled.

Bihan: @r4victor I defined a new wait_for_operation() because `self.tpu_client.create_node(request=create_node_request)` returns type `Operation`. The type `Operation` does not have field `error_code` like type `ExtendedOperation` used in wait_for_extended_operation(). 

To check the error handling I set default timeout to 3s and it resulted as below 
`[21:04:49] ERROR    dstack._internal.core.backends.gcp.resources:256 Error during tpu instance creation: Operation did not complete within the designated timeout
                    of 3 seconds.                                                                                                                                
           ERROR    dstack._internal.core.backends.gcp.resources:257 Operation ID: <google.api_core.operation.Operation object at 0x129af3710>    
 RuntimeError: Operation did not complete within the designated timeout of 3 seconds.                                                         
           DEBUG    dstack._internal.server.background.tasks.process_submitted_jobs:159 job(1e2ec0)witty-stingray-1-0-0: provisioning failed                     
           INFO     dstack._internal.server.background.tasks.process_runs:333 run(f96c86)witty-stingray-1: run status has changed SUBMITTED -> TERMINATING       
           INFO     dstack._internal.server.services.jobs:265 job(1e2ec0)witty-stingray-1-0-0: job status is FAILED, reason: FAILED_TO_START_DUE_TO_NO_CAPACITY  
[21:16:34] INFO     dstack._internal.server.services.runs:821 run(f96c86)witty-stingray-1: run status has changed TERMINATING -> FAILED, reason: JOB_FAILED   `

Any suggestions are welcome.

r4victor: Ah, I see now. Then it should be ok.

PR #1289, opened 2024-05-30
More flexible OCI credentials configuration

jvstme:
- In client credentials, replace the key path with the key content to allow configuring without any file paths
- In default credentials, allow overriding the config file path and choosing a profile from the config file

This is part of the OCI backend implementation (#1194). Behind the OCI_BACKEND feature flag.

### DIFF
--- a/src/dstack/_internal/core/models/backends/oci.py
+++ b/src/dstack/_internal/core/models/backends/oci.py
@@ -18,8 +18,8 @@ class OCIClientCreds(CoreModel):
     type: Annotated[Literal["client"], Field(description="The type of credentials")] = "client"
     user: Annotated[str, Field(description="User OCID")]
     tenancy: Annotated[str, Field(description="Tenancy OCID")]
-    key_file: Annotated[Path, Field(description="Path to the PEM key")]
-    fingerprint: Annotated[str, Field(description="Key fingerprint")]
+    key_content: Annotated[str, Field(description="Content of the user's private PEM key")]
+    fingerprint: Annotated[str, Field(description="User's public key fingerprint")]
     region: Annotated[
         str, Field(description="Name or key of any region the tenancy is subscribed to")
     ]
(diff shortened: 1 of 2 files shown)


src/dstack/_internal/core/models/backends/oci.py
```diff
@@ -18,15 +18,19 @@ class OCIClientCreds(CoreModel):
     type: Annotated[Literal["client"], Field(description="The type of credentials")] = "client"
     user: Annotated[str, Field(description="User OCID")]
     tenancy: Annotated[str, Field(description="Tenancy OCID")]
-    key_file: Annotated[Path, Field(description="Path to the PEM key")]
-    fingerprint: Annotated[str, Field(description="Key fingerprint")]
+    key_content: Annotated[str, Field(description="Content of the user's private PEM key")]
```
r4victor, line 21: Does it mean specifying the contents of the key file is the only way to set up client creds? Why not allow specifying path or content as we do for other providers?

jvstme: Well, this would be yet another way to configure the OCI backend and I'm not sure what it would contribute apart from code and documentation complexity. I think what we have now should cover all the use cases.

If the user doesn't want to use files, they can use client credentials and specify the raw key content. If they do want to use files, they can use default credentials, i.e. OCI CLI-compatible config file, optionally specifying the path to this file.

jvstme: In other words, the OCI alternative for
```yaml
- type: gcp
  project_id: gcp-project-id
  creds:
    type: service_account
    filename: ~/.dstack/server/gcp-024ed630eab5.json
```
is
```yaml
- type: oci
  creds:
    type: default
    file: ~/.dstack/server/oci-config
```

r4victor (approved):

PR #1306, opened 2024-06-04
[Docs] Fix Azure config section layout

jvstme:
The notes for subscription_id and required permissions are useful for any type of credentials, but were only visible if the "Default credentials" tab was open.


jvstme:
The diff is a bit unclear on GitHub. This PR does not change any text, it only removes indentation.

peterschmidt85 (approved):

PR #1333, opened 2024-06-13
Respect registry_auth for RunPod

### LINKED ISSUE #1297: [Bug]: Private registry_auth does not seem to be respected for Runpod deployments
### Steps to reproduce

1. example dstack configuration, `example.dstack.yml`:
```
type: task

image: <private docker hub repo>/example:latest
registry_auth:
  username: <private docker hub username>
  password: <private docker hub password/token>

backends: [runpod]
```

2. `dstack run . -f example.dstack.yml` works and starts provisioning
3. The pod is created in Runpod as expected
4. The pod fails to pull the referenced image with ` pull access denied for <projectname>, repository does not exist or may require 'docker login' `

### Actual behaviour

Screenshot of logs from runpod:
image

### Expected behaviour

_No response_

### dstack version

0.18.2

### Server logs

_No response_

### Additional information

_No response_

smokfyz:
Fixes #1297 

Add support for registry_auth for the RunPod and VastAI backends. RunPod does not have the feature to pass credentials directly to the create pod API. Instead, there is a containerRegistryAuths object that should be created and passed to the editPod API to use private registries. To remove stale objects, a function was implemented to delete stale objects every 24 hours during some user request executions.

This solution is considered a temporary workaround because it could have issues when running at scale. The containerRegistryAuth only supports querying all objects at once, and only one containerRegistryAuth can be deleted per API call. Therefore, if there are many users, it can become a bottleneck.

### DIFF
--- a/src/dstack/_internal/core/backends/runpod/compute.py
+++ b/src/dstack/_internal/core/backends/runpod/compute.py
@@ -1,6 +1,6 @@
 import json
 import uuid
-from datetime import datetime, timedelta, timezone
+from datetime import timedelta
 from typing import List, Optional
 
 from dstack._internal.core.backends.base import Compute
@@ -27,12 +27,10 @@ from dstack._internal.utils.logging import get_logger
 
 logger = get_logger(__name__)
 
-CONTAINER_REGISTRY_AUTH_DELETE_TIMEOUT = 60  # 60 seconds
-CONTAINER_REGISTRY_AUTH_CLEANUP_INTERVAL = 60 * 60 * 24  # 1 day
+CONTAINER_REGISTRY_AUTH_CLEANUP_INTERVAL = 60 * 60 * 24  # 24 hour
 
 
 class RunpodCompute(Compute):
-    _instance_id_to_container_registry_auth_id_mapping = {}
     _last_cleanup_time = None
 
     def __init__(self, config):
@@ -103,9 +101,6 @@ class RunpodCompute(Compute):
                 container_disk_in_gb=disk_size,
                 container_registry_auth_id=container_registry_auth_id,
             )
-            self._instance_id_to_container_registry_auth_id_mapping[instance_id] = (
-                container_registry_auth_id
-            )
 
         if (
             self._last_cleanup_time is None
@@ -135,7 +130,6 @@ class RunpodCompute(Compute):
     ) -> None:
         try:
             self.api_client.terminate_pod(instance_id)
-            self._delete_container_registry_auth_for_instance_id(instance_id)
         except BackendError as e:
             if e.args[0] == "Instance Not Found":
                 logger.debug("The instance with name %s not found", instance_id)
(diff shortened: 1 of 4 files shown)


smokfyz:
The RunPod "create pod" API does not support setting up Docker registry credentials directly. Therefore, I implemented a workaround using the container registry auth API to dynamically create and delete authentication credentials. However, during testing of the periodic deletion of stale records, I encountered an issue where the API was unable to delete some old credentials for unknown reasons. Therefore, this approach is not suitable for production use at this time. I will contact RunPod support for assistance.

src/dstack/_internal/core/backends/runpod/compute.py
```diff
@@ -120,6 +149,27 @@ def update_provisioning_data(self, provisioning_data: JobProvisioningData) -> No
...
+    def _clean_stale_container_registry_auths(self) -> None:
+        container_registry_auths = self.api_client.get_container_registry_auths()
+
+        # Container_registry_auths sorted by creation time so try to delete the oldest first
+        # when we reach container_registry_auths that is still in use, we stop
+        for container_registry_auth in container_registry_auths:
+            try:
+                self.api_client.delete_container_registry_auth(container_registry_auth["id"])
+            except Exception:
+                break
```
r4victor, line 175: I don't see a good reason to stop at the first failed delete. Maybe go over all the entries ignoring those that are in use?

src/dstack/_internal/core/backends/runpod/compute.py
```diff
@@ -80,7 +91,25 @@ def run_job(
             ports="10022/tcp",
             bid_per_gpu=instance_offer.price if instance_offer.instance.resources.spot else None,
         )
+
         instance_id = resp["id"]
+
+        if container_registry_auth_id is not None:
+            instance_id = self.api_client.edit_pod(
```
r4victor, line 102: Please leave a comment on why this "edit" approach is implemented and possibly TODO to replace it once runpod updates the API.

r4victor (approved):

src/dstack/_internal/core/backends/runpod/compute.py
smokfyz: You are right, it's better to leave a comment. I have added it.

src/dstack/_internal/core/backends/runpod/compute.py
smokfyz: It doesn't make sense because if we encounter an error, it likely means that many of the following records are also still being used by some pods (because they are sorted by creation time). Therefore, attempting to delete these records would be unnecessary work.

smokfyz: > ignoring those that are in use

We can't know if they are in use without trying to delete them because there is a time window between pod termination and the unlocking of containerRegistryAuth for deletion.

PR #1334, opened 2024-06-13
Support multi-node tasks on OCI

jvstme:
This commit adds support for multi-node tasks on OCI by allowing network traffic between instances in the same project and region. To achieve that, a network security group is created or discovered on instance launch and instance's OS-level firewall is reconfigured on boot via cloud-init.

#1194

### DIFF
--- a/src/dstack/_internal/core/backends/oci/resources.py
+++ b/src/dstack/_internal/core/backends/oci/resources.py
@@ -559,6 +559,8 @@ def get_or_create_security_group(
 def update_security_group_rules_for_runner_instances(
     security_group_id: str, client: oci.core.VirtualNetworkClient
 ) -> None:
+    # These rules are combined with subnet's default Security List that allows
+    # ingress TCP on port 22 from anywhere
     rules = [
         SecurityRule(
             description="Allow all traffic within this security group",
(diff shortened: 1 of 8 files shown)


src/dstack/_internal/core/backends/oci/resources.py
```diff
@@ -502,3 +539,89 @@ def update_route_table(
...
+    rules = [
+        SecurityRule(
+            description="Allow all traffic within this security group",
+            direction=oci.core.models.AddSecurityRuleDetails.DIRECTION_INGRESS,
+            source_type=oci.core.models.AddSecurityRuleDetails.SOURCE_TYPE_NETWORK_SECURITY_GROUP,
+            source=security_group_id,
+            protocol="all",
+        ),
+    ]
+    update_security_group_rules(security_group_id, rules, client)
```
r4victor: What allows ingress traffic on port 22? Is it allowed by default in OCI? Won't it be deleted when we remove all existing security groups? Please leave a comment if it's not obvious.

r4victor (approved):

src/dstack/_internal/core/backends/oci/resources.py
jvstme: Yes, SSH is allowed by subnet's default security list. Security lists are separate from security groups, actions on security groups do not affect the security list

PR #1354, opened 2024-06-24
tpu initial release

Bihan:
**Fixes**
1. TPU is detected from actual workloads
`--privileged` flag added to docker run
2. Env variable `PJRT_DEVICE` set to TPU
This is necessary otherwise Warning is issued while running training script.
3. Ensure all single-VM TPUs can be used
v2, v3, v4, v5p, v5litepod are different versions of TPUs provides by GCP. Except v4, all TPU versions containing number v2-{number} where number <= 8 is a single VM TPU. v4 version are all TPU Pods
Note: To use TPU version v4, v5p and v5litepod quotas should be requested.
4. dstack-runner automatically sets env LD_LIBRARY_PATH in the container

**Before Starting the Test**
1. Since there is modification in dstack-shim and dstack-runner ensure that latest dstack-shim-linux-amd64 and dstack-runner binary is used.

**How to test TPU** 
1. `% dstack run . -b gcp --gpu tpu-v2-8`
2. After provisioning is complete set env LD_LIBRARY_PATH in the container as below
`(workflow) root@t1v-n-345435e8-w-0:~ export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$(python3-config --prefix)/lib"`
3. `(workflow) root@t1v-n-345435e8-w-0:~ pip install torch~=2.3.0 torch_xla[tpu]~=2.3.0 torchvision -f https://storage.googleapis.com/libtpu-releases/index.html`
4. `(workflow) root@t1v-n-345435e8-w-0:~ git clone --recursive https://github.com/pytorch/xla.git`
5. `(workflow) root@t1v-n-345435e8-w-0:~ python3 xla/test/test_train_mp_imagenet.py --fake_data --model=resnet50 --num_epochs=1`

**Test Using Task**

```type: task

python: "3.11"


commands:
  - pip install torch~=2.3.0 torch_xla[tpu]~=2.3.0 torchvision -f https://storage.googleapis.com/libtpu-releases/index.html
  - git clone --recursive https://github.com/pytorch/xla.git
  - python3 xla/test/test_train_mp_imagenet.py --fake_data --model=resnet50 --num_epochs=1

# (Optional) Configure `gpu`, `memory`, `disk`, etc
resources:
  gpu:  tpu-v2-8

### DIFF
--- a/runner/internal/shim/docker.go
+++ b/runner/internal/shim/docker.go
@@ -318,11 +318,11 @@ func createContainer(ctx context.Context, client docker.APIClient, runnerDir str
 		Entrypoint:   []string{"/bin/sh", "-c"},
 		ExposedPorts: exposePorts(dockerParams.DockerPorts()...),
 		Env: []string{
-			"PJRT_DEVICE=TPU",
+			fmt.Sprintf("PJRT_DEVICE=%s", dockerParams.DockerPJRTDevice()),
 		},
 	}
 	hostConfig := &container.HostConfig{
-	    Privileged:      dockerParams.DockerPrivileged(),
+		Privileged:      dockerParams.DockerPrivileged(),
 		NetworkMode:     getNetworkMode(),
 		PortBindings:    bindPorts(dockerParams.DockerPorts()...),
 		PublishAllPorts: true,
@@ -432,6 +432,10 @@ func (c CLIArgs) DockerKeepContainer() bool {
 
 func (c CLIArgs) DockerPrivileged() bool {
 	return c.Docker.Privileged
+}
+
+func (c CLIArgs) DockerPJRTDevice() string {
+	return c.Docker.PJRTDevice
 }
 
 func (c CLIArgs) DockerShellCommands(publicKeys []string) []string {
--- a/runner/internal/shim/docker.go
+++ b/runner/internal/shim/docker.go
@@ -312,14 +312,18 @@ func createContainer(ctx context.Context, client docker.APIClient, runnerDir str
 		return "", tracerr.Wrap(err)
 	}
 
+	//Set the environment variables
+	envVars := []string{}
+	if dockerParams.DockerPJRTDevice() != "" {
+		envVars = append(envVars, fmt.Sprintf("PJRT_DEVICE=%s", dockerParams.DockerPJRTDevice()))
+	}
+
 	containerConfig := &container.Config{
 		Image:        taskConfig.ImageName,
 		Cmd:          []string{strings.Join(dockerParams.DockerShellCommands(taskConfig.PublicKeys), " && ")},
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: runner/internal/shim/docker.go [hunk 1]
```diff
@@ -317,8 +317,12 @@ func createContainer(ctx context.Context, client docker.APIClient, runnerDir str
 		Cmd:          []string{strings.Join(dockerParams.DockerShellCommands(taskConfig.PublicKeys), " && ")},
 		Entrypoint:   []string{"/bin/sh", "-c"},
 		ExposedPorts: exposePorts(dockerParams.DockerPorts()...),
+		Env: []string{
+			"PJRT_DEVICE=TPU",
```
r4victor: Does it mean we always set PJRT_DEVICE=TPU? So even when not running TPU but CUDA? I think we should set PJRT_DEVICE=TPU when running on TPUs only.

Bihan: @r4victor Should I set it as optional flag in docker run  something similar to privileged_flag below?
`nohup dstack-shim {dev_flag} docker --keep-container {privileged_flag} {pjrt_device}>{DSTACK_WORKING_DIR}/shim.log 2>&1 &`

r4victor: I'd set PJRT_DEVICE in the runner where we execute the job and set other envs:

https://github.com/dstackai/dstack/blob/62d25c3029c191dc2d7c7d7f50e09e413d93c1d0/runner/internal/executor/executor.go#L196-L207

This is good because we avoid introducing another place where we set envs. This would require passing whether tpu is used or not to the runner API but this should not be hard to do.

Setting PJRT_DEVICE via shim arg would probably work as well. Feel free to go this route if it works.

Bihan: @r4victor "This is good because we avoid introducing another place where we set envs." I think is a very valid point. I will set in the runner's jobEnvs

runner/internal/shim/docker.go
```diff
@@ -317,8 +317,12 @@ func createContainer(ctx context.Context, client docker.APIClient, runnerDir str
 		Cmd:          []string{strings.Join(dockerParams.DockerShellCommands(taskConfig.PublicKeys), " && ")},
 		Entrypoint:   []string{"/bin/sh", "-c"},
 		ExposedPorts: exposePorts(dockerParams.DockerPorts()...),
+		Env: []string{
+			fmt.Sprintf("PJRT_DEVICE=%s", dockerParams.DockerPJRTDevice()),
```
r4victor: If pjrt-device arg is not set, we still set PJRT_DEVICE="", which may not be the same as not setting PJRT_DEVICE. Let's add PJRT_DEVICE to Env only if DockerPJRTDevice() is not empty string?

r4victor (approved):

PR #1364, opened 2024-06-27
Support dstack volumes

### LINKED ISSUE #1158: [Feature]: Support dstack volumes
### Problem

Currently, dstack lacks built-in functionality that would allow users to persist data between runs. Cloud providers usually provide data persistence via network volumes. The proposal is to introduce volumes to dstack.

* There should be a way to register an existing volume with dstack such as an existing EBS volume (aka external volumes)
* Users may not need to create volumes themselves, so dstack should be able to create and manage volumes (aka dstack-managed volumes).

### Solution

Users will manage volumes via `dstack apply`. The new `dstack apply` configuration of type `volume` is to be introduced. Here's some examples of `volume` configurations:

* AWS EBS dstack-managed volume
  ```
  type: volume
  name: my-dstack-aws-volume
  backend: aws
  region: eu-west-1
  size: 500GB
  ```

* AWS EBS external volume
  ```
  type: volume
  name: my-external-aws-volume
  backend: aws
  region: eu-west-1
  volume_id: vol23at3432
  ```

Once a volume is already added to dstack, it can be mounted in a run like this:

```
type: task
commands:
  - ...
volumes:
  - name: volume-name
    path: /dstack_data
```

dstack will try to provision the instance in the backend/region of that volume. In case of no availability, the run will fail – other backends/regions won’t be tried since the specified volume cannot be mounted there.

### Scope and Future plans

We'll start by implementing volumes support for AWS backend. Other backends such as GCP, Azure, OCI, runpod are likely to follow.

dstack will implement default volume configurations for all backends to allow for simple volume configurations as above. Users may also need to specify advanced configurations to make use of backend-specific features. This can be done by introducing a `spec` property of different types for different volumes storage options such as EBS/Persistent Disks/HyperDisks/etc:

```
type: volume
name: my-dstack-aws-advanced-volume
backend: aws
region: eu-west-1
size: 500GB
spec:
  type: ebs
  volume_type: io2
  iops: 256000
  az: eu-west-1a
```

The following is not planned for initial volume release but may be added later:

* Volumes pricing.
* Volumes support in dstack Sky (requires pricing).
* Different volume kinds such as GCP’s Hyperdisks and Persistent Disks. 
* Volumes support for Tensordock and VastAI. These backends have no explicit support for volumes. Persistence can only be achieved via instance stopping/restarting.
* Volumes support for Kubernetes. Not clear which volumes types to support and how to do that.
* Attaching volumes to multiple instances.

r4victor:
Closes #1158 

This PR adds support for network volumes in dstack.

* Introduces `volume` configuration.
* Implements `dstack apply`, `dstack delete`, `dstack volume list` for volumes.
* Adds async API for managing volumes and background processing task to process volumes.
* Extends shim capabilities to format and mount volumes on the host and mount volumes in the container.
* Extends Compute interface to support volumes.
* Implements dstack-managed and external volumes for AWS.
* Updates docs reference with volumes.
* MInor refactoring.

Notes:
* Both external and dstack-managed volumes are supported.
* Attaching multiple volumes per run is supported.
* Volumes can be used with instances in pools – they are attached/detached automatically with job start/end.
* There is no way to see volume attachments/mount via CLI.
* No way to manually attach/detach volumes.

Test volumes:

1. Create a volume with `dstack apply`:

```
type: volume
name: my-dstack-aws-volume
backend: aws
region: eu-central-1
size: 100GB
```

2. Run a dev environment with the volume attached:

```
type: dev-environment
ide: vscode
volumes:
  - name: my-dstack-aws-volume
    path: /my_volume_data
```

3. Save data on the volume, stop the run, run the dev env again and see the volume data persist.


.github/workflows/build.yml
jvstme, line 1: Is it necessary to disable these? Sometimes they can prevent bugs, as happened [here](https://github.com/dstackai/dstack/pull/1197#issuecomment-2095513614)

docs/docs/reference/dstack.yml/task.md
```diff
@@ -357,3 +357,11 @@ The `task` configuration type supports many other options. See below.
       show_root_heading: false
       type:
         required: true
+
+## `volumes`
```
jvstme:
```suggestion
## `volumes[n]`
```

runner/internal/shim/docker.go
```diff
@@ -200,6 +217,114 @@ func (d DockerRunner) GetState() (RunnerStatus, ContainerStatus, string, JobResu
...
+	}
+	err = mountDisk(deviceName, getVolumeMountPoint(volume.Name))
+	if err != nil {
+		return tracerr.Wrap(err)
+	}
+	return nil
+}
+
+func getVolumeMountPoint(volumeName string) string {
+	return fmt.Sprintf("/%s", volumeName)
```
jvstme: Is it the user-provided volume name? In that case mounting in the fs root can lead to name collisions, e.g. if the volume name is `var` or `etc`. I suggest mounting in a directory unique to dstack, e.g. `/media/dstack-volumes/<volume-name>`

runner/internal/shim/docker.go
```diff
@@ -428,6 +560,26 @@ func requestGpuIfAvailable(ctx context.Context, client docker.APIClient) ([]cont
...
+			}
+		}
+		mounts = append(mounts, mount.Mount{Type: mount.TypeBind, Source: source, Target: mountPoint.Path})
+	}
+	return mounts, nil
+}
+
+func getVolumeSource(volumeName string) string {
+	// Nest volume data so that users don't see FS-specific files like lost+foind dir
+	return fmt.Sprintf("/%s/data", volumeName)
```
jvstme: Does it mean users will have to put their data into `/data` in external volumes too? I think this won't be acceptable to many users.

I also think seeing the lost+found dir is not an issue. On the contrary, this dir is created so that users can act on it.

So I would suggest removing the `/data` prefix.

src/dstack/_internal/cli/commands/run.py
```diff
@@ -254,29 +254,29 @@ def _print_finished_message(run: Run):
...
-            f"Error: {run._run.jobs[0].job_submissions[0].termination_reason_message} "
+            "Cannot create container.\n"
+            f"Error: {termination_reason_message}\n"
             "Check CLI and server logs for more details."
         )
     elif termination_reason == JobTerminationReason.EXECUTOR_ERROR:
         message = (
-            f"Error: {run._run.jobs[0].job_submissions[0].termination_reason_message} "
-            "Check CLI and server logs for more details."
+            f"Error: {termination_reason_message}\n" "Check CLI and server logs for more details."
```
jvstme: (nit)
```suggestion
            f"Error: {termination_reason_message}\nCheck CLI and server logs for more details."
```

src/dstack/_internal/cli/commands/volume.py
```diff
@@ -0,0 +1,30 @@
...
+    def _register(self):
+        super()._register()
+        self._parser.set_defaults(subfunc=self._list)
+        subparsers = self._parser.add_subparsers(dest="action")
+
+        list_parser = subparsers.add_parser(
+            "list", help="List volumes", formatter_class=self._parser.formatter_class
+        )
+        list_parser.add_argument(
+            "-v", "--verbose", action="store_true", help="Show more information"
```
jvstme, line 20: This flag does not really change the output, which is a bit confusing. Maybe consider removing

src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -96,6 +110,41 @@ def terminate_gateway(
...
+
+    def delete_volume(self, volume: Volume):
+        """
+        Deletes a volume.
+        """
+        raise NotImplementedError()
+
+    def attach_volume(self, volume: Volume, instance_id: str) -> VolumeAttachmentData:
+        """
+        Attaches a volumes to the instance.
```
jvstme:
```suggestion
        Attaches a volume to the instance.
```

src/dstack/_internal/core/models/volumes.py
```diff
@@ -0,0 +1,60 @@
...
+    volume_id: Annotated[
+        Optional[str],
+        Field(description="The volume ID. Must be specified when registering external volumes"),
+    ] = None
+
+
+class VolumeProvisioningData(CoreModel):
+    volume_id: str
+    size_gb: int
+    availability_zone: Optional[str] = None
```
jvstme, line 38: (nit) I'd make a required field in `AWSVolumeBackendData` instead of the optional field here. Better for type safety

src/dstack/_internal/core/backends/aws/compute.py
```diff
@@ -361,20 +372,130 @@ def terminate_gateway(
...
 
-def _supported_instances(offer: InstanceOffer) -> bool:
-    for family in ["t2.small", "c5.", "m5.", "p3.", "g5.", "g4dn.", "p4d.", "p4de."]:
-        if offer.instance.name.startswith(family):
-            return True
-    return False
+        tags = [
+            {"Key": "Name", "Value": volume.configuration.name},
+            {"Key": "owner", "Value": "dstack"},
+            # {"Key": "dstack_project", "Value": configuration.project_name},
```
jvstme, line 419: Commenting out looks unintentional

src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -413,59 +430,34 @@ def _process_pulling_with_shim(
...
-            )
-            logger.debug("runner healthcheck: %s", container_status.dict())
-            job_model.termination_reason = JobTerminationReason.CREATING_CONTAINER_ERROR
-            job_model.termination_reason_message = (
-                f"Failed to run docker pull or docker create: {container_status.error}"
-            )
-            return False
-        return True  # runner is not available yet, but shim is alive (pulling)
-
+    # shim_status.state == "running"
```
jvstme: Why is it commented out?

src/dstack/_internal/server/background/tasks/process_submitted_jobs.py
```diff
@@ -120,74 +132,103 @@ async def _process_submitted_job(session: AsyncSession, job_model: JobModel):
...
+        await session.commit()
+        return
+
     # Try to provision on an instance from the pool
     pool = await get_or_create_pool_by_name(
         session=session,
         project=project_model,
         pool_name=profile.pool_name,
     )
+    # TODO: Take volumes into account when picking instances
```
jvstme: I think it's already done, right?
```suggestion
```

src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -630,6 +634,15 @@ def gateway_model_to_gateway(gateway_model: GatewayModel) -> Gateway:
...
 
 def _validate_gateway_configuration(configuration: GatewayConfiguration):
+    if configuration.backend not in BACKENDS_WITH_GATEWAY_SUPPORT:
+        raise ServerClientError(
+            f"Gateways are not supported for {configuration.backend.value} backend. "
+            f"Supported backends: {[b.value for b in BACKENDS_WITH_GATEWAY_SUPPORT]}."
+        )
+
+    if configuration.name is not None:
+        validate_dstack_resource_name(configuration.name)
```
jvstme, line 645: Does additional validation break backward compatibility?

src/dstack/_internal/server/services/jobs/__init__.py
```diff
@@ -316,3 +327,45 @@ def group_jobs_by_replica_latest(jobs: List[JobModel]) -> Iterable[Tuple[int, Li
...
             replica_jobs.append(latest_job_submission)
         yield replica_num, replica_jobs
+
+
+async def detach_volumes_from_instance(
+    project: ProjectModel,
+    instance: InstanceModel,
+    jpd: JobProvisioningData,
+):
+    # TODO: Volumes lock
```
jvstme: Is it ok to merge without it?

src/dstack/_internal/server/services/volumes.py
```diff
@@ -0,0 +1,259 @@
...
+    if configuration.name is not None:
+        volume_model = await get_project_volume_model_by_name(
+            session=session,
+            project=project,
+            name=configuration.name,
+        )
+        if volume_model is not None:
+            raise ResourceExistsError()
+
+    if configuration.name is None:
```
jvstme: (nit) Two identical `if`s
```suggestion
```

src/dstack/_internal/server/utils/common.py
```diff
@@ -86,6 +86,23 @@ async def wait_to_lock(lock: asyncio.Lock, locked: Set[KeyT], key: KeyT, *, dela
...
+    Lock is released during the sleep.
+    """
+    left_to_lock = keys.copy()
+    while len(left_to_lock) > 0:
+        async with lock:
+            for key in left_to_lock:
+                if key not in locked:
+                    locked.add(key)
+                    left_to_lock.remove(key)
+        await asyncio.sleep(delay)
```
jvstme, line 104: I think a deadlock is possible. Please correct me if I'm wrong.

Task T1 wants to lock keys K1, K2.
Task T2 wants to lock keys K2, K1.
Task T3 holds the lock for key K2.

T1 locks K1 and waits until K2 is released.
T3 releases K2.
T2 locks K2 and waits until K1is released.

T1 and T2 are deadlocked.

jvstme (commented):
I tried running the example from the PR description, but I always get this error in CLI:
```
Run failed with error code TERMINATED_BY_SERVER.
Error: Failed to attach volume
Check CLI and server logs for more details.
```
And this in server logs:
```
                    botocore.exceptions.ClientError: An error occurred (InvalidVolume.ZoneMismatch) when calling the AttachVolume operation: The volume       
                    'vol-019061f7850a760a0' is not in the same availability zone as instance 'i-0624790e5b10382ec'
```
I thinks something needs to be adjusted to avoid this.

But overall that's some huge work here, thank you 🙏

.github/workflows/build.yml
r4victor: They are disabled only for PRs from forks where those jobs fail anyway. We could probably separate generation / upload steps so that only uploads need to be disabled.

runner/internal/shim/docker.go
r4victor: That's an overlook. I intended to use `data` only for dstack-managed volumes. But you're probably right that we don't need to hide lost+found, so let's always mount the root of the volumes.

src/dstack/_internal/core/backends/aws/compute.py
r4victor: That's actually a TODO. Volume is currently missing project_name and it seemed a bit involved to pass it there but I should probably take care of it.

src/dstack/_internal/server/background/tasks/process_running_jobs.py
r4victor: It's a comment that we expect shim_status.state == "running" here. Should make it more clear.

src/dstack/_internal/server/services/gateways/__init__.py
r4victor: Formally yes, but I don't think it's critical. I consider no name validation at all a bug that should be fixed.

src/dstack/_internal/server/services/jobs/__init__.py
r4victor: No need for a lock there actually since volumes in use cannot be deleted.

src/dstack/_internal/server/services/volumes.py
r4victor: They are not identical: `is None` vs `is not None`. Maybe replace with else for readability...

src/dstack/_internal/server/utils/common.py
r4victor: You're right. If we make it clear to the caller that keys should be sorted, then the deadlock should not be possible?

r4victor: I also think that we should always set timeout for wait_to_lock() and wait_to_lock_many() to prevent from deadlock but this should be done in a separate issue to carefully check for timeout consequences.

src/dstack/_internal/core/models/volumes.py
r4victor: That's what I did it for JobProvisioningData and I actually regretted it – there've been several places where we could handle availability_zone uniformly. Tradeoffs here...

r4victor:
> I tried running the example from the PR description, but I always get this error in CLI:
> 
> ```
> Run failed with error code TERMINATED_BY_SERVER.
> Error: Failed to attach volume
> Check CLI and server logs for more details.
> ```
> 
> And this in server logs:
> 
> ```
>                     botocore.exceptions.ClientError: An error occurred (InvalidVolume.ZoneMismatch) when calling the AttachVolume operation: The volume       
>                     'vol-019061f7850a760a0' is not in the same availability zone as instance 'i-0624790e5b10382ec'
> ```
> 
> I thinks something needs to be adjusted to avoid this.
> 
> But overall that's some huge work here, thank you 🙏

I'm going to handle this error for already created instances so that stacktrace is not logged but the behavior will be the same.

When provisioning new instances with volumes, we can take zone into account. Take care of it.

src/dstack/_internal/server/utils/common.py
jvstme: > If we make it clear to the caller that keys should be sorted, then the deadlock should not be possible?

I think so. Or accept the keys as an unordered collection, e.g. set, and sort it inside `wait_to_lock_many`

PR #1371, opened 2024-06-28
Fix idle instances re-use in TPU

Bihan:
**Important!** This fix is dependent on [PR](https://github.com/dstackai/gpuhunt/pull/78)


Bihan:
Only [PR](https://github.com/dstackai/gpuhunt/pull/78) on gpuhunt will suffice. This PR is not needed.

PR #1373, opened 2024-06-30
[Docs] Recipe for model conversion: HF -> TRT-LLM

dat-lequoc:
Add documentation demonstrating how to use Task to convert Hugging Face models to TensorRT-LLM engines for high-throughput inference

### DIFF
--- a/examples/model-conversion/trt-llm/compile.sh
+++ b/examples/model-conversion/trt-llm/compile.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+
+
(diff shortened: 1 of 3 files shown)


peterschmidt85:
Thanks a lot, @quocdat-le-insacvl but this may only make sense with the example of deploying the converted model via the Triton server.

peterschmidt85:
For now, I'm closing the PR. I hope that's OK. Ideally, we can re-open it if we figure out how to do it end-to-end, including deployment.

PR #1391, opened 2024-07-04
Optimize loading `dstack` Docker images

### LINKED ISSUE #1280: [Bug]: Pre-pulled `dstackai/base` docker images are never used on CPU instances
### Steps to reproduce

Run this configuration on a CPU instance in any VM-based cloud (AWS, GCP, OCI, Azure).
```shell
> cat .dstack.yml
type: task
python: "3.11"
commands:
  - apt-key list | grep nvidia

> dstack run . --gpu 0 -b aws
```

### Actual behaviour

The pulling stage takes minutes, which means the pre-pulled `dstackai/base:py3.11-0.4rc4` image is **not** used. The run logs contain nvidia mentions, which means the `dstackai/base:py3.11-0.4rc4-cuda-12.1` image is used instead, which is not pre-pulled.
```
> dstack logs shaggy-mayfly-1
Warning: apt-key output should not be parsed (stdout is not a terminal)
uid           [ unknown] cudatools <cudatools@nvidia.com>
```

### Expected behaviour

The pulling stage only takes a few seconds, since the pre-pulled image is used.

### dstack version

master

### Server logs

_No response_

### Additional information

Possible solutions:
1. Choose the correct docker image based on whether the instance type is CUDA or non-CUDA. See [this TODO](https://github.com/dstackai/dstack/blob/3fb54eb050ca21318acdd124d0d37ad4feec3cbe/src/dstack/_internal/server/services/jobs/configurators/base.py#L42). But be sure to choose the image only when launching a specific instance type, as right now the `get_default_image` function is called when we cannot yet be sure whether the instance type will be CUDA or non-CUDA.
1. Ditch the non-CUDA docker images, pre-pull the CUDA docker images for both CUDA and non-CUDA VM images.

After a discussion the team decided that solution 2 is preferable.

jvstme:
- Discontinue support for non-CUDA dstackai/base Docker images. CUDA images are compatible with CPU instances and in fact are already used on them
- Pre-pull CUDA dstackai/base images in both CUDA and non-CUDA packer images
- Pre-pull Python 3.12 dstackai/base Docker images in all packer images
- Minor refactoring

Fixes #1280


r4victor (approved):
🔥

PR #1409, opened 2024-07-10
update with multi-node mode

deep-diver:
This PR extends the existing example on Alignment Handbook on multiple node configurations.

### DIFF
--- a/examples/fine-tuning/alignment-handbook/README.md
+++ b/examples/fine-tuning/alignment-handbook/README.md
@@ -45,7 +45,7 @@ ACCELERATE_LOG_LEVEL=info accelerate launch \
 For more details and other alignment methods, please check out the
 alignment-handbook's [official repository](https://github.com/huggingface/alignment-handbook).
 
-## Running via `dstack`
+## Running via `dstack` {#running-via-dstack}
 
 This example demonstrate how to run an Alignment Handbook recipe via `dstack`.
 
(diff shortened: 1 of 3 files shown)


peterschmidt85:
@deep-diver I don't see the multi-node task configuration

examples/fine-tuning/alignment-handbook/README.md
```diff
@@ -105,7 +105,65 @@ WANDB_API_KEY=<...> \
...
+      --main_process_ip=$DSTACK_MASTER_NODE_IP \
+      --main_process_port=8008 \
+      --machine_rank=$DSTACK_NODE_RANK \
+      --num_processes=$DSTACK_GPUS_NUM \
+      --num_machines=$DSTACK_NODES_NUM \
+      scripts/run_sft.py recipes/custom/config.yaml
+ports:
+  - 6006 
+resources:
+  gpu: 1..2
```
peterschmidt85: Just curious, why `1..2`? Shall it be `24GB:2`?

deep-diver: Sure. No problem! 
If there is a way to configure this at runtime, that would be much better IMO

deep-diver:
> @deep-diver I don't see the multi-node task configuration

What do you mean?

peterschmidt85:
> What do you mean?
@deep-diver I mean you have only `train.dstack.yaml` with a single node but you don't have a `train-distrib.dstack.yaml` with multiple nodes as well as other configuration files.
As a user how do I run the multi-node configuration with a single command without writing the configuration file myself?

examples/fine-tuning/alignment-handbook/README.md
peterschmidt85: My preference is always the declarative approach over runtime. In the runtime, the user always can override if needed.

deep-diver:
@peterschmidt85 updated!

Issue #1466, opened 2024-07-29
[Bug]: Support for Proxy in the Network


ulhaqi12:
### Problem

I am using dstack on a local machine and trying to run a job on a remote VM instance having GPU.

There is a proxy in the network for which I must set http_peroxy, https_proxy variables to access the internet. The dstack.shim daemon wasn't able to access the internet so I set the proxy variables for it using the following command:

```
$ sudo systemctl edit dstack-shim.service

Copied the following content in that file and saved it:

[Service]
Environment="HTTP_PROXY=http://your.proxy.server:port"
Environment="HTTPS_PROXY=https://your.proxy.server:port"
Environment="NO_PROXY=localhost,127.0.0.1"
```

After this dstack.shim daemon was working fine but now the container that dstack.shim daemon is running (dstackai/base:py3.11-0.4-cuda-12.1) doesn't have proxy variables. How can I set proxy variables for the image it is pulling from Dockerhub?

### Would you like to help us implement this feature by sending a PR?

Yes

peterschmidt85:
@ulhaqi12 Thank you for reporting this one! Please let us take a look and get back to you with an update shortly!

peterschmidt85:
@ulhaqi12 this issue is fixed in https://github.com/dstackai/dstack/releases/tag/0.18.9
Please give a try and let me know if it works now!

PR #1505, opened 2024-08-05
Add env setting to fleet config for on-prem fleets

un-def:
See: https://github.com/dstackai/dstack/issues/1466

### DIFF
--- a/src/dstack/_internal/cli/services/configurators/base.py
+++ b/src/dstack/_internal/cli/services/configurators/base.py
@@ -78,7 +78,6 @@ class BaseApplyConfigurator(ABC):
 
 
 class ApplyEnvVarsConfiguratorMixin(BaseApplyConfigurator):
-
     @classmethod
     def register_args(cls, parser: argparse.ArgumentParser):
         super().register_args(parser)
--- a/src/dstack/_internal/cli/services/configurators/base.py
+++ b/src/dstack/_internal/cli/services/configurators/base.py
@@ -1,11 +1,16 @@
 import argparse
+import os
 from abc import ABC, abstractmethod
-from typing import List
+from typing import List, cast
 
+from dstack._internal.cli.services.args import env_var
+from dstack._internal.core.errors import ConfigurationError
+from dstack._internal.core.models.common import is_core_model_instance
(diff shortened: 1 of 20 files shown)


src/dstack/_internal/cli/services/configurators/base.py
```diff
@@ -70,3 +75,29 @@ def register_args(cls, parser: argparse.ArgumentParser):
         args with different parser to show unified help.
         """
         pass
+
+
+class ApplyEnvVarsConfiguratorMixin(BaseApplyConfigurator):
```
r4victor: Any reason to inherit Mixin from BaseApplyConfigurator? I'd expect Mixin to not inherit from the base class.

un-def: For type annotations only, no reason from the runtime point of view. I have no strong opinion on this, though.

src/dstack/_internal/core/models/common.py
```diff
@@ -65,7 +65,10 @@ class Config:
     password: Annotated[str, Field(description="The password or access token")]
 
 
-def is_core_model_instance(instance: Any, class_: Type[CoreModel]) -> bool:
+_CM = TypeVar("_CM", bound=CoreModel)
+
+
+def is_core_model_instance(instance: Any, class_: Type[_CM]) -> TypeGuard[_CM]:
```
r4victor, line 71: Didn't know about TypeGuard. That's cool, thanks!

src/dstack/_internal/cli/services/configurators/base.py
r4victor: I see. I'd not depend on super to register mixin args but provide a method like register_env_args() similar to apply_env_vars() and call it from the child explicitly. It's easier to reason about and doesn't require to inherit the mixin.

r4victor (approved):

src/dstack/_internal/cli/services/configurators/base.py
un-def: Agree, I'll update the PR.

jvstme (approved):

PR #1544, opened 2024-08-12
Update Alignment Handbook tutorial with LLaMA3.1 8B

deep-diver:
This PR proposes to update Alignment Handbook tutorial with LLaMA3.1 8B

cc: @peterschmidt85 for visibility.


peterschmidt85:
@deep-diver Unfortunately, too many conflicts with the base branch prevent an easy merge.
Any chance, we could create a new PR instead of this one and try to limit the changes to?
1. Update the config. Avoid moving it and keeping the old one.
2. Ensure we don't use a fork of Alignment Handbook as they already merged your PR with the fixes for Llama 3.1.
3. See if we can update the multi-node example also to use Llama 3.1.

peterschmidt85:
@deep-diver Lets close it for now? I think it will be easier to redo it from scratch next time.

PR #1561, opened 2024-08-15
Implement encryption at rest

### LINKED ISSUE #1532: Provide mechanism to encrypt sensitive data in the DB
Currently dstack server stores all data in plaintext which includes backend credentials. All sensitive data should be stored encrypted to mitigate risks of a leaked database. To support this for different databases, the dstack server should implement encryption for specific sensitive columns which would include user tokens and backend creds, and secrets values once secrets are supported (#708). The proposed encryption scheme is to let user specify a secret value via env that is used to derive an encryption key (via pbkdf2) for AES-256 symmetric encryption.

Implementation details:
* Migration to encrypted values should be done on server init (secret key may not be available when running db migrations). The server should get a persistent global state to store whether encryption was enabled and also to store salt used to derive the key from secret.
* Handle the lost secret case. There should be a way to make the server work again by creating new admin user, reconfiguring backends, etc.

r4victor:
Closes #1532 

This PR:
* Allows configuring encryption via `server/config.yml`.
* Implements `aes` and `identity` encryption keys.
* Implements encryption of `BackendModel.auth` and `UserModel.token` using configured encryption keys.
* Updates the `server/config.yml` docs with mini-guide on encryption.

Implementation details:
* Values of all encrypted columns are prefixed with `enc:{key_type}:{key_name}:` to be able to identify encryption scheme. There is a migration to prefix exiting values with `enc:identity:noname:`.
* `UserModel.token_hash` is introduced to be able to search for user by token after tokens are encrypted.


docs/docs/reference/server/config.yml.md
```diff
@@ -744,6 +745,82 @@ In case of a self-managed cluster, also specify the IP address of any node in th
...
+encryption:
+  keys:
+  - type: aes
+    secret: cR2r1JmkPyL6edBQeHKz6ZBjCfS2oWk87Gc2G3wHVoA=
+  - type: aes
+    secret: E5yzN6V3XvBq/f085ISWFCdgnOGED0kuFaAkASlmmO4=
+projects: ...
+```
+
+Old keys may be deleted once all existing records were updated to re-encrypt sensitive data.
```
jvstme, line 791: How will the admin know if all existing records are re-encrypted? Is there a way re-encrypt everything at once?

docs/docs/reference/server/config.yml.md
```diff
@@ -744,6 +745,82 @@ In case of a self-managed cluster, also specify the IP address of any node in th
...
+## Encryption
+
+`dstack` support encryption at rest that allows
+encrypting sensitive data such as backend credentials before storing it in the DB.
+It's an additional application-level encryption layer on top of file-system and DB encryption.
+
+By default, the `dstack` server stores all data in plaintext.
+To enable encryption, specify one or more encryption keys:
+
+```yaml
```
jvstme, line 759: (nit) Here and below, consider using `<div editor-title="~/.dstack/server/config.yml"></div>`

src/dstack/_internal/server/models.py
```diff
@@ -65,6 +68,59 @@ def process_result_value(self, value, dialect):
         return value
 
 
+class DecryptedString(CoreModel):
+    """
+    A type for representing plaintext strings encrypted with `EncryptedString`.
+    Besides the string, stores information if the decryption was successfull.
```
jvstme:
```suggestion
    Besides the string, stores information if the decryption was successful.
```

src/dstack/_internal/server/testing/common.py
```diff
@@ -56,9 +57,12 @@
     VolumeModel,
 )
 from dstack._internal.server.services.jobs import get_job_specs_from_run_spec
+from dstack._internal.server.services.users import get_token_hash
 
 
-def get_auth_headers(token: str) -> Dict:
+def get_auth_headers(token: Union[DecryptedString, str]) -> Dict:
```
jvstme, line 63: (nit) Looks like `str` can be dropped, all callers use `DecryptedString`

src/dstack/_internal/server/services/encryption/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+def init_encryption_keys(encryption_key_configs: List[AnyEncryptionKeyConfig]):
+    global _encryption_keys
+    _encryption_keys = [get_encryption_key(c) for c in encryption_key_configs]
+    if not any(isinstance(key, IdentityEncryptionKey) for key in _encryption_keys):
+        _encryption_keys.append(get_identity_encryption_key())
+
+
+@contextmanager
+def encryption_keys_context(encryption_keys: List[EncryptionKey]):
```
jvstme, line 57: (nit) Consider moving to `test_encryption.py` or `testing/common.py` or documenting that this function cannot be used in server code, as it breaks thread safety

src/dstack/_internal/server/services/encryption/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    global _encryption_keys
+    prev_encryption_keys = _encryption_keys
+    _encryption_keys = encryption_keys
+    yield
```
un-def: This should probably be wrapped in `try ... finally`

un-def (approved):

src/dstack/_internal/server/models.py
```diff
@@ -65,6 +68,59 @@ def process_result_value(self, value, dialect):
...
+
+    def process_result_value(self, value: Optional[str], dialect) -> Optional[DecryptedString]:
+        if value is None:
+            return value
+        try:
+            plaintext = EncryptedString._decrypt_func(value)
+            return DecryptedString(plaintext=plaintext, decrypted=True)
+        except Exception as e:
+            logger.debug("Failed to decrypt encrypted string: %s", repr(e))
+            return DecryptedString(plaintext="", decrypted=False, exc=e)
```
jvstme: It could be subjective, but this design looks a bit error-prone to me. If the caller forgets to check `decrypted`, they may continue using `plaintext=""`, as if decryption worked. This could potentially lead to security issues, such as being able to log in to any account using an empty password.

In other words, in the current design the caller's default is to ignore decryption errors. I think the default should should be to fail.

As an alternative solution, I can suggest decrypting at field access time (i.e. the interface would be `some_field.decrypt()` instead of `some_field.plaintext`) and throwing an exception if decryption fails. And if the caller _really_ needs to ignore the exception, they will catch it and ignore it explicitly.

src/dstack/_internal/server/services/encryption/keys/aes.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __init__(self, config: AESEncryptionKeyConfig) -> None:
+        self.key = b64decode(config.secret)
+
+    def encrypt(self, plaintext: str) -> str:
+        # Generate a random 16-byte (128-bit) IV
+        iv = os.urandom(16)
+
+        # Create AES-GCM Cipher object and encrypt
+        cipher = Cipher(algorithms.AES(self.key), modes.GCM(iv), backend=default_backend())
```
jvstme: Did you consider using [`aead.AESGCM`](https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.modes.GCM) (with `associated_data=None`) instead of `modes.GCM`? I'm not sure if it would change anything apart from the interface, but the [docs](https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.modes.GCM) suggest doing so:
> If you are encrypting data that can fit into memory you should strongly consider using AESGCM instead of this.

r4victor: I began working on the AES implementation so that it can be extended to support different modes. Probably makes sense to switch to [aead.AESGCM](https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.modes.GCM) to simplify the code.

src/dstack/_internal/server/models.py
r4victor: This is a very good point about decryption errors go silent. I wasn't quite satisfied with the access time decryption design I started with so not sure about transitioning to it, but I will definitely replace direct `.plaintext` access with a method that raises an error if there were any.

docs/docs/reference/server/config.yml.md
r4victor: A good question. This is TBD. Currently, it's up to users to go over all resources and overwrite them (e.g. like Kubernetes that does not offer a way to re-encrypt at once). Not very convenient but failing to decrypt something should not be critical. I think we're going to introduce an API to trigger re-encryption later.

src/dstack/_internal/server/services/encryption/keys/aes.py
r4victor: Moved to [aead.AESGCM](https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.modes.GCM)

docs/docs/reference/server/config.yml.md
r4victor: Added `name` to `aes` key config. Now every encrypted value is prefixed with a key name (`enc:{key_type}:{key_name}:`). DB admins should be able to identify all keys currently used for encryption.

jvstme (approved):

PR #1588, opened 2024-08-20
Approximate on-prem GPU memory size

### LINKED ISSUE #1523: Approximate GPU memory size for on-premises instances
### Current

When adding on-premises instances, `dstack` records the GPU memory size reported by `nvidia-smi`. However, this memory size can be smaller than the physical GPU memory size, e.g. A10 will appear to have 22GB instead of 24GB.

```
 FLEET         INSTANCE  BACKEND       RESOURCES                                   PRICE  STATUS  CREATED    
 sour-penguin  0         ssh (remote)  30xCPU, 236GB, 1xA10 (22GB), 33.8GB (disk)  $0.0   idle    56 sec ago
```

This may hurt the UX, e.g. the instance will not be offered with the `gpu: 24GB..` filter.

### Proposed

Compare the GPU memory size from `nvidia-smi` to the memory size this GPU model is known to have. If they differ only slightly, use the known memory size. If they differ substantially or the GPU model is unknown, use the `nvidia-smi` memory size.

jvstme:
If the GPU memory size reported by `nvidia-smi` on
an on-prem instance differs slightly from the
memory this GPU model is known to have, `dstack`
will now use the known memory.

Additionally, this commit adjusts the translation
of on-prem GPU names for some edge cases so that
more on-prem GPU names match cloud GPU names and
are recognized during memory approximation.

Closes #1523

### DIFF
--- a/src/dstack/_internal/utils/gpu.py
+++ b/src/dstack/_internal/utils/gpu.py
@@ -20,8 +20,10 @@ def convert_gpu_name(name: str) -> str:
         return name.replace(" ", "")
 
     name = name.replace(" Ti", "Ti")
+    name = name.replace(" NVL", "NVL")
+    name = name.replace(" Ada Generation", "Ada")
     name = name.replace("RTX ", "RTX")
-    m = re.search(r"([A|H|L|P|T|V]\d+[Ti]?)", name)
+    m = re.search(r"([A|H|L|P|T|V]\d+\w*)", name)
     if m is not None:
         return m.group(0)
     return name.replace(" ", "")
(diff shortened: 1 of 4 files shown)


jvstme:
This PR will be ready for review after https://github.com/dstackai/gpuhunt/pull/94 is merged and the new `gpuhunt` version is released and bumped in the current branch.

src/dstack/_internal/utils/gpu.py
```diff
@@ -20,8 +20,10 @@ def convert_gpu_name(name: str) -> str:
         return name.replace(" ", "")
 
     name = name.replace(" Ti", "Ti")
+    name = name.replace(" NVL", "NVL")
+    name = name.replace(" Ada Generation", "Ada")
     name = name.replace("RTX ", "RTX")
-    m = re.search(r"([A|H|L|P|T|V]\d+[Ti]?)", name)
+    m = re.search(r"([A|H|L|P|T|V]\d+\w*)", name)
```
un-def: Not directly related to this PR, but `[A|H|L|P|T|V]` should be `[AHLPTV]`, `|` has no special meaning inside sets, it's just a literal bar.

un-def (approved):

PR #1597, opened 2024-08-21
Add LogStorage interface, CloudWatch Logs impl

### LINKED ISSUE #1577: [Feature]: CloudWatch Logs storage
### Problem

Currently, logs from jobs and runners are stored on the disk. I can be inconvenient in some deployments because requires additional actions to ensure log persistence. Cloud-based log storage brings log persistence, retention policies, and more, with near-zero maintenance.

### Solution

1. Make the log storage configurable.
2. Implement CloudWatch Logs storage.

### Workaround

_No response_

Yes

un-def:
Closes: https://github.com/dstackai/dstack/issues/1577


src/dstack/_internal/server/services/logs.py
```diff
@@ -1,117 +1,342 @@
...
+
+def _datetime_to_unix_time_ms(dt: datetime) -> int:
+    return int(dt.timestamp() * 1000)
+
+
+def _b64encode_raw_message(message: bytes) -> str:
+    return base64.b64encode(message).decode()
+
+
+class LogStorageError(Exception):
```
r4victor: All dstack exceptions should inherit DstackError.

src/dstack/_internal/server/services/logs.py
```diff
@@ -1,117 +1,342 @@
...
+    pass
+
+
+class LogStorage(ABC):
+    @abstractmethod
+    def poll_logs(self, project: ProjectModel, request: PollLogsRequest) -> JobSubmissionLogs: ...
```
r4victor: Not against ellipsis per se, but maybe use `pass` for consistency?

src/dstack/_internal/server/services/logs.py
```diff
@@ -1,117 +1,342 @@
...
+        _default_log_storage = FileLogStorage()
+    atexit.register(_default_log_storage.close)
+    return _default_log_storage
+
+
+_default_log_storage = get_default_log_storage()
+
+
+poll_logs = _default_log_storage.poll_logs
+write_logs = _default_log_storage.write_logs
```
r4victor: The functions are synchronous and they are called synchronously too. The calls should be made async (e.g. wrapped with `run_async`) to avoid blocking the event loop.

src/dstack/_internal/server/services/logs.py
```diff
@@ -1,117 +1,342 @@
...
-        )
-    if len(job_logs) > 0:
-        job_log_file_path = _get_job_log_file_path(
-            project_name=project.name,
-            run_name=run_name,
-            job_submission_id=job_submission_id,
+logger = get_logger(__name__)
+
+
+def _unix_time_ms_to_datetime(unix_time_ms: int) -> datetime:
```
r4victor: I recommend keeping private definitions at the bottom of modules for consistency.

src/dstack/_internal/server/services/logs.py
un-def: Ellipsis denotes `abstractmethod`s without implementation (`poll_logs` and `write_logs`), as it used in the Python documentation. `close`, on the other hand, is not abstract and has default implementation, thus `pass`.

But since only `pass` is used elsewhere in the project, I've replaced `...` with `pass` for consistency

src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -495,6 +495,7 @@ def _process_running(
         timestamp = job_model.runner_timestamp
     resp = runner_client.pull(timestamp)  # raises error if runner is down, causes retry
     job_model.runner_timestamp = resp.last_updated
+    # may raise LogStorageError
```
r4victor: Maybe documenting why this is ok?

r4victor (approved):

src/dstack/_internal/server/services/logs.py
un-def: This service is used in both sync and async contexts (more specifically, `write_logs` is used in a sync function already executed inside an executor, and `poll_logs` is used in an async FastAPI handler). I've added both sync and async global proxy functions for the default log storage methods: https://github.com/dstackai/dstack/pull/1597/commits/2df00938cc2e174d33caf5b57360f96d80f2625f

src/dstack/_internal/server/services/logs.py
```diff
@@ -1,117 +1,380 @@
...
+        return _default_log_storage
+    if settings.SERVER_CLOUDWATCH_LOG_GROUP:
+        if BOTO_AVAILABLE:
+            try:
+                _default_log_storage = CloudWatchLogStorage(
+                    group=settings.SERVER_CLOUDWATCH_LOG_GROUP,
+                    region=settings.SERVER_CLOUDWATCH_LOG_REGION,
+                )
+            except LogStorageError as e:
+                logger.warning("Failed to initialize CloudWatch Logs storage: %s", e)
```
jvstme: (nit) Here and below, I'd suggest using `logger.error` to trigger Sentry. Otherwise configuration errors may be unnoticed, which may lead to logs loss

jvstme (approved):

PR #1600, opened 2024-08-21
Remove S3 fallback region, rely on user configuration

### LINKED ISSUE #1599: S3Storage should use the user's default region if dstack-specific env var is not set
Currently, if `DSTACK_SERVER_BUCKET_REGION` is not set, `eu-west-1` is used as a default region.

https://github.com/dstackai/dstack/blob/b45cbf848db7b5b7b64b4df1e515b8446f4b17d1/src/dstack/_internal/server/settings.py#L34

https://github.com/dstackai/dstack/blob/b45cbf848db7b5b7b64b4df1e515b8446f4b17d1/src/dstack/_internal/server/services/storage.py#L13-L20

Arguably, the user's default region should be used instead. It's up to the user to configure their defaults, for example, via `~/.aws/config` or `AWS_DEFAULT_REGION`.

un-def:
Closes: https://github.com/dstackai/dstack/issues/1599


r4victor (approved):
Please set DSTACK_SERVER_BUCKET_REGION for dstack-cloud.

PR #1606, opened 2024-08-22
[Docs] Add CloudWatch Logs instruction

un-def:


peterschmidt85:
@un-def Made a minor change, please check if it's OK! I just made it more brief.

Issue #1443, opened 2024-07-24
[Feature]: Support keys other than RSA for adding remote instances


jvstme:
### Problem

When adding SSH instances using fleets or `dstack pool add-ssh`, it is only possible to use RSA keys. Adding other types of keys results in server errors, see #1442.

While RSA is the most common key algorithm, other algorithms are also widely adopted and are used in newer versions of SSH software by default. Currently, `dstack` users that don't use RSA keys for their on-premises instances will experience difficulties adding those instances to `dstack`.

### Solution

Add support for the following key algorithms: DSA, ECDSA, Ed25519.

### Would you like to help us implement this feature by sending a PR?

Yes

PR #1641, opened 2024-08-29
Support RSA, ECDSA, ed25519 keys(#1443)

swsvc:
DSA/DSS key type requires password when connecting via standard ssh client. Solution will require more code to write, so it is not implemented right now


un-def (approved):
Thank you for your contribution

> DSA/DSS [...]  not implemented right now

I think this is not a problem at all, DSA keys have a bad reputation and will be finally removed from OpenSSH in less than six months.

PR #1651, opened 2024-09-02
Implement Postgres locking

### LINKED ISSUE #1632: Support multi-replica server
Current dstack design imposes limitations that allow only a single server replica to run at any given moment.

Main multi-replica blockers are:

* The lack of concurrency handling for multi-process server. The locking is done in memory via Python async locks.
* The gateway can work with at most one server instance due to the server using a fixed 8001 port for the gateway-server reverse tunnel. 

dstack already supports Postgres as an SQLite alternative. The scope of this issue is to implement concurrent updates handling via Postgres to enable multi-replica server setup.

The gateway issue is to be addressed in #1633

r4victor:
Closes #1632

This PR implements database-level locking for handling concurrent processing in multi-replica server with Postgres setting. More specifically, it:

* Implements resource locking via select for update.
* Refactors code to be suitable for select for update locking. This includes replacing joinedload's, refetching attributes, ensuring the locks held for the entire transaction duration, splitting submitted jobs processing into two steps (assign instance or no instance/provision instance).
* Refactors in-memory locking to replace custom per-table locks (*_LOCK) with a unified locker interface.
* Fixes bugs with in-memory locking.
* Adds contributing guide on locking implementation.

Tested on SQLite:
* [x] Launch runs on new instances
* [x] Launch runs on existing instances
* [x] Stop runs
* [x] Delete runs
* [x] Create/delete cloud fleets
* [x] Create/delete ssh fleets
* [x] Create/attach/detach/delete volumes
* [x] Create/delete gateways 
* [x] Run multi-replica services

Tested on Postgres:
* [x] Launch runs on new instances (multi-replica server)
* [x] Launch runs on existing instances (multi-replica server)
* [x] Launch many concurrent runs (multi-replica server)
* [x] Stop runs (multi-replica server)
* [x] Create/delete cloud fleets (multi-replica server)
* [x] Create/attach/detach/delete volumes (multi-replica server)
* [x] Create/delete gateways (single replica server)
* [x] Run multi-replica services (single replica server)


contributing/LOCKING.md
```diff
@@ -0,0 +1,88 @@
...
+so `dstack` implements in-memory resource-level locking.
+In-memory locking works correctly under the assumption that there is only one server instance (process),
+which is a `dstack` limitation when using SQLite.
+
+The in-memory locking is implemented via locksets.
+Locksets are Python sets that store IDs of locked resources.
+Concurrent access to locksets is guarded with asyncio locks:
+
+```python
+lock, locket = get_lockset("my_table")
```
jvstme:
```suggestion
lock, lockset = get_lockset("my_table")
```

src/dstack/_internal/server/background/tasks/process_instances.py
```diff
@@ -99,54 +100,249 @@ async def process_instances() -> None:
...
                     ),
-                    InstanceModel.id.not_in(PROCESSING_INSTANCES_IDS),
+                    InstanceModel.id.not_in(lockset),
                 )
+                .options(lazyload(InstanceModel.job))
+                .order_by(InstanceModel.last_processed_at.asc())
+                .limit(1)
+                .with_for_update(skip_locked=True)
             )
-            instances = res.all()
```
jvstme, line 105: Why discard processing multiple instances at once? Isn't it much faster than waiting for the next scheduled `process_instances` task in 4-6 seconds?

src/dstack/_internal/server/migrations/versions/7b24b1c8eba7_add_instancemodel_last_processed_at.py
```diff
@@ -0,0 +1,68 @@
+"""Add InstanceModel.last_processed_at
```
jvstme, line 1: (nit) And `jobs.instance_assigned`

un-def (approved):

contributing/LOCKING.md
```diff
@@ -0,0 +1,88 @@
...
+
+There are few places that rely on advisory locks as when generating unique resource names.
+
+## Working with locks
+
+Concurrency is hard. Below you'll find common patterns and gotchas when working with locks to make it a bit more manageable.
+
+**A task should acquire locks on resources it modifies**
+
+This is a common sense approach. An alternative could be the inverse: job processing cannot run in parallel with run processing, so job processing takes run lock. This indirection complicates things and is discouraged.
```
jvstme: Sorry, I didn't quite understand the example before actually looking at the run and job processing code :)
```suggestion
This is a common sense approach. An alternative could be the inverse: job processing cannot run in parallel with run processing, so job processing takes run lock. This indirection complicates things and is discouraged. In this example, run processing should take job lock instead.
```

jvstme (approved):
@r4victor, thanks for this work. I found 3 unresolved TODOs/FIXMEs apart from compatibility TODOs. Are they all fine to merge?

src/dstack/_internal/server/background/tasks/process_instances.py
r4victor: Before the PR, instances were processed concurrently with `asyncio.as_completed` which cannot be done when the processing happens in the scope of one transaction (it's not safe to share an sqlalchemy session between coroutines). Processing multiple instances both concurrently and sequentially is also problematic because instance processing can take time (e.g. remote instance waiting) which would lead to blocking processing. The current approach tolerates some instances being slow to process. It's consistent with other processing tasks.

If the processing throughput/latency becomes a problem, we could just decrease processing interval / increase max_instances.

r4victor:
> 3 unresolved TODOs/FIXMEs

Those left were present before the changes and are not critical.

Issue #1633, opened 2024-08-28
Make gateway work with multiple server instances


r4victor:
Currently a gateway can work with at most one server instance due to the server using a fixed 8001 port for the gateway-server reverse tunnel. (The server establishes a reverse tunnel for the gateway, making the server API accessible to the gateway on port 8001. This is currently used for user token verification. However, since the port is fixed, two server replicas cannot create a tunnel on port 8001.) This is a blocker for #1632.

A possible solution to this problem is to establish the reverse tunnel via a Unix domain socket instead of 8001 port. This is how the direct server->gateway tunnel works. So, the server replicas will be accessible on the gateway through paths like `/somepath/server1`, `/somepath/server2`, etc, and the gateway will be able to choose any server.

Another limitation related to gateways is that the server maintains an in-memory pool of gateway connections that initialized once at server startup. New connections are added when processing submitted gateways. Since it takes place on one server replica, other replicas won't establish the connection. Submitting a service would fail on those replicas. Perhaps the easiest solution would be to establish the connection when submitting a service if it's not in the connection pool yet. There should also be a background task to keep the pool in sync with the db state to remove connections for deleted gateways.

We also need to be mindful of gateway updates compatibility to support zero-downtime (e.g. blue-green) deployments. The server updates the gateways to the latest version on startup. With a zero-downtime deployment, there will still be old server replicas running when a new one starts, so the new version of gateway should always be compatible with the old server. If incompatible changes are required, introduce them in two releases as with the incompatible db changes: 1) add and adapt code 2) remove. Gateway update process should be safe to run concurrently (e.g. take advisory locks).

To sum up:
* [x] Use Unix domain sockets for gateway-server communication.
* [ ] Adapt gateway connections management to work with multi-replica server.
* [x] Ensure the gateway update process works correctly with multi-replica server.

Issue #1595, opened 2024-08-21
[Feature]: Allow running services and models without a gateway


peterschmidt85:
**Use case:**

- As a user, I want to run a service/model and make it accessible to my project team.
- I don't want to set up a domain.
- I don't want to make the endpoint public.

**Suggestion:**
- In addition to the existing ability to create "managed" gateways (public IP address, custom domain), allow the dstack server to handle service/model traffic directly.

**Motivation:**
- So far, services have primarily been used for development, but the need to set up a domain and create a public endpoint poses a significant barrier. If the dstack server could run services/models out of the box, it would greatly simplify usage for both individuals and companies.

peterschmidt85:
Another reason why this issue is important is that in our documentation, we often use tasks for deploying models because it doesn’t require creating a gateway, which is also not supported for all backends. However, when a model is running as a task, the user cannot use the UI to interact with the model.

jvstme:
### Implementation progress

- [ ] Python-based reverse proxy for services
    - [x] HTTP proxying
    - [x] Request/response streaming
    - [ ] Websocket proxying
    - [x] Proxying to multiple service replicas
    - [x] Auth
    - [ ] Configuring whether path prefix is stripped
- [x] Running services without a gateway
- [x] OpenAI endpoint
- [x] #1954
- [x] Configuring which gateway to use for the run
- [ ] Collecting stats for replicas autoscaling
- [x] Using the new proxy implementation on gateways
    - [x] Nginx- and certbot- related logic
    - [x] Installation and updates

### Usage instructions for the current version

1. Make sure your project doesn't have a default gateway or set `gateway: false` in your service configuration.
1. Run the service.
1. Your service is now available at `<dstack-server-base-url>/proxy/services/<project-name>/<run-name>/`. If the service defines a model mapping, the OpenAI-compatible API is available at `<dstack-server-base-url>/proxy/models/<project-name>/`

peterschmidt85:
Plus, ensure the UI can leverage the built-in proxy

jvstme:
Gateways have become optional in `dstack` 0.18.23, so we will close this issue. The remaining technical debt has been moved to separate issues to allow for more granular prioritization:
- #2252
- #2251
- #2250
- #2238
- #2237
- #2232
- #2231

PR #1669, opened 2024-09-06
Refactor `SSHTunnel`

jvstme:
Replace 4 different `*SSHTunnel` classes with a
common `SSHTunnel` class. The class implements
both remote-to-local and local-to-remote port
forwarding where both TCP and Unix sockets can be
used on either end, which will be useful for
#1633 and #1595.

This commit does not change `dstack` behavior for
end users.


src/dstack/_internal/core/services/ssh/tunnel.py
```diff
@@ -1,77 +1,175 @@
...
+@dataclass
+class UnixSocket(Socket):
+    path: PathLike
+
+    def render(self) -> str:
+        return str(self.path)
+
+
+@dataclass
+class TCPSocket(Socket):
```
un-def: <sup>terminological nitpicking</sup>

I think `IPSocket` would be a more correct name, since `Unix` in `UnixSocket` means _socket_family_ aka _domain_ (`AF_UNIX` for Unix domain sockets, `AF_INET` for IPv4 sockets and `AF_INET6` for IPv6 sockets). while `TCP` means _protocol_[^1]. Although SSH without additional software is only capable of tunneling TCP traffic, I wouldn't make a data object representing “just” IP protocol address be aware of this detail.

[^1]: `AF_UNIX` sockets do not have protocols, but they do have `socket_type`, both `SOCK_STREAM` (≈ `IPPROTO_TCP` IP protocol)  and `SOCK_DGRAM` (≈ `IPPROTO_UDP`).

See: [`socket(2)`](https://man7.org/linux/man-pages/man2/socket.2.html), [`unix(7)`](https://man7.org/linux/man-pages/man7/unix.7.html), [`ip(7)`](https://man7.org/linux/man-pages/man7/ip.7.html).

un-def (approved):

Issue #1644, opened 2024-08-30
[Feature]: Windows support


un-def:
### Problem

Currently, `dstack` does not work properly on Windows.

In particular, port forwarding, attached logs,`ssh <run_name>` do not work due to OpenSSH_for_Windows [limitations](https://github.com/PowerShell/Win32-OpenSSH/wiki/Project-Scope).

### Workaround

It's possible to use `dstack` with WSL after some additional setup (see below).

### Would you like to help us implement this feature by sending a PR?

No

un-def:
# How to use dstack on Windows with WSL

Until `dstack` adds native Windows support, you can use `dstack` with WSL. To ensure that VS Code works with your dev environments, update the VS Code settings according to the guide below.

## Prerequisites

- WSL and the Linux distribution are installed
- VS Code is installed 
- The [Remote - SSH](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-ssh) extension is installed
- `dstack` is installed under WSL

## Update the VS Code settings

1. Create the following script file under WSL, e.g. `setup.bash`:

```shell
set -euo pipefail
userprofile=$(wslpath "$(cmd.exe /c '<nul set /p=%USERPROFILE%' 2> /dev/null)")
sshdir=$userprofile/.ssh
mkdir -p "$sshdir"
printf 'Include "%s"\n' "$(wslpath -w ~/.dstack/ssh/config)" >> "$sshdir/config"
shimpath=$userprofile/wsl-ssh.bat
echo "wsl -d $WSL_DISTRO_NAME ssh %*" > "$shimpath"
printf '%q\n' "$(wslpath -w "$shimpath")"
```

2. Run `setup.bash` under WSL:

```shell
bash ./setup.bash
```

The script will output the path of the `wsl-ssh.bat`, which will look something like:

```shell
C:\\Users\\<user name>\\wsl-ssh.bat
```

3. Update the VS Code IDE's `settings.json`:

```json
{ 
  "remote.SSH.path": "C:\\Users\\<user name>\\wsl-ssh.bat",
  "remote.SSH.enableDynamicForwarding": false
}
```

Here, the value of `"remote.SSH.path"` should be the path outputted by the previous step.

Now, when you run a dev environment with `dstack` under WSL, VS Code will be able to properly open it.

PR #1675, opened 2024-09-09
Add initial Windows support

un-def:
* Only client (not server)
* Only MSYS2 OpenSSH client is supported (expected to be installed with Git for Windows)

Issue: https://github.com/dstackai/dstack/issues/1644

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,7 +60,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ macos-latest, ubuntu-latest]
+        os: [ macos-latest, ubuntu-latest, windows-latest ]
         python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
     steps:
       - uses: actions/checkout@v4
@@ -75,7 +75,8 @@ jobs:
         with:
           name: frontend-build
           path: src/dstack/_internal/server/statics
-      - name: Run pytest
+      - name: Run pytest on POSIX
+        if: matrix.os != 'windows-latest'
         # Skip Postgres tests on macos since macos runner doesn't have Docker.
         # Skip Postgres tests for Python 3.8 since testcontainers<4 doesn't support asyncpg correctly.
         run: |
@@ -84,6 +85,10 @@ jobs:
             RUNPOSTGRES="--runpostgres"
           fi
           pytest src/tests --runui $RUNPOSTGRES
+      - name: Run pytest on Windows
+        if: matrix.os == 'windows-latest'
+        run: |
+          pytest src/tests --runui --runpostgres
 
   update-get-dstack:
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
--- a/src/dstack/_internal/core/services/ssh/tunnel.py
+++ b/src/dstack/_internal/core/services/ssh/tunnel.py
@@ -178,8 +178,9 @@ class SSHTunnel:
 
     def open(self) -> None:
         # We cannot use `stderr=subprocess.PIPE` here since the forked process (daemon) does not
-        # close standard streams (it calls `daemon(3)` with `noclose` set to `1`), therefore
-        # we will wait EOF from the pipe as long as the daemon exists.
+        # close standard streams if ProxyJump is used, therefore we will wait EOF from the pipe
+        # as long as the daemon exists.
+        self._remove_log_file()
         try:
             r = subprocess.run(
                 self.open_command(),
@@ -193,23 +194,26 @@ class SSHTunnel:
             raise SSHError(msg) from e
         if r.returncode == 0:
             return
-        with open(self.log_path, "rb") as f:
-            error = f.read()
-        logger.debug("SSH tunnel failed: %s", error)
-        raise get_ssh_error(error)
+        stderr = self._read_log_file()
+        logger.debug("SSH tunnel failed: %s", stderr)
+        raise get_ssh_error(stderr)
 
     async def aopen(self) -> None:
+        loop = asyncio.get_event_loop()
+        await loop.run_in_executor(None, self._remove_log_file)
         proc = await asyncio.create_subprocess_exec(
-            *self.open_command(), stdout=subprocess.DEVNULL, stderr=subprocess.PIPE
+            *self.open_command(), stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
         )
         try:
-            _, stderr = await asyncio.wait_for(proc.communicate(), SSH_TIMEOUT)
+            await asyncio.wait_for(proc.communicate(), SSH_TIMEOUT)
         except asyncio.TimeoutError as e:
+            proc.kill()
             msg = f"SSH tunnel to {self.destination} did not open in {SSH_TIMEOUT} seconds"
             logger.debug(msg)
             raise SSHError(msg) from e
         if proc.returncode == 0:
             return
+        stderr = await loop.run_in_executor(None, self._read_log_file)
         logger.debug("SSH tunnel failed: %s", stderr)
         raise get_ssh_error(stderr)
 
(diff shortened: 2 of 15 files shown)


src/tests/_internal/core/services/ssh/test_tunnel.py
```diff
@@ -58,23 +84,34 @@ def test_open_command_with_temp_identity_file(self) -> None:
...
         command = " ".join(tunnel.open_command())
         match = re.fullmatch(
-            r"ssh -F none -f -N -M -S /tmp/control.sock -i (\S+) ubuntu@my-server", command
+            (
+                rf"/usr/bin/ssh -F none -i {mocked_temp_dir}/identity "
+                rf"-E {mocked_temp_dir}/tunnel.log "
+                r"-N -f -o ControlMaster=auto -S /tmp/control.sock ubuntu@my-server"
+            ),
+            command,
         )
```
jvstme: (nit) This is no longer a regex, so string comparison can be used.

Same on line 105

src/tests/_internal/core/services/ssh/test_tunnel.py
```diff
@@ -10,22 +14,41 @@
...
+        ssh_client_info = SSHClientInfo.from_raw_version("OpenSSH_9.7p1", Path("/usr/bin/ssh"))
+        monkeypatch.setattr("dstack._internal.utils.ssh._ssh_client_info", ssh_client_info)
+        return ssh_client_info
+
+    @pytest.fixture
+    def mocked_temp_dir(self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Path:
+        mock = Mock(spec_spec=tempfile.TemporaryDirectory)
+        mock.name = str(tmp_path)
+        monkeypatch.setattr(SSHTunnel, "_init_temp_dir", Mock(return_value=mock))
+        return tmp_path
```
jvstme: (nit) If mocking is only needed to know the directory path, I'd suggest getting it from `SSHTunnel.temp_dir` instead. It seems to be easier and better for testing, since code running in tests will match that running in production.

src/dstack/_internal/core/services/ssh/tunnel.py
```diff
@@ -161,22 +182,24 @@ def proxy_command(self) -> Optional[List[str]]:
...
+                timeout=SSH_TIMEOUT,
+            )
+        except subprocess.TimeoutExpired as e:
+            msg = f"SSH tunnel to {self.destination} did not open in {SSH_TIMEOUT} seconds"
+            logger.debug(msg)
+            raise SSHError(msg) from e
         if r.returncode == 0:
             return
+        with open(self.log_path, "rb") as f:
+            error = f.read()
```
jvstme: If the tunnel is restarted multiple times (e.g. [here](https://github.com/dstackai/dstack/blob/7245280f71f9045655c8f35a3098d6cc12af702b/src/dstack/_internal/server/services/gateways/connection.py#L72)), won't `self.log_path` accumulate error messages from all restarts and lead to misleading or unintelligible logs and exceptions?

jvstme (approved):

src/dstack/_internal/core/services/ssh/tunnel.py
jvstme: **upd**: the example is using `SSHTunnel.aopen`, so it won't be the case there. But then another question is — will `SSHTunnel.aopen` detect any errors considering it is not using `self.log_path`?

src/dstack/_internal/utils/ssh.py
```diff
@@ -1,20 +1,24 @@
...
+from typing import Dict, Optional, Tuple, Union
 
 import paramiko
 from filelock import FileLock
 from paramiko.config import SSHConfig
 from paramiko.pkey import PKey, PublicBlob
 from paramiko.ssh_exception import SSHException
 
+from dstack._internal.compat import IS_WINDOWS
+from dstack._internal.core.errors import SSHError
```
r4victor: The current import hierarchy convention is utils<-core<-services, so utils should not import from core (even from errors). In other words, utils should not depend on dstack-specific definitions from other modules. Maybe move new ssh functionality to dstack._internal.services.ssh?

r4victor (approved):

src/dstack/_internal/core/services/ssh/tunnel.py
un-def: Good catch, thanks!

1. I've added a cleanup routine to `open`/`aopen` methods.
2. After some digging, I've found conditions when ssh blocks stderr (i.e. does not close stderr stream):
    * when `-v` CLI flag is used (we don't use it)
    * when `ProxyJump` is used (we use it, at least with the sync interface, see `SSHAttach`)
    
    the async interface () is also blocked under either of these conditions, so I used the same logic with the log file: https://github.com/dstackai/dstack/pull/1675/commits/1a9fc3be7cb84e25baa3bdf8c7d0095c770124bf#diff-aa9ac39757de67663ddb3e073012670ae11d79d80fbec53a9e6f823bd5225d47R216
3. `subprocess.run` [kills the process on timeout](https://github.com/python/cpython/blob/ff3bc82f7c9882c27aad597aac79355da7257186/Lib/subprocess.py#L551-L552), added `proc.kill()` to `aopen()`: https://github.com/dstackai/dstack/pull/1675/commits/1a9fc3be7cb84e25baa3bdf8c7d0095c770124bf#diff-aa9ac39757de67663ddb3e073012670ae11d79d80fbec53a9e6f823bd5225d47R210

.github/workflows/build.yml
```diff
@@ -84,6 +85,10 @@ jobs:
             RUNPOSTGRES="--runpostgres"
           fi
           pytest src/tests --runui $RUNPOSTGRES
+      - name: Run pytest on Windows
+        if: matrix.os == 'windows-latest'
+        run: |
+          pytest src/tests --runui --runpostgres
```
jvstme, line 91: Does it have to be a separate step? Looks the same as the previous one

.github/workflows/build.yml
jvstme, line 1: Please add Windows to `release.yml` too

src/dstack/_internal/core/services/ssh/tunnel.py
```diff
@@ -198,23 +195,26 @@ def open(self) -> None:
...
-            error = f.read()
-        logger.debug("SSH tunnel failed: %s", error)
-        raise get_ssh_error(error)
+        stderr = self._read_log_file()
+        logger.debug("SSH tunnel failed: %s", stderr)
+        raise get_ssh_error(stderr)
 
     async def aopen(self) -> None:
+        loop = asyncio.get_event_loop()
+        await loop.run_in_executor(None, self._remove_log_file)
```
jvstme, line 204: (nit) Consider using `dstack._internal.server.utils.common.run_async`.

I also think we usually perform file operations directly in coroutines, although it is questionable.

jvstme (approved):

src/dstack/_internal/core/services/ssh/tunnel.py
un-def: 1. Is it okay to import from `_internal.server` inside `_internal.core`? 
2. Yeah, I was not sure if it's acceptable to run blocking file I/O operations (even if the filesystem is local) inside async methods, but I didn't find neither other usages in the code nor dependencies like `aiofiles`.

cc @r4victor

.github/workflows/build.yml
un-def: In POSIX step POSIX shell syntax is used (`if [ ... ]; then`), on Windows runner PowerShell is used as a shell: https://github.com/un-def/dstack/actions/runs/10793434509/job/29935425474#step:6:19

src/dstack/_internal/core/services/ssh/tunnel.py
r4victor: > Is it okay to import from _internal.server inside _internal.core?

It's better not to, so I'd leave the code as is for now. Later, run_async can be moved to core.

PR #1706, opened 2024-09-18
[shim] Add old container cleanup routine

### LINKED ISSUE #1243: `--keep-container` leads to `no space left on device`
Steps to reproduce

1.  Add an instance with `dstack pool add-ssh`
2. Run multiple runs

Actual:

```
Cannot create container. Error: Failed to run docker pull or docker create: Cannot create dir for runner: mkdir
/root/.dstack/runners/20240518-152340: no space left on device Check CLI and server logs for more details.
```

Expected:

Old containers should stay on the instance.

> Question: Why do we use `--keep-container`?

un-def:
* Remove all containers but the last one when a container exits unless `--keep-container` is used
* Don't use `--keep-container` by default

Fixes: https://github.com/dstackai/dstack/issues/1243

### DIFF
--- a/runner/README.md
+++ b/runner/README.md
@@ -25,7 +25,7 @@ Here's the steps to build `dstack-shim` and `dstack-runner` and run `dstack` wit
 3. Start the shim:
 
     ```shell
-    ./shim --home $RUNNER_DIR --runner-binary-path $COMPILED_RUNNER_PATH docker --ssh-key $DSTACK_PUBLIC_KEY --keep-container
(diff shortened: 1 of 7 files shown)


jvstme (approved):
Maybe drop the `--keep-container` flag from shim altogether considering it's no longer used? It won't break backward compatibility, since shim version is tied to dstack server version

r4victor (approved):

jvstme (approved):

PR #1712, opened 2024-09-19
Write events to CloudWatch in batches

### LINKED ISSUE #1711: [Bug]: `CloudWatchLogStorage` doesn't respect `PutLogEvents` size limits
### Steps to reproduce

1. Start a run producing lots of logs (either more than 256KiB per event or more than 1MiB in a short time span, ≈ 5s)

### Actual behaviour

```
CloudWatch Logs error: InvalidParameterException: An error occurred (InvalidParameterException) when calling the PutLogEvents operation: [...]
```

### Expected behaviour

_No response_

### dstack version

0.18.14

### Server logs

_No response_

### Additional information

https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html

> * The maximum batch size is 1,048,576 bytes. This size is calculated as the sum of all event messages in UTF-8, plus 26 bytes for each log event.
> * Each log event can be no larger than 256 KB.
> * The maximum number of log events in a batch is 10,000.

un-def:
Fixes: https://github.com/dstackai/dstack/issues/1711

### DIFF
--- a/src/dstack/_internal/server/services/logs.py
+++ b/src/dstack/_internal/server/services/logs.py
@@ -1,10 +1,11 @@
 import atexit
 import base64
+import itertools
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Iterator, List, Optional, Set, TypedDict, Union
+from typing import Iterator, List, Optional, Set, Tuple, TypedDict, Union
 from uuid import UUID
 
 from dstack._internal.core.errors import DstackError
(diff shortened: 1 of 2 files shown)


src/dstack/_internal/server/services/logs.py
```diff
@@ -169,24 +180,70 @@ def write_logs(
...
+            cw_event = self._runner_log_event_to_cloudwatch_event(event)
+            # as message is base64-encoded, length in bytes = length in code points.
+            message_size = len(cw_event["message"]) + self.MESSAGE_OVERHEAD_SIZE
+            if message_size > self.MESSAGE_MAX_SIZE:
+                logger.warning(
+                    "Stream %s: skipping event %d, message exceeds max size",
+                    stream,
+                    event.timestamp,
+                )
+                continue
```
jvstme: If it ever happens, I think skipping the event will be quite unexpected. The user just won't find the message in the run logs and the server admin will have to deal with warnings they can't do anything about.

Maybe split such long messages into several events or cut them?

jvstme (approved):

src/dstack/_internal/server/services/logs.py
un-def: It turned out that we should never hit this limit, as the underlying code in the runner uses 32KiB buffer when copying from PTY to log writer. I've increased log level to ERROR and added a comment explaining where this assumption comes from.

jvstme (approved):

PR #1714, opened 2024-09-20
[shim] Request all Nvidia driver capabilities

### LINKED ISSUE #1713: [Feature]: Enable all Nvidia driver capabilities
### Problem

As `dstack` executes runs in docker containers, to use GPU features, e.g., CUDA, HW video encoding/decoding, OpenGL, Vulkan, Nvidia Container Runtime must be configured to use specific driver capabilities. By default, it's configured with two capabilities:

* `compute` required for CUDA and OpenCL applications.
* `utility` required for using `nvidia-smi` and NVML.

That means that currently it's not possible to run some workload that requires, say, `NVENC` or OpenGL.

### Solution

To maximize compatibility with all sorts of GPU workloads, `dstack` could create containers with all available driver capabilities: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.16.0/docker-specialized.html?highlight=capabilities#driver-capabilities

### Workaround

_No response_

Yes

un-def:
Closes: https://github.com/dstackai/dstack/issues/1713

### DIFF
--- a/runner/internal/shim/docker.go
+++ b/runner/internal/shim/docker.go
@@ -595,7 +595,13 @@ func requestGpuIfAvailable(ctx context.Context, client docker.APIClient) ([]cont
 	for runtime := range info.Runtimes {
 		if runtime == consts.NVIDIA_RUNTIME {
 			return []container.DeviceRequest{
-				{Capabilities: [][]string{{"gpu"}}, Count: -1}, // --gpus=all
+				{
+					// Request all capabilities to maximize compatibility with all sorts of GPU workloads.
+					// Default capabilities: utility, compute.
+					// https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.16.0/docker-specialized.html
+					Capabilities: [][]string{{"gpu", "utility", "compute", "graphics", "video", "display", "compat32"}},
+					Count:        -1, // --gpus=all
+				},
 			}, nil
 		}
 	}


runner/internal/shim/docker.go
```diff
@@ -595,7 +595,13 @@ func requestGpuIfAvailable(ctx context.Context, client docker.APIClient) ([]cont
 	for runtime := range info.Runtimes {
 		if runtime == consts.NVIDIA_RUNTIME {
 			return []container.DeviceRequest{
-				{Capabilities: [][]string{{"gpu"}}, Count: -1}, // --gpus=all
+				{
+					// Request all capabilities to maximize compatibility with all sorts of GPU workloads.
+					// Default capabilities: utility, compute.
+					// https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.16.0/docker-specialized.html
+					Capabilities: [][]string{{"gpu", "utility", "compute", "graphics", "video", "display", "compat32"}},
```
jvstme, line 602: (nit) Consider using [`"all"`](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.16.0/docker-specialized.html?highlight=all#driver-capabilities) if it's allowed here

jvstme (approved):

runner/internal/shim/docker.go
un-def: It' s only works with `NVIDIA_DRIVER_CAPABILITIES` env var, not CLI (`--gpus 'all,"capabilities=all"'` won't work), not API

PR #1716, opened 2024-09-23
Add missing project SSH key to on-prem instances

### LINKED ISSUE #1640: [Bug]: Cannot run services on on-prem instances
### Steps to reproduce

1. Add a gateway in any cloud
2. Add an on-prem instance
3. Try running any service with the cloud gateway and on-prem instance

### Actual behaviour

```
Run failed with error code GATEWAY_ERROR.
```

### Expected behaviour

Service starts successfully.

### dstack version

master

### Server logs

```shell
[18:33:24] INFO     dstack._internal.server.background.tasks.process_running_jobs:411 job(883487)llama31-service-ollama-0-0: now is PULLING
[18:34:15] WARNING  dstack._internal.server.background.tasks.process_running_jobs:278 job(883487)llama31-service-ollama-0-0: failed to register service replica: SSHError: SSH tunnel
                    failed to start, age=0:00:56.605746
           INFO     dstack._internal.server.background.tasks.process_runs:338 run(7c3ca0)llama31-service-ollama: run status has changed PROVISIONING -> TERMINATING
```

### Additional information

The reason is that the gateway cannot connect to the on-prem instance, as the instance does not have the project's public key in `~/.ssh/authorized_keys`. This may also mean that other project users cannot use the on-prem instance for any runs but it needs to be verified.

### Workaround

Get the project's public key from the database.
```shell
sqlite3 ~/.dstack/server/data/sqlite.db "SELECT ssh_public_key FROM projects WHERE name = 'main'"
```
Manually add the key to `~/.ssh/authorized_keys` on the on-prem instance.

un-def:
Fixes: https://github.com/dstackai/dstack/issues/1640

### DIFF
--- a/src/dstack/_internal/server/background/tasks/process_instances.py
+++ b/src/dstack/_internal/server/background/tasks/process_instances.py
@@ -230,9 +230,12 @@ async def _add_remote(instance: InstanceModel) -> None:
             )
             return
 
+        authorized_keys = [pk.public.strip() for pk in remote_details.ssh_keys]
+        authorized_keys.append(instance.project.ssh_public_key.strip())
+
         try:
             future = asyncio.get_running_loop().run_in_executor(
-                None, _deploy_instance, remote_details, pkeys
+                None, _deploy_instance, remote_details, pkeys, authorized_keys
             )
             deploy_timeout = 20 * 60  # 20 minutes
             result = await asyncio.wait_for(future, timeout=deploy_timeout)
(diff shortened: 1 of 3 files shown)


jvstme (approved):

r4victor:
It'd be nice to refactor the code to use project.ssh_private_key when connecting to remote instances such as here:

https://github.com/dstackai/dstack/blob/1ce1243e2acb0cc594494bb99e984cb6652d9ced/src/dstack/_internal/server/background/tasks/process_running_jobs.py#L136-L146

It may not be ideal doing in this release since it may break connecting to old instances. @un-def, maybe leave TODOs for that?

r4victor (approved):

jvstme:
> This may also mean that other project users cannot use the on-prem instance for any runs but it needs to be verified.

@un-def, if this is still the case, please also add a new issue to fix it later by using the project key for remote instance connections from server and CLI.

un-def:
@jvstme, it is not and was not the case, when a run is submitted to the shim, the user's public key (either provided by the user or generated by `dstack`), which is a part of the submitted task body, is dynamically added to `authorized_keys`.

The project key, as far as I understand, it's not even available to the user.

Issue #1631, opened 2024-08-27
[Feature]: Initial support for routers (using AWS Bedrock)


jvstme:
### Problem

Some users need to work with both custom and off-the-shelf language models. Deploying off-the-shelf models with `dstack` may be less convenient and financially viable than using platforms such as AWS Bedrock or Vertex AI, which provide models as a service. In addition, some proprietary models are only available through MaaS platforms. This results in users having to switch back and forth between `dstack` and MaaS platforms.

### Solution

Add support for MaaS platforms starting with AWS Bedrock. Add new `router` configurations that will allow Bedrock models to be exposed through the `dstack-gateway` OpenAI-compatible API.

### Workaround

Users can use MaaS platforms directly or use proxy solutions to bring MaaS models and models deployed by `dstack` into a single interface.

### Implementation Steps

- [ ] Storing data about models in `dstack-gateway`
- [ ] Internal API for model management in `dstack-gateway`
- [ ] Bedrock implementation in `dstack-gateway`
- [ ] MaaS-specific settings in backend configuration
- [ ] Storing data about models and routers in `dstack-server`
- [ ] `dstack-server` routers API
- [ ] Routers Python API
- [ ] `dstack apply` router configurations
- [ ] `dstack router list` and `dstack router delete` commands
- [ ] Documentation

PR #1718, opened 2024-09-23
Gateway-in-server early prototype

jvstme:
This commit implements most of the reverse
proxying logic for gateway-in-server. It also
includes a dependency injection mechanism that
will allow the new gateway app to work with
different repo (storage) implementations in-server
and remotely.

For this prototype, gateway-in-server duplicates
remote gateways, i.e. all services are available
both on a remote gateway and on gateway-in-server.
This will be changed later.

Behind the GATEWAY_IN_SERVER feature flag.

Part of #1595

Note: unit tests will likely follow later, as it should be
easier to test with the in-memory repo implementation
that will be added later for remote gateways.

### DIFF
--- a/src/dstack/_internal/gateway/deps.py
+++ b/src/dstack/_internal/gateway/deps.py
@@ -8,6 +8,13 @@ from dstack._internal.gateway.repos.base import BaseGatewayRepo
 
 
 class BaseGatewayDependencyInjector(ABC):
+    """
+    The gateway uses different implementations of this injector in different
+    environments: in-serer and on a remote host. An instance with the injector interface
+    stored in FastAPI's app.state.gateway_dependency_injector configures the gateway to
+    use a specific set of dependencies, e.g. a specific repo implementation.
+    """
+
     @abstractmethod
     async def get_repo(self) -> AsyncGenerator[BaseGatewayRepo, None]:
         if False:
--- a/src/dstack/_internal/gateway/deps.py
+++ b/src/dstack/_internal/gateway/deps.py
@@ -1,0 +1,28 @@
+from abc import ABC, abstractmethod
+from typing import AsyncGenerator
+
+from fastapi import Depends, Request
+from typing_extensions import Annotated
+
+from dstack._internal.gateway.repos.base import BaseGatewayRepo
+
+
+class BaseGatewayDependencyInjector(ABC):
+    @abstractmethod
+    async def get_repo(self) -> AsyncGenerator[BaseGatewayRepo, None]:
+        if False:
+            yield  # show type checkers this is a generator
+
+
+async def get_injector(request: Request) -> BaseGatewayDependencyInjector:
+    injector = request.app.state.gateway_dependency_injector
+    if not isinstance(injector, BaseGatewayDependencyInjector):
+        raise RuntimeError(f"Wrong BaseGatewayDependencyInjector type {type(injector)}")
+    return injector
+
+
+async def get_gateway_repo(
+    injector: Annotated[BaseGatewayDependencyInjector, Depends(get_injector)],
+) -> AsyncGenerator[BaseGatewayRepo, None]:
+    async for repo in injector.get_repo():
+        yield repo
--- a/src/dstack/_internal/gateway/services/service_proxy.py
+++ b/src/dstack/_internal/gateway/services/service_proxy.py
@@ -5,7 +5,7 @@ import fastapi
 import httpx
 from starlette.requests import ClientDisconnect
 
-from dstack._internal.gateway.repos.base import BaseGatewayRepo
+from dstack._internal.gateway.repos.base import BaseGatewayRepo, Replica, Service
 from dstack._internal.gateway.services.service_connection import service_replica_connection_pool
 from dstack._internal.utils.logging import get_logger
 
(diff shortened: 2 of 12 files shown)


src/dstack/_internal/gateway/services/service_proxy.py
```diff
@@ -0,0 +1,157 @@
...
+    except httpx.RequestError as e:
+        logger.debug(
+            "Error requesting %s %s for replica %s: %r",
+            upstream_request.method,
+            upstream_request.url,
+            replica.id,
+            e,
+        )
+        if isinstance(e, TimeoutError):
+            raise fastapi.HTTPException(fastapi.status.HTTP_504_GATEWAY_TIMEOUT)
```
r4victor: Shouldn't we handle `httpx.TimeoutException`?

src/dstack/_internal/gateway/services/service_proxy.py
```diff
@@ -0,0 +1,157 @@
...
+    project_name: str,
+    run_name: str,
+    path: str,
+    request: fastapi.Request,
+    repo: BaseGatewayRepo,
+) -> fastapi.responses.Response:
+    if "Upgrade" in request.headers:
+        raise fastapi.exceptions.HTTPException(
+            fastapi.status.HTTP_400_BAD_REQUEST, "Upgrading connections is not supported"
+        )
```
r4victor, line 25: Does it mean websocket is not supported?

src/dstack/_internal/gateway/deps.py
```diff
@@ -0,0 +1,28 @@
+from abc import ABC, abstractmethod
+from typing import AsyncGenerator
+
+from fastapi import Depends, Request
+from typing_extensions import Annotated
+
+from dstack._internal.gateway.repos.base import BaseGatewayRepo
+
+
+class BaseGatewayDependencyInjector(ABC):
```
r4victor, line 10: This Injector approach requires explanation, i.e. comment that we use it so that gateway routes can depend on different injectors when running on a server and on a remote gateway.

r4victor:
> unit tests will likely follow later, as it should be
easier to test with the in-memory repo implementation
that will be added later for remote gateways.

Which isn't coming soon necessarily, right? I think it's crucial to cover proxy() since it adds non-trivial logic that's going to be a pain to maintain/develop without tests.

src/dstack/_internal/gateway/services/service_proxy.py
```diff
@@ -0,0 +1,157 @@
...
+    return fastapi.responses.StreamingResponse(
+        stream_response(upstream_response, replica.id),
+        status_code=upstream_response.status_code,
+        headers=upstream_response.headers,
+    )
+
+
+async def stream_response(
+    response: httpx.Response, replica_id: str
+) -> AsyncGenerator[bytes, None]:
```
un-def, line 88: I think `AsyncIterator[YieldType]` could be used in place of `AsyncGenerator[YieldType, None]`

src/dstack/_internal/server/app.py
```diff
@@ -166,6 +174,10 @@ def register_routes(app: FastAPI, ui: bool = True):
     app.include_router(gateways.router)
     app.include_router(volumes.root_router)
     app.include_router(volumes.project_router)
+    if FeatureFlags.GATEWAY_IN_SERVER:
+        app.include_router(
+            service_proxy.router, prefix="/gateway/services", tags=["gateway-in-server"]
```
r4victor, line 179: /gateway/ in the path would be misleading when the server is proxying the requests. gateway-in-server should remain an internal concept.

r4victor:
Overall, I don't see a good reason to have a common repository interface (BaseGatewayRepo) and the injection mechanism unless we're going to move remote gateway from nginx to Python proxying (which I don't think is a good idea). The server and remote gateways would need different repo interfaces, although they might share some methods.

src/dstack/_internal/server/app.py
r4victor: Also, I remember we agreed that we need two separate concepts: gateways and service proxies, so I'd expect not to see "gateway-in-server" in the code as well. Currently, it's unclear what gateway is since the term is used for different things.

jvstme: Gateway is a component of dstack used to access services and models. It can run in two modes: in-server and remotely. Each mode has it's own specifics, e.g. remote gateways need to be created, while gateway-in-server exists always, remote gateways can serve services at subdomains, while gateway-in-server can't, etc.

I think this is easier for users to understand than a new term. "Service proxy" isn't a suitable term because this component will also be used for LLMs, not just services. So it will need a more generic name, like "Proxy". And having two similar terms like "Gateway" and "Proxy" can cause confusion.

src/dstack/_internal/gateway/services/service_proxy.py
jvstme: Yes, it [will](https://github.com/dstackai/dstack/blob/363b308d075f624803129acee29f43fa3e5f728c/src/dstack/_internal/gateway/routers/service_proxy.py#L35) be supported later

src/dstack/_internal/gateway/services/service_proxy.py
jvstme: I subjectively prefer more generic return types in public interfaces and more specific return types in private ones. `AsyncGenerator` is more specific than `AsyncIterator` here

jvstme:
> I don't see a good reason to have a common repository interface (BaseGatewayRepo) and the injection mechanism

Even if we keep Nginx for remote gateways, two gateway modes will still have much shared logic that will benefit from a common repo interface: retrieving model mappings for the OpenAI endpoint, service and replica details for restarting SSH tunnels after gateway restarts, user tokens for authentication, backend credentials for #1631, and any other data for future features.

Some repo methods will be useless for one of the gateway operation modes, so these methods will have to be empty in the respective repo implementation. This may be a complication, but the alternatives are hard repo dependencies in gateway code with lots of if-else statements, or completely different applications for remote gateways and gateway-in-server. Both seem like a greater complication to me, so I'd rather go with BaseGatewayRepo. If it turns out to be inconvenient, it won't be difficult to remove it at a later stage.

r4victor (approved):

PR #1732, opened 2024-09-26
Respect CloudWatch timestamp constraints

### LINKED ISSUE #1731: [Bug]: `CloudWatchLogStorage` doesn't respect `PutLogEvents` time constraints
https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html

> * None of the log events in the batch can be more than 2 hours in the future.
> * None of the log events in the batch can be more than 14 days in the past. Also, none of the log events can be from earlier than the retention period of the log group.
> * The log events in the batch must be in chronological order by their timestamp.
> * A batch of log events in a single request cannot span more than 24 hours. Otherwise, the operation fails.

Related issue: https://github.com/dstackai/dstack/issues/1727

un-def:
Fixes: https://github.com/dstackai/dstack/issues/1731

### DIFF
--- a/src/dstack/_internal/server/services/logs.py
+++ b/src/dstack/_internal/server/services/logs.py
@@ -255,7 +255,6 @@ class CloudWatchLogStorage(LogStorage):
                 excessive_event = event
                 break
             if now_timestamp - timestamp > self.PAST_EVENT_MAX_DELTA:
-                print("past", now_timestamp - timestamp, event.message)
                 skipped_past_events += 1
                 continue
             if timestamp - now_timestamp > self.FUTURE_EVENT_MAX_DELTA:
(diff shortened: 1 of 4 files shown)


src/dstack/_internal/server/services/logs.py
```diff
@@ -219,13 +235,32 @@ def _get_batch_iter(
...
             if not event.message:
                 continue
+            timestamp = event.timestamp
+            if first_timestamp is None:
+                first_timestamp = timestamp
+            elif timestamp - first_timestamp > self.BATCH_MAX_SPAN:
+                excessive_event = event
+                break
+            if now_timestamp - timestamp > self.PAST_EVENT_MAX_DELTA:
+                print("past", now_timestamp - timestamp, event.message)
```
r4victor: Forgot to remove?

r4victor (approved):

src/dstack/_internal/server/services/logs.py
un-def: Yep, thanks!

PR #1734, opened 2024-09-27
dstack-proxy naming tweaks

jvstme:
- Rename the new software component responsible for proxying to dstack-proxy and drop the terms "gateway-in-server" and "remote gateway". Once this component is ready to run both in-server and on a gateway instance, the word "gateway" will only refer to gateway instances.
- Move the in-server service proxy router from /gateway/services to /services
- Add a few docstrings and missing \_\_init\_\_ files

Part of #1595


r4victor (approved):
nice

PR #1768, opened 2024-10-02
chore: update provisioning.py

eltociear:
faild -> failed


peterschmidt85:
@eltociear Thanks for the contribution! We usually prioritize contributions that have a bigger impact, so it might be better to check out the existing issues or explore ways to make an even more meaningful improvement to dstack.

Issue #1471, opened 2024-07-30
[UX] For dev-environments, print the prompt after running the`init` commands


peterschmidt85:
**Steps to reproduce:**

1. Define a `dev-environment` with `init`:

```
type: dev-environment
name: vscode

python: "3.11"
ide: vscode

init:
  - pip install tqdm
```

2. Run it

**Actual behaviour:**

```
vscode provisioning completed (running)
/tmp/vscode-server- 100%[===================>]  57.28M   315MB/s    in 0.2s
mkdir: created directory '/root/.vscode-server'
mkdir: created directory '/root/.vscode-server/bin'
mkdir: created directory '/root/.vscode-server/bin/38c31bc77e0dd6ae88a4e9cc93428cc27a56ba40'
Installing extensions...
Installing extension 'ms-python.python'...
Installing extension 'ms-toolsai.jupyter'...
Extension 'ms-toolsai.jupyter-keymap' v1.1.2 was successfully installed.
Extension 'ms-toolsai.vscode-jupyter-slideshow' v0.1.6 was successfully installed.
Extension 'ms-toolsai.vscode-jupyter-cell-tags' v0.1.9 was successfully installed.
Extension 'ms-toolsai.jupyter-renderers' v1.0.19 was successfully installed.
Extension 'ms-toolsai.jupyter' v2024.8.1 was successfully installed.
Extension 'ms-python.debugpy' v2024.10.0 was successfully installed.
Extension 'ms-python.python' v2024.14.1 was successfully installed.
Extension 'ms-python.vscode-pylance' v2024.10.1 was successfully installed.
pip install ipykernel...

To open in VS Code Desktop, use link below:

  vscode://vscode-remote/ssh-remote+vscode/workflow

To connect via SSH, use: `ssh vscode`

To exit, press Ctrl+C.Collecting tqdm
  Downloading tqdm-4.66.5-py3-none-any.whl.metadata (57 kB)
Downloading tqdm-4.66.5-py3-none-any.whl (78 kB)
Installing collected packages: tqdm
Successfully installed tqdm-4.66.5
```

**Expected behaviour:**

```
vscode provisioning completed (running)
/tmp/vscode-server- 100%[===================>]  57.28M   315MB/s    in 0.2s
mkdir: created directory '/root/.vscode-server'
mkdir: created directory '/root/.vscode-server/bin'
[...]

PR #1818, opened 2024-10-11
[UX] Move SSH and vscode remote command to after init configurations

fool1280:
Issue #1471

Screenshot 2024-10-11 at 1 44 01 AM

### DIFF
--- a/src/dstack/_internal/server/services/jobs/configurators/dev.py
+++ b/src/dstack/_internal/server/services/jobs/configurators/dev.py
@@ -32,13 +32,14 @@ class DevEnvironmentJobConfigurator(JobConfigurator):
         commands.append(INSTALL_IPYKERNEL)
         commands += self.run_spec.configuration.setup
         commands.append("echo ''")
+        commands += self.run_spec.configuration.init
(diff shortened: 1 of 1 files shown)


SamirPS (approved):

fool1280:
forgot to add tests 😭 will do

src/dstack/_internal/server/services/jobs/configurators/dev.py
```diff
@@ -32,13 +32,14 @@ def _shell_commands(self) -> List[str]:
         commands.append(INSTALL_IPYKERNEL)
         commands += self.run_spec.configuration.setup
         commands.append("echo ''")
+        commands += self.run_spec.configuration.init
+        commands.append("echo ''")
```
r4victor: `echo ''` now runs twice in a row if self.run_spec.configuration.init is empty, which breaks the tests. I guess the second `echo ''` is not needed if self.run_spec.configuration.init is empty.

fool1280: done!

PR #1821, opened 2024-10-11
Run container as --privileged if DSTACK_DOCKER_PRIVILEGED

### LINKED ISSUE #1820: [Feature]: Ability to run privileged containers for Docker-in-Docker support
### Problem

Currently if you try to run a DinD container like what you get from [here](https://github.com/ehfd/nvidia-dind), it doesn't work because dstack generally launches containers without the `--privileged` docker flag. That flag is necessary for DinD to work.

### Solution

Could add an env var or yaml flag to run containers with `--privileged`.

### Workaround

No.

Yes

mtaran:
This is needed to run Docker-in-Docker.

I'm *guessing* the env vars passed from the `env:` section at the root of the task yaml make it to the initialization of `dstack-shim`, but please let me know if not!

Fixes #1820

### DIFF
--- a/runner/cmd/shim/main.go
+++ b/runner/cmd/shim/main.go
@@ -93,6 +93,7 @@ func main() {
 						Name:        "privileged",
 						Usage:       "Give extended privileges to the container",
 						Destination: &args.Docker.Privileged,
+						EnvVars:     []string{"DSTACK_DOCKER_PRIVILEGED"},
 					},
 					&cli.StringFlag{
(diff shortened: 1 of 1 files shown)


un-def:
> the env vars passed from the `env:` section at the root of the task yaml make it to the initialization of `dstack-shim`

Unfortunately, this doesn't work that way, but the new variable may still useful nonetheless (esp. for SSH fleets). Thanks!

Issue #1461, opened 2024-07-26
Detect the default Python version in CLI, not in dstack-server


jvstme:
### Problem

If the user omits the [`python`](https://dstack.ai/docs/reference/dstack.yml/dev-environment/#python-version) property in the run configuration, `dstack` is supposed to use the Python version installed in the user's environment. This allows `dstack` environments to be more similar to users' local environments.

However, dstack currently [detects](https://github.com/dstackai/dstack/blob/ea7b645db93077b712f041ddf06727264b89379d/src/dstack/_internal/server/services/jobs/configurators/base.py#L30) the Python version in `dstack-server`. While this works for the simplest setups, in many setups `dstack-server` runs separately from users' environments, so users actually get the Python version from the `dstack-server` environment, not their local environment.

### Solution

Detect the default Python version in `dstack` CLI, not in `dstack-server`.

easedeath:
Hi can i work on this issue?

peterschmidt85:
> Hi can i work on this issue?

@easedeath You don't need to ask. If you know how to fix this, just send a PR! Thank you

PR #1831, opened 2024-10-15
Fix: Detect default Python version in dstack CLI environment

manvith12:
This pull request addresses the issue of Python version detection in the dstack CLI. Previously, the application retrieved the Python version from the dstack-server environment, which could lead to inconsistencies for users running the CLI in different environments. 
issue: #1461 

### Changes Made
- Modified the `get_default_python_version` function to detect the Python version directly from the user's local CLI environment using `sys.version_info`.
- Implemented validation of the detected Python version with the existing `PythonVersion` class.
- Enhanced error handling to raise a `ServerClientError` if the detected version is not supported, providing clear feedback to users.

### Additional Notes
No new dependencies were introduced, and the existing error handling mechanisms were utilized to maintain consistency throughout the codebase.

### DIFF
--- a/src/dstack/_internal/server/services/jobs/configurators/base.py
+++ b/src/dstack/_internal/server/services/jobs/configurators/base.py
@@ -27,10 +27,6 @@ from dstack._internal.server.services.docker import ImageConfig, get_image_confi
 from dstack._internal.server.utils.common import run_async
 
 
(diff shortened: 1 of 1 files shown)


r4victor:
The PR does not change any behavior but only adds a docstring.

PR #1859, opened 2024-10-17
[dind] Move dind processes to a separate cgroup

### LINKED ISSUE #1854: [Bug]: DinD doesn't allow setting memory or CPU limits
### Steps to reproduce

1. make a `repro.dstack.yml` with:
```yaml
type: task
name: my-repro-task
image: dstackai/dind:latest
privileged: true
commands:
  - start-dockerd
  - sleep infinity
resources:
  cpu: 4..
  memory: 6GB..
```
2. `dstack apply -f repro.dstack.yml -y`
3. in another terminal: `dstack attach my-repro-task`
4. in yet another terminal: `ssh my-repro-task`
5. in the ssh session, try to run `docker run --rm --cpus=1 hello-world` or `docker run --rm --memory=1gb hello-world`

### Actual behaviour

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: cannot enter cgroupv2 "/sys/fs/cgroup/docker" with domain controllers -- it is in threaded mode: unknown.
```

### Expected behaviour

Container runs and is limited in the amount of memory/CPUs it can use.

### dstack version

0.18.18

### Server logs

_No response_

### Additional information

Googled for the error message and it seems like there should be some way to do this, e.g. some stackoverflow links pointed to https://github.com/earthly/earthly/pull/3203

un-def:
See: https://github.com/moby/moby/blob/65cfcc2/hack/dind#L59
See: https://github.com/earthly/earthly/blob/08b0d1f/buildkitd/dockerd-wrapper.sh#L63
Fixes: https://github.com/dstackai/dstack/issues/1854

### DIFF
--- a/docker/dind/start-dockerd
+++ b/docker/dind/start-dockerd
@@ -1,32 +1,58 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-mkdir /mnt/_tmp
-if ! mount -t tmpfs none /mnt/_tmp 2> /dev/null; then
-    echo 'docker privileged mode required'
-    rm -r /mnt/_tmp
-    exit 101
-fi
-umount /mnt/_tmp
-rm -r /mnt/_tmp
-
-if ! supervisorctl status > /dev/null; then
-    supervisord -c /etc/supervisord.conf
-    action='started'
-else
-    supervisorctl restart dockerd > /dev/null
-    action='restarted'
-fi
-
-for _i in {1..10}; do
-    if supervisorctl tail dockerd | grep -q 'API listen on'; then
-        echo "dockerd ${action}"
-        exit 0
+check_privileged_mode_or_die() {
+    mkdir /mnt/_tmp
+    if ! mount -t tmpfs none /mnt/_tmp 2> /dev/null; then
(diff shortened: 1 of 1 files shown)


docker/dind/start-dockerd
```diff
@@ -1,32 +1,58 @@
...
-for _i in {1..10}; do
-    if supervisorctl tail dockerd | grep -q 'API listen on'; then
-        echo "dockerd ${action}"
-        exit 0
+move_processes_to_nested_cgroup() {
+    # detect cgroups v2
+    if [[ -f /sys/fs/cgroup/cgroup.controllers ]]; then
+        local group=/sys/fs/cgroup/dind
+        mkdir -p ${group}
+        xargs -rn1 < /sys/fs/cgroup/cgroup.procs > ${group}/cgroup.procs || true
```
r4victor: It would be really helpful to comment why nesting is necessary because it's super non-obvious. As I understand, we cannot have a parent group with running processes, so we move them into a "dummy" nested group?

r4victor (approved):

PR #1864, opened 2024-10-19
refactor: streamline networking handling in config.py

DhanushNehru:


peterschmidt85:
@DhanushNehru, it might be worth focusing on changes that are more impactful. 🙏🏽

PR #1869, opened 2024-10-21
Allow running services without a gateway

jvstme:
Behind the PROXY feature flag.
Part of #1595


src/dstack/_internal/server/services/gateways/__init__.py
jvstme, line 1: The `services/gateways` directory needs rearranging - with the new terms, only some of this logic is gateway-specific, while some is proxy-specific. Will do in another PR.

r4victor (approved):

PR #1871, opened 2024-10-21
[Docs] Rename `HUGGING_FACE_HUB_TOKEN` to `HF_TOKEN`

peterschmidt85:


src/dstack/_internal/server/services/gateways/options.py
```diff
@@ -35,9 +35,9 @@ def get_tokenizer_config(model_id: str, hf_token: Optional[str] = None) -> dict:
         if resp.status_code == 403:
             raise ServerClientError("Private HF models are not supported")
         if resp.status_code == 401:
-            message = "Failed to access gated model. Specify HUGGING_FACE_HUB_TOKEN env."
+            message = "Failed to access gated model. Specify HF_TOKEN env."
```
jvstme, line 38: Yet `dstack` is still looking for `HUGGING_FACE_HUB_TOKEN`, not `HF_TOKEN`. See line 13 above.

I'd suggest looking for `HF_TOKEN` and using `HUGGING_FACE_HUB_TOKEN` as a fallback for compatibility.

```python
hf_token = env.get("HF_TOKEN", env.get("HUGGING_FACE_HUB_TOKEN"))
```

PR #1895, opened 2024-10-24
Add instance volumes

### LINKED ISSUE #1894: [Feature]: Instance (aka host, local) volumes (aka mounts)
### Problem

Currently, the only way to share any data between runs is to use volumes, which are supported only by a small number of backends and not supported on SSH fleets at all.

If it was possible to specify an arbitrary bind mount in the run configuration — `/path/on/fleet/instance:/path/inside/run/container` (akin Docker `-v` [bind mounts](https://docs.docker.com/engine/storage/bind-mounts/)), it would be possible to at least partially work around this limitation.

Possible use cases:

* Cache shared between runs (any runs, not necessarily the same run configuration), e.g., `~/.cache/huggingface`, `~/.cache/pip`. As it’s acceptable for the software cache to be empty, it’s okay if the host directory doesn’t exist or contains irrelevant data.
* Persistent(-ish) storage. If the user is sure that nobody reuses their fleet instance, then they can use the instance mount point as some sort of persistent storage. The lifetime of such storage is bound to the lifetime of the instance.
* Network mounts that are not yet supported by `dstack` (NFS, SMB, etc.) on SSH fleet instances.

### Solution

_No response_

### Workaround

_No response_

Yes

un-def:
Closes: https://github.com/dstackai/dstack/issues/1894

### DIFF
--- a/docs/docs/concepts/volumes.md
+++ b/docs/docs/concepts/volumes.md
@@ -1,12 +1,16 @@
 # Volumes
 
-Volumes allow you to persist data between runs. `dstack` allows to create and attach volumes to 
-dev environments, tasks, and services.
+Volumes allow you to persist data between runs. `dstack` supports two kinds of volumes: [network volumes](#network-volumes)
+and [instance volumes](#instance-volumes).
 
-> Volumes are currently supported with the `aws`, `gcp`, and `runpod` backends.
+## Network volumes
+
+`dstack` allows to create and attach network volumes to dev environments, tasks, and services.
+
+> Network volumes are currently supported with the `aws`, `gcp`, and `runpod` backends.
 Support for other backends and SSH fleets is coming soon.
 
-## Define a configuration
+### Define a configuration
 
 First, create a YAML file in your project folder. Its name must end with `.dstack.yml` (e.g. `.dstack.yml` or `vol.dstack.yml`
 are both acceptable).
@@ -38,7 +42,7 @@ If you use this configuration, `dstack` will create a new volume based on the sp
     See [.dstack.yml](../reference/dstack.yml/volume.md) for all the options supported by
     volumes, along with multiple examples.
 
-## Create, register, or update a volume
+### Create, register, or update a volume
 
 To create or register the volume, simply call the `dstack apply` command:
 
@@ -59,10 +63,11 @@ Volume my-new-volume does not exist yet. Create the volume? [y/n]: y
 
 Once created, the volume can be attached with dev environments, tasks, and services.
 
-## Attach a volume
+### Attach a volume { #attach-network-volume }
 
-Dev environments, tasks, and services let you attach any number of volumes.
-To attach a volume, simply specify its name using the `volumes` property and specify where to mount its contents:
+Dev environments, tasks, and services let you attach any number of network volumes.
+To attach a network volume, simply specify its name using the `volumes` property
+and specify where to mount its contents:
 
 <div editor-title=".dstack.yml"> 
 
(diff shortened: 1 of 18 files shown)


jvstme (approved):
That was fast!

src/dstack/_internal/core/models/configurations.py
```diff
@@ -143,6 +150,14 @@ def convert_python(cls, v, values) -> Optional[PythonVersion]:
             return PythonVersion(v)
         return v
 
+    @validator("volumes", each_item=True)
+    def convert_volumes(cls, v) -> Union[VolumeMountPoint, InstanceMountPoint]:
+        if isinstance(v, str):
+            if v.startswith("/"):
+                return InstanceMountPoint.parse(v)
+            return VolumeMountPoint.parse(v)
```
r4victor: So now we support `volume-name:mount_path`? Please mention such things in the PR description or in the docs.

r4victor: Is it a good idea at all? User may try to use relative paths for instance mounts but will get an error about volumes?

r4victor (approved):

r4victor:
Updating docs as a part of the PR makes it much easier to review.

src/dstack/_internal/core/models/configurations.py
un-def: I've improved mount point type detection, if the left part contains a slash in any position, we are sure it's an instance mount point: https://github.com/dstackai/dstack/pull/1895/commits/8e0396e37ac5291dcc8a71469bb49053773c49e9#diff-6b2463188801b3ca761b5edf2026af9f3c226b150fb67b5c7fad1665bdeacc9eR96-R99

The only case when the ambiguity is still present is when the user tries to mount a relative instance path `foo` with the `foo:/path/to` syntax — we assume it's a volume mount and fail with `Volume foo not found`.

Not sure if the verbose message [here](https://github.com/dstackai/dstack/blob/d8b11a3381a3d0a9073eb95843e402acc1a405ca/src/dstack/_internal/server/services/runs.py#L635) (something like “to use a network volume {name_or_rel_path}, create it with the volume configuration, to mount a local path, specify an absolute path `/{name_or_rel_path}`”) would greatly improve user experience, for example, it the user uses the full form syntax:

```yaml
volumes:
  - name: foo
    path: /volume
```

it's clear that they tries to mount a network volume, but the error message would still suggest a local volume fix as well. What do you think?

r4victor: > we assume it's a volume mount and fail with 

I think it's ok. No need for a different error message.

jvstme (approved):

docs/docs/concepts/volumes.md
```diff
@@ -112,19 +121,99 @@ $ dstack delete -f vol.dstack.yaml
...
+```yaml
+type: dev-environment
+# A name of the dev environment
+name: vscode-vol
+
+ide: vscode
+
+# Map the instance path to any container path
+volumes:
+  - instance_path: /mnt/volume
```
peterschmidt85, line 157: Unsure about `instance_path` and `path`.
Perhaps `instance_path` and `container_path` may be more consistent.
Also, I personally suggest promoting everywhere the short syntax:

```
volumes:
  - <instance path>:<container path>
 ```

File: src/dstack/_internal/core/models/volumes.py [hunk 3]
```diff
@@ -115,8 +115,8 @@ def parse(cls, v: str) -> Self:
 
 
 class InstanceMountPoint(CoreModel):
-    instance_path: Annotated[str, Field(description="The path on the instance (host)")]
-    path: Annotated[str, Field(description="The path in the container")]
+    instance_path: Annotated[str, Field(description="The absolute path on the instance (host)")]
+    path: Annotated[str, Field(description="The absolute path in the container")]
```
peterschmidt85, line 119: Does it mean relative paths are forbidden for container paths too?

peterschmidt85 (approved):

PR #1899, opened 2024-10-27
Update Base Dockerfile

AranavMahalpure:
Conditional SSH Installation:
By using the INSTALL_SSH build argument, SSH installation becomes optional, reducing security risks if SSH isn’t needed.

Automatic Conda Environment Activation:
Set the SHELL to bash and configured the ENTRYPOINT to source .bashrc so that workflow activates on container startup, ensuring Python and CUDA libraries work as expected.

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -1,30 +1,58 @@
+# Use the CUDA base image with Ubuntu 20.04
 FROM nvidia/cuda:12.1.0-base-ubuntu20.04
 
-ARG PYTHON
+# Set up build arguments and environment variables
+ARG PYTHON=3.8  # Default Python version if not specified
+ARG INSTALL_SSH=false  # Control SSH installation with a build argument
 ENV PYTHON=$PYTHON
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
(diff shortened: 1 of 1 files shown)


peterschmidt85:
@AranavMahalpure I suggest before submitting a PR, starting with an issue.

AranavMahalpure:
Thank you for the suggestion! I appreciate your guidance on this process.

I understand that starting with an issue allows for better discussion and planning before making changes. I will create an issue to outline the proposed changes regarding the conditional SSH installation and the automatic activation of the Conda environment. This way, we can gather feedback and ensure that everyone is aligned before moving forward with the PR.

Once the issue is created and discussed, I can proceed with the PR based on the feedback received.

Thanks again for your support!

peterschmidt85:
@AranavMahalpure Thank you.

Creating an issue helps everyone understand the problem, understand its impact and scope, and focus on the problem rather than doing an improvement just for the sake of it.

AranavMahalpure:
I completely agree! Defining an issue helps everyone understand the problem's impact and focus on effective solutions rather than making changes just for the sake of it.

AranavMahalpure:
Thanks for your guidance and support.  It will help me throughout  my Tech Journey.
thanks man.

PR #1900, opened 2024-10-27
Update Dockerfile

Reckon-11:
Consistent Indentation and Comments:
Organized comments to clarify each section’s purpose, improving readability.

Reduced Redundant Commands:
Consolidated commands in RUN for faster layer creation and reduced image size.

Metadata Labels:
Enhanced maintainability with structured LABEL statements, making it easy to add or modify metadata.

### DIFF
--- a/docker/dind/Dockerfile
+++ b/docker/dind/Dockerfile
@@ -1,24 +1,31 @@
-# Keep BASE_VERSION (base image tag) and BASE_SHA256 in sync
+# Define and use build arguments for base image and SHA
 ARG BASE_VERSION=24.04-20240827003913
 ARG BASE_SHA256=c9812aad4dbb79d800e5aecdfe41b66ffe47b73f0c070b5143c5acdcd6643ba1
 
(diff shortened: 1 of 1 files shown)


peterschmidt85:
@Reckon-11 I suggest before submitting a PR, starting with an issue.

Reckon-11:
Thank you for the suggestion! I appreciate your guidance on this process.

I understand that starting with an issue allows for better discussion and planning before making changes.
But the thing is their is no issue in this.i create the
PR for code improvement and some of Image size reduction and Label them so that they can easily modified in future .but I understand your suggestion on this.
Thanks again for your support!

peterschmidt85:
@Reckon-11 I totally understand you too. When the scope of change and reason is clear, we require no issue. When it's not, it's always better to start with an issue. This helps everyone understand the problem, understand its impact and scope, and focus on the problem rather than doing an improvement just for the sake of it.

Reckon-11:
I understand  👍

PR #1901, opened 2024-10-27
Fix infinite recursion with pydantic-duality>=1.2.3

zmievsa:


.gitignore
```diff
@@ -18,4 +18,5 @@ venv/
 build/
 .DS_Store
 .fleet
-.env
\ No newline at end of file
+.env
+.venv
```
peterschmidt85: Why is `.venv` needed?

setup.py
```diff
@@ -46,8 +46,7 @@ def get_long_description():
     "starlette>=0.26.0",
     "uvicorn",
     "pydantic>=1.10.10,<2.0.0",
-    # https://github.com/zmievsa/pydantic-duality/issues/21
-    "pydantic-duality<=1.2.2",
+    "pydantic-duality>=1.2.3",
```
peterschmidt85, line 49: I suppose 1.2.3 won't work, will it?

zmievsa:
@peterschmidt85 Here's the fix from pydantic-duality (which will make this PR obsolete but still safe so feel free not to merge it):
https://github.com/zmievsa/pydantic-duality/pull/23

I will merge it soon.

And here's the minimal reproducible example:

```python
import abc

from pydantic_duality import DualBaseModel


class AzureStoredConfig(DualBaseModel):
    resource_group: str


# from cryptography.hazmat.primitives.hashes import HashContext

# Which actually contains the code below:

# ```


class HashContext(metaclass=abc.ABCMeta):
    pass


from cryptography.hazmat.bindings._rust import openssl as rust_openssl

HashContext.register(rust_openssl.hashes.Hash)


# ```


class AzureConfig(AzureStoredConfig, BaseModel):
    pass


class GCPVolumeDiskBackendData(DualBaseModel):
    disk_type: str

```

Somehow your combination of actions causes  `issubclass(str, BaseModel)` to actually call `DualBaseModel.__subclasshook__` with `DualBaseModel` as the first argument.

zmievsa:
@peterschmidt85 @r4victor Here's the fix:
https://github.com/dstackai/dstack/pull/1902

peterschmidt85:
@zmievsa This PR can be closed? We only need to merge #1902? Correct? Or both required?

zmievsa:
This one can be safely closed. Though I recommend double checking whether you want to have it as BaseModel or not.

PR #1902, opened 2024-10-27
Update pydantic-duality to fix the infinite recursion bug

### LINKED ISSUE #21: After update to 1.2.3: RecursionError: maximum recursion depth exceeded
With previous versions, everything is fine.

With 1.2.3, we got the following:

```
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:136: in __new__
    new_class._generate_alternative_classes(
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:169: in _generate_alternative_classes
    _alter_attrs(attrs, name + request_suffix, REQUEST_ATTR),
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:71: in _alter_attrs
    annotations[key] = _resolve_annotation(val, attr)
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:54: in _resolve_annotation
    if inspect.isclass(annotation) and issubclass(annotation, BaseModel):
<frozen abc>:123: in __subclasscheck__
    ???
<frozen abc>:123: in __subclasscheck__
    ???
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:220: in __subclasscheck__
    return type.__subclasscheck__(cls, subclass) or issubclass(subclass, (cls.__request__, cls.__response__, cls.__patch_request__))
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:199: in __getattribute__
    return getattr(type.__getattribute__(self, REQUEST_ATTR), attr)
venv/lib/python3.11/site-packages/cached_classproperty/__init__.py:120: in __get__
    val = self.func(owner)
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:86: in constructor_wrapper
    obj = constructor()
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:177: in <lambda>
    tuple(_resolve_annotation(b, RESPONSE_ATTR) for b in bases),
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:177: in <genexpr>
    tuple(_resolve_annotation(b, RESPONSE_ATTR) for b in bases),
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:43: in _resolve_annotation
    return getattr(annotation, attr)
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:199: in __getattribute__
    return getattr(type.__getattribute__(self, REQUEST_ATTR), attr)
venv/lib/python3.11/site-packages/cached_classproperty/__init__.py:120: in __get__
    val = self.func(owner)
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:86: in constructor_wrapper
    obj = constructor()
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:178: in <lambda>
    _alter_attrs(attrs, name + response_suffix, RESPONSE_ATTR),
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:71: in _alter_attrs
    annotations[key] = _resolve_annotation(val, attr)
venv/lib/python3.11/site-packages/pydantic_duality/__init__.py:54: in _resolve_annotation
    if inspect.isclass(annotation) and issubclass(annotation, BaseModel):
<frozen abc>:123: in __subclasscheck__
    ???
E   RecursionError: maximum recursion depth exceeded
!!! Recursion detected (same locals & position)
=========================================================== short test summary info ============================================================
ERROR  - RecursionError: maximum recursion depth exc
```

More details in https://github.com/dstackai/dstack/actions/runs/11539562439/job/32119396209

zmievsa:
It was a long one. Here's the entire process for the fix:

Resolves https://github.com/zmievsa/pydantic-duality/issues/21

1. Here's the minimal reproducible example:

```python
import abc

from pydantic_duality import DualBaseModel


class AzureStoredConfig(DualBaseModel):
    resource_group: str


# from cryptography.hazmat.primitives.hashes import HashContext

# Which actually contains the code below:

# ```


class HashContext(metaclass=abc.ABCMeta):
    pass


from cryptography.hazmat.bindings._rust import openssl as rust_openssl

HashContext.register(rust_openssl.hashes.Hash)


# ```


class AzureConfig(AzureStoredConfig, BaseModel):
    pass


class GCPVolumeDiskBackendData(DualBaseModel):
    disk_type: str
```

2. [Here's the fix in pydantic-duality](https://github.com/zmievsa/pydantic-duality/pull/23)
3. [Here's an alternative fix within dstack](https://github.com/dstackai/dstack/pull/1901)

I am still not sure what exactly is causing the bug but here's what happens:

1. Because pydantic-duality now uses `DualBaseModel.__response__` in subclass checks, the response model gets generated (exactly once though) during the very first subclass check
2. In one of those checks, we have `issubclass(annotation, pydantic.BaseModel)`. The specific class that caused the issue here is `str` so `issubclass(str, pydantic.BaseModel)`
3. For some unknown and weird reason, it calls `DualBaseModel.__subclasshook__` instead of using `BaseModel`'s or `type`'s default subclass check. Not sure how it is possible but I verified that everything from the code block above is absolutely necessary for this situation to happen

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,8 +46,7 @@ BASE_DEPS = [
     "starlette>=0.26.0",
     "uvicorn",
     "pydantic>=1.10.10,<2.0.0",
-    # https://github.com/zmievsa/pydantic-duality/issues/21
-    "pydantic-duality<=1.2.2",
+    "pydantic-duality>=1.2.4",
     "sqlalchemy[asyncio]>=2.0.0",
     "sqlalchemy_utils>=0.40.0",
     "alembic>=1.10.2",


zmievsa:
NOTICE that I dropped support for python 3.8 as it reached its EOL. Is that fine with you?

peterschmidt85:
cc @jvstme on Python 3.8

r4victor:
@zmievsa, thanks for the writeup and the fix! We're going to support 3.13 and drop 3.8 also, so that's fine.

Regarding the issue, I have a hypothesis what can be the root cause:

> For some unknown and weird reason, it calls `DualBaseModel.__subclasshook__` instead of using BaseModel's or type's default subclass check.

This is due to this `ABCMeta.__subclasscheck__` logic that calls `__subclasscheck__` of subclassed metaclasses:

https://github.com/python/cpython/blob/19e93e2e269889ecb3c4c039091abff489f247c2/Lib/_py_abc.py#L140-L144

The problem only appears if there is any other `MyMetaclass.register()` call because it invalidates the negative subclass cache:

https://github.com/python/cpython/blob/19e93e2e269889ecb3c4c039091abff489f247c2/Lib/_py_abc.py#L28-L33

So the behaviour does not depend on any `cryptography` logic. This example will do:

```python
import abc

from pydantic_duality import DualBaseModel


class AzureStoredConfig(DualBaseModel):
    resource_group: str


class MyMeta(metaclass=abc.ABCMeta):
    pass


MyMeta.register(int)


class AzureConfig(AzureStoredConfig, BaseModel):
    pass


class GCPVolumeDiskBackendData(DualBaseModel):
    disk_type: str
```

r4victor (approved):

zmievsa:
@r4victor nice catch! You are likely right.

jvstme:
I will take care of merging this PR just a bit later, once we drop Python 3.8. Probably today. Thanks, @zmievsa!

**upd**: will rebase and merge after #1910

SamirPS (approved):

PR #1915, opened 2024-10-29
minor bug fix. Signed-off-by: Ritesh Udgata udgataritesh2005@gmail.com

### LINKED ISSUE #1590: [UX] Support a shorter syntax for model mapping
Currently, I have to specify:

```
model:
  type: chat
  name: meta-llama/Meta-Llama-3.1-70B-Instruct
  format: openai
```

I suggest that we allow to specify

```
model: meta-llama/Meta-Llama-3.1-70B-Instruct
```

In that case `type` and `format` will use the default values.

Ritesh-Udgata:
closes #1590

### DIFF
--- a/examples/deployment/tgi/amd/service.dstack.yml
+++ b/examples/deployment/tgi/amd/service.dstack.yml
@@ -17,7 +17,4 @@ resources:
 
 spot_policy: auto
 
-model:
-  type: chat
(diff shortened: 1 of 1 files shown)


peterschmidt85:
@Ritesh-Udgata The shorter syntax isn't supported yet.
#1590 is about adding this support

PR #1920, opened 2024-10-30
Allow to customize AMIs used by AWS backend

### LINKED ISSUE #1913: [Feature]: [AWS] Allow to specify AMI identifiers in the backend config
### Problem

Currently, it's not possible to use AMIs other than hardcoded in the `aws` backend, which means that (at least):

* it's not possible to customize instances, e.g., automount some network volumes, configure Docker or use a different Docker version, etc.
* some instance features are not available due to the missing kernel modules
* on NVIDIA instances, users can only use Docker images with the CUDA version compatible with the host CUDA version chosen by dstack (it's 12.1 currently) and only those CUDA libraries that are installed in the dstack-provided AMIs.

### Solution

Add a backend option to override the default dstack-provided AMIs.

### Workaround

_No response_

Yes

un-def:
Closes: https://github.com/dstackai/dstack/issues/1913

### DIFF
--- a/docs/docs/reference/server/config.yml.md
+++ b/docs/docs/reference/server/config.yml.md
@@ -179,6 +179,42 @@ There are two ways to configure AWS: using an access key or using the default cr
     Using private subnets assumes that both the `dstack` server and users can access the configured VPC's private subnets.
     Additionally, private subnets must have outbound internet connectivity provided by NAT Gateway, Transit Gateway, or other mechanism.
 
+??? info "OS images"
+    By default, `dstack` uses its own [AMI :material-arrow-top-right-thin:{ .external }](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html)
+    optimized for `dstack`.
+    To use your own or other third-party images, set the `os_images` property:
+
+    ```yaml
+    projects:
+      - name: main
+        backends:
+          - type: aws
+            creds:
+              type: default
+
+            os_images:
+              cpu:
+                name: my-ami-for-cpu-instances
+                owner: self
+                user: dstack
+              nvidia:
+                name: 'Some ThirdParty CUDA image'
+                owner: 123456789012
+                user: ubuntu
+    ```
+
+    Here, both `cpu` and `nvidia` properties are optional, but if the property is not set, you won´t be able to use the corresponding instance types.
+
+    The `name` is an AMI name.
+    The `owner` is either an AWS account ID (a 12-digit number) or a special value `self` indicating the current account.
+    The `user` specifies an OS user for instance provisioning.
+
+    !!! info "Image requirements"
+        * SSH server listening on port 22
+        * `user` with passwordless sudo access
+        * Docker is installed
+        * (For NVIDIA instances) NVIDIA/CUDA drivers are installed
+
 #### Azure
 
 There are two ways to configure Azure: using a client secret or using the default credentials.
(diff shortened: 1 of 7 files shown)


r4victor (approved):

docs/docs/reference/server/config.yml.md
```diff
@@ -179,6 +179,42 @@ There are two ways to configure AWS: using an access key or using the default cr
...
+    Here, both `cpu` and `nvidia` properties are optional, but if the property is not set, you won´t be able to use the corresponding instance types.
+
+    The `name` is an AMI name.
+    The `owner` is either an AWS account ID (a 12-digit number) or a special value `self` indicating the current account.
+    The `user` specifies an OS user for instance provisioning.
+
+    !!! info "Image requirements"
+        * SSH server listening on port 22
+        * `user` with passwordless sudo access
+        * Docker is installed
```
jvstme, line 215: And enabled (`systemctl enable docker`). I think it's not enabled by default after installation

docs/docs/reference/server/config.yml.md
```diff
@@ -179,6 +179,42 @@ There are two ways to configure AWS: using an access key or using the default cr
...
+        * (For NVIDIA instances) NVIDIA/CUDA drivers are installed
```
jvstme: And `nvidia-container-toolkit`

src/dstack/_internal/core/backends/aws/resources.py
```diff
@@ -6,22 +6,45 @@
...
+def get_image_id_and_username(
+    ec2_client: botocore.client.BaseClient,
+    cuda: bool,
+    image_config: Optional[AWSOSImageConfig] = None,
+) -> tuple[str, str]:
+    if image_config is not None:
+        image = image_config.nvidia if cuda else image_config.cpu
+        if image is None:
+            logger.warning("%s image not configured", "nvidia" if cuda else "cpu")
+            raise ComputeResourceNotFoundError()
```
jvstme, line 26: (nit) Btw, alternatively we could filter out the offers that don't have an image configured. This way, we won't waste provisioning attempts and will show more realistic run plans. That would especially be useful for Azure, e.g. to filter out A10 offers if the GRID image is not configured.

Yet the current implementation provides more insight into why the offers aren't working, so I'm not sure which alternative is better and we can stick with the current one.

jvstme (approved):

docs/docs/reference/server/config.yml.md
un-def: At least on Debian, Docker installed from `https://download.docker.com/linux/` is enabled by default, ~~moreover, it also uses socket activation~~ UPD active state is not relevant to enabled state

```
debian@debian:~$ systemctl status docker
● docker.service - Docker Application Container Engine
     Loaded: loaded (/lib/systemd/system/docker.service; enabled; preset: enabled)
```

Anyway, I think it's too obvious to mention if someone is already into custom OS images, and I would also avoid any specific instructions on how to check it/enable, as this is init-specific (there are still distributions that use init systems other than systemd :) )

src/dstack/_internal/core/backends/aws/resources.py
un-def: As for now, I would stick to the current implementation, as I assume that the setup where custom AMIs are used, but not all (currently, all = only two) kinds are configured, is rare. In the future, if there are more kinds of images, we could revisit the logic.

Issue #1699, opened 2024-09-17
[Bug]: Adding on-prem node/instance to two different fleets shouldn't be allowed


james-boydell:
### Steps to reproduce

Create a fleet with an on-prem node/instance. Change the name of the fleet in the config and apply. The on-prem node will now be a part of two different fleets. 

### Expected behaviour

When adding a node/instance to a fleet, it should check if that node/instance is already part of a fleet

### dstack version

0.18.13 and 0.18.14rc1

r4victor:
Closed by #1938

PR #1938, opened 2024-10-31
Return error when instance added to multiple fleets(#1699)

swsvc:


src/dstack/_internal/server/services/fleets.py
```diff
@@ -137,6 +140,27 @@ async def get_plan(
...
+
+        instances_already_in_fleet = []
+        for new_instance in spec.configuration.ssh_config.hosts:
+            if new_instance in existing_hosts:
+                instances_already_in_fleet.append(new_instance)
+
+        if instances_already_in_fleet:
+            raise ConfigurationValidationError(
+                detail=f"Instances [{', '.join(instances_already_in_fleet)}] are already assigned to a fleet."
+            )
```
r4victor: Please raise `ServerClientError("error message")`. It will be converted to a 400 response and handled correctly by the CLI. There is no need to introduce new errors.

src/dstack/_internal/server/services/fleets.py
```diff
@@ -137,6 +140,27 @@ async def get_plan(
...
+
+        existing_hosts = set()
+        for instance in active_instances:
+            instance_conn_info = RemoteConnectionInfo.parse_raw(
+                cast(str, instance.remote_connection_info)
+            )
+            existing_hosts.add(instance_conn_info.host)
+
+        instances_already_in_fleet = []
+        for new_instance in spec.configuration.ssh_config.hosts:
```
r4victor: `ssh_config.hosts` may store hostnames but it may also store structs containing a hostname. You need to handle the second case.

https://github.com/dstackai/dstack/blob/19cd2fdc667f312720f5b228920870ecdc79892a/src/dstack/_internal/core/models/fleets.py#L64-L69

src/dstack/_internal/server/services/fleets.py
```diff
@@ -137,6 +140,27 @@ async def get_plan(
 ) -> FleetPlan:
     # TODO: refactor offers logic into a separate module to avoid depending on runs
 
+    if spec.configuration.ssh_config and spec.configuration.ssh_config.hosts:
```
r4victor: The entire check should be moved into a separate function like `_check_ssh_hosts_not_yet_added()`

swsvc:
Done, please review fixes

r4victor:
@swsvc, merged, thanks!

PR #1945, opened 2024-11-01
[shim] Change NVIDIA GPU detection method

### LINKED ISSUE #1942: [Bug]: Not possible to use OS image with NVIDIA drivers on CPU instances
### Steps to reproduce

*Note*: currently, only AWS backend allows to customize OS images, but the issue is not backend-specific.

1. Configure a backend to use an OS image with NVIDIA drivers with CPU instances:
    ```yaml
    projects:
      - name: main
        backends:
          - type: aws
            creds:
              type: default
            os_images:
              cpu:
                name: dstack-cuda-0.6
                owner: 142421590066 
                user: ubuntu
    ```
2. Start a run on a CPU instance:
    ```yaml
    type: dev-environment
    ide: vscode
    resources:
      gpu: 0
    ```

### Actual behaviour

```
Run failed with error code CONTAINER_EXITED_WITH_ERROR.
Error: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start
container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: Auto-detected mode
as 'legacy'
nvidia-container-cli: initialization error: nvml error: driver not loaded: unknown
```

### Expected behaviour

_No response_

### dstack version

0.18.22

### Server logs

_No response_

### Additional information

_No response_

un-def:
* Check for `/dev/nvidiactl`, not `nvidia-smi` binary, to detect NVIDIA GPU. Unlike the binary, the devfs file only exists if some conditions are met (the exact way how `/dev/ndivia*` character device files are created is complicated and setup-specific — involving some of: kernel module, udev, modprobe, nvidia-persistenced, X server, and more — but in general, it should be safe to assume that if NVIDIA GPU is available, then `/dev/nvidiactl` does exist.
* Run `nvidia-smi` to get GPU info directly on the host, not inside a container. Using Docker is completely unnecessary, as NVIDIA Container Toolkit mounts libs and executables from the host — dstack-provided Docker image doesn't even contain `nvidia-smi` binary, it's always a bind-mounted file from the host.

Fixes: https://github.com/dstackai/dstack/issues/1942

### DIFF
--- a/runner/internal/shim/gpu.go
+++ b/runner/internal/shim/gpu.go
@@ -8,14 +8,12 @@ import (
 	"io"
 	"log"
 	"os"
-	"os/exec"
 	"strconv"
 	"strings"
 
 	execute "github.com/alexellis/go-execute/v2"
 )
 
-const nvidiaSmiImage = "dstackai/base:py3.13-0.6-cuda-12.1"
 const amdSmiImage = "un1def/amd-smi:6.2.2-0"
 
 type GpuVendor string
@@ -36,7 +34,7 @@ func GetGpuVendor() GpuVendor {
 	if _, err := os.Stat("/dev/kfd"); !errors.Is(err, os.ErrNotExist) {
 		return Amd
 	}
-	if _, err := exec.LookPath("nvidia-smi"); err == nil {
+	if _, err := os.Stat("/dev/nvidiactl"); !errors.Is(err, os.ErrNotExist) {
 		return Nvidia
 	}
 	return NoVendor
@@ -56,14 +54,8 @@ func getNvidiaGpuInfo() []GpuInfo {
 	gpus := []GpuInfo{}
 
 	cmd := execute.ExecTask{
-		Command: "docker",
-		Args: []string{
-			"run",
-			"--rm",
-			"--gpus", "all",
-			nvidiaSmiImage,
-			"nvidia-smi", "--query-gpu=gpu_name,memory.total", "--format=csv,nounits",
-		},
+		Command:     "nvidia-smi",
+		Args:        []string{"--query-gpu=gpu_name,memory.total", "--format=csv,nounits"},
 		StreamStdio: false,
 	}
 	res, err := cmd.Execute(context.Background())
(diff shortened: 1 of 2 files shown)


runner/internal/shim/gpu.go
```diff
@@ -56,14 +54,8 @@ func getNvidiaGpuInfo() []GpuInfo {
...
 
 	cmd := execute.ExecTask{
-		Command: "docker",
-		Args: []string{
-			"run",
-			"--rm",
-			"--gpus", "all",
-			nvidiaSmiImage,
-			"nvidia-smi", "--query-gpu=gpu_name,memory.total", "--format=csv,nounits",
-		},
```
jvstme, line 66: Running `nvidia-smi` in Docker actually has a small advantage, since it also validates that the NVIDIA container runtime is installed and working. However, it's not reporting errors to the server, so the erroneous instance is still added to the fleet, just without any GPUs in resources. Considering this validation doesn't work properly, I think it's OK to drop it for now the way it's done in this PR and later implement it properly in #1947.

jvstme (approved):

runner/internal/shim/gpu.go
un-def: Right, didn't think about it in this context. The motivation to get rid of Docker here was to not rely on some third-party (from the point of view of the self-hosted dstack administrator) binary blob (dstack-provided Docker image) — yes, we have these images pre-pulled in the dstack-provided OS images, but if I use SSH fleets or a cloud provider with my own OS images, this check may slow down provisioning a bit (the time to pull our Docker image), and left the image that I have no intention to use cached on the instance.

I think that it would be enough to check [Client.Info](https://pkg.go.dev/github.com/docker/docker/client#Client.Info) to ensure that 1) Docker is installed and running, and 2) NVIDIA Container Toolkit runtime is installed.

We still rely on `un1def/amd-smi` image as a temporary solution to get the info about AMD GPU, but I think we can either make `amd-smi` a hard requirement (include it in dstack-provided OS images and require it to be present on AMD-powered user-provided OS images and SSH instances) or bring our statically-linked `amd-smi` binary (not sure if it as easy to do as to say).

PR #1948, opened 2024-11-03
Add user tag to AWS and GCP volumes

### LINKED ISSUE #1943: [Feature]: Add dstack_user tag to AWS volumes
### Problem

add the `dstack_user` tag to volumes, as this is useful for cost tracking

### Solution

_No response_

### Workaround

_No response_

No

james-boydell:
fixes: #1943 

-extended to gcp as well


src/dstack/_internal/server/routers/volumes.py
```diff
@@ -66,12 +66,14 @@ async def get_volume(
 async def create_volume(
     body: CreateVolumeRequest,
     session: AsyncSession = Depends(get_session),
+    user: UserModel = Depends(Authenticated()),
     user_project: Tuple[UserModel, ProjectModel] = Depends(ProjectMember()),
 ) -> Volume:
     _, project = user_project
```
r4victor: No need to get user via Authenticated() since it's returned by ProjectMember(). Just do `user, project = user_project` like in other endpoints.

src/dstack/_internal/server/migrations/versions/9be642a33d08_added_user_id_and_user_to_volumes_table.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        batch_op.create_foreign_key(
+            batch_op.f("fk_volumes_user_id_users"),
+            "users",
+            ["user_id"],
+            ["id"],
+            ondelete="CASCADE",
+        )
+
+    # ### end Alembic commands ###
```
r4victor: This migration won't work on existing installations that already have volumes in the db. (You cannot add non-nullable field to existing rows).  I can suggest two solutions:

1. Just make VolumeModel.user_id/VolumeModel.user optional and work around that.
2. Ideally, we'd like to make user required. To do that, you can write a migration that adds a new nullable user field, then populates existing rows with values (e.g. from volume_model.project.owner), then makes the field non-nullable.

Solution 2 is preferable to avoid handling optional user case in the future, but I'm ok with merging Solution 1 as well if you're not feeling like writing data migrations.

r4victor:
@james-boydell, I also recommend running Python tests locally for a faster feedback loop. It should be fairly simple to set up: install deps from requirements_dev.txt and run `pytest src/tests/`.

src/dstack/_internal/server/migrations/versions/9be642a33d08_added_user_id_and_user_to_volumes_table.py
james-boydell: Good point. let me try and populate existing volumes with the user from volume's project owner.

src/dstack/_internal/server/routers/volumes.py
james-boydell: fixed

src/dstack/_internal/server/migrations/versions/9be642a33d08_added_user_id_and_user_to_volumes_table.py
james-boydell: updated migration to set volume.user_id to the projects owner

r4victor (approved):

r4victor:
@james-boydell, thanks for the PR, nice work! I added a commit to handle backward compatibility. Merging it.

r4victor:
Found missing VolumeModel.user loading in few places. Fixed it.

PR #1957, opened 2024-11-04
[Docs] Document `resources.gpu.vendor` property

un-def:
`GPUSpecSchema` is used in the documentation, `GPUSpec` is used elsewhere (including `dstack.api` as `GPU`).


un-def:
image

r4victor (approved):

PR #1972, opened 2024-11-07
Allow specifying gateway in service configurations

jvstme:
Part of #1595


src/dstack/_internal/core/models/configurations.py
```diff
@@ -267,6 +276,16 @@ def convert_replicas(cls, v: Any) -> Range[int]:
             )
         return v
 
+    @validator("gateway")
+    def validate_gateway(
```
un-def, line 280: Is this for the sake of human-readable error message? Otherwise, `Optional[Union[Literal[False], str]]` would do the same validation automatically (didn't check though)

un-def (approved):

src/dstack/_internal/core/models/configurations.py
jvstme: I tried `Union[Literal[False], str]` first, but this way `true` passes validation and is converted to `"true"`, which is not expected

PR #1977, opened 2024-11-09
AWS Capacity Reservation support

### LINKED ISSUE #1155: [Feature]: Support the Capacity Blocks for ML feature for AWS
### Problem

With AWS, [Capacity Blocks for ML](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-capacity-blocks.html) is the main way to access H100 today.

### Solution

_No response_

### Workaround

_No response_

Yes

timsolovev:
Closes #1155. 

Implemented `reservation` profile parameter. It is not a final version wanted to check a few things first:

- Wasn't sure if I should include any explicit validation on CLI client, such as verifying that `spot_policy: spot` and `reservation: <id>` are not provided simultaneously. It makes more sense to implicitly filter out `spot: true` offerings during the offer acquisition, but could not find a way to do it in the least intrusive way. 

- Couldn't squeeze in server-side capacity block validations inside `AWSCompute` without modifying the signatures of existing functions. For example, capacity blocks should certainly be validated in regards to their available capacity and instance type, and this is something that cannot extracted just from `requirements` alone. Is it okay if I expand the base class `Compute` with additional methods for reservation validation?

This is only a fleet implementation in the moment, some minor changes for dev environments are left out on purpose to avoid intrusive plumbing without checking up with you first. Thanks!


timsolovev:
One addition: from what it seems, Capacity Reservations and Capacity Blocks share the same API with negligible differences, so one can replace the other for testing, at least in this scenario.

jvstme (commented):
Hi @solovyevt, thanks for this draft, everything looks good so far.

> explicit validation on CLI client, such as verifying that `spot_policy: spot` and `reservation: <id>` are not provided simultaneously

I think validation in CLI is optional. The server will still need to filter out spot offers if the configuration specifies `spot_policy: auto`.

> filter out `spot: true` offerings during the offer acquisition, but could not find a way to do it in the least intrusive way

As an option, this filtering can be done in `AWSCompute.get_offers` if you'll make this method aware of whether a capacity block is used or not.

> Is it okay if I expand the base class Compute with additional methods for reservation validation?

Yes, feel free to add new methods or update the signatures of existing ones as you see fit. I think the validation could be done in `AWSCompute.get_offers` if you make it aware of capacity blocks, but that choice is up to you.

timsolovev:
@jvstme Ready for review

src/dstack/_internal/cli/utils/fleet.py
```diff
@@ -51,6 +53,7 @@ def get_fleets_table(fleets: List[Fleet], verbose: bool = False) -> Table:
 
             row = [
                 fleet.name if i == 0 else "",
+                fleet.spec.configuration.reservation if i == 0 else "",
```
r4victor: Only include reservation value if verbose. Otherwise it shifts all the values without verbose:

```
✗ dstack fleet
 FLEET                  INSTANCE  BACKEND  RESOUR…  PRICE     STATUS  CREATED      
 popular-rattlesnake-1            0        aws      2xCPU,    $0.107  provis…  now 
                                           (eu-we…  8GB,                           
                                                    100.0GB                        
                                                    (disk) 
```

src/dstack/_internal/core/backends/aws/resources.py
```diff
@@ -596,3 +602,35 @@ def _is_private_subnet_with_internet_egress(
...
     return False
+
+
+def get_reservation(
+    ec2_client: botocore.client.BaseClient,
+    reservation_id: str,
+    instance_count: Optional[int] = 0,
+    instance_types: Optional[List[str]] = None,
+    is_capacity_block: Optional[bool] = False,
+) -> Optional[str]:
```
r4victor: I believe it's not an str that's returned?

src/dstack/_internal/core/backends/aws/resources.py
```diff
@@ -596,3 +602,35 @@ def _is_private_subnet_with_internet_egress(
...
 
     return False
+
+
+def get_reservation(
+    ec2_client: botocore.client.BaseClient,
+    reservation_id: str,
+    instance_count: Optional[int] = 0,
+    instance_types: Optional[List[str]] = None,
+    is_capacity_block: Optional[bool] = False,
```
r4victor: I don't think there is a need to make instance_count or is_capacity_block optional – just having default values would do.

r4victor:
Pointed out some issues, but LGTM overall. Leaving it to @jvstme for a more thorough review.

src/dstack/_internal/core/models/profiles.py
```diff
@@ -109,6 +109,10 @@ class ProfileParams(CoreModel):
             description="The cloud-specific instance types to consider for provisioning (e.g., `[p3.8xlarge, n1-standard-4]`)"
         ),
     ]
+    reservation: Annotated[
+        Optional[str],
+        Field(description="The existing reservation for the instances"),
+    ] = None
```
r4victor: Introducing a new field breaks client backward compatibility since a new client will send a value to an old server that will reject it. The CLI can explicitly handle that (some examples https://github.com/dstackai/dstack/blob/a7410ac605a37bc4342c9b4f5a1a079c882674f9/src/dstack/api/server/_runs.py#L115-L127)

Probably @jvstme can help with that.

src/dstack/_internal/core/models/fleets.py
```diff
@@ -100,6 +100,10 @@ class InstanceGroupParams(CoreModel):
         Optional[InstanceGroupPlacement],
         Field(description="The placement of instances: `any` or `cluster`"),
     ] = None
+    reservation: Annotated[
+        Optional[str],
+        Field(description="The existing reservation for the instances"),
+    ] = None
```
r4victor, line 134: Just pointing out another client backward incompatibility that we need to address.

src/dstack/_internal/cli/utils/fleet.py
timsolovev: Fixed in [5647aff](https://github.com/dstackai/dstack/pull/1977/commits/5647aff2094cde0300eb360ffecfcf47bee3335f)

src/dstack/_internal/core/backends/aws/resources.py
timsolovev: Fixed in [d48d3ec](https://github.com/dstackai/dstack/pull/1977/commits/d48d3ecefcf08c2e5d9da6a516222e512e666f03)

src/dstack/_internal/core/backends/aws/resources.py
timsolovev: Fixed in [d48d3ec](https://github.com/dstackai/dstack/pull/1977/commits/d48d3ecefcf08c2e5d9da6a516222e512e666f03)

src/dstack/_internal/core/models/profiles.py
timsolovev: I've addressed backward compatibility in [2b93408](https://github.com/dstackai/dstack/pull/1977/commits/2b93408dd4291071c89c8115f0fea639e402fddd), but would be great to have it validated. Tested myself against the latest server version, i.e. `0.18.25`.

r4victor:
@solovyev, fix the tests please.

timsolovev:
@r4victor Python tests are fixed. For the future, is it possible to run them locally? Couldn't find any info in contribution docs.

r4victor:
@solovyevt, we somehow missed the section on tests in contributing. I updated it: https://github.com/dstackai/dstack/blob/master/CONTRIBUTING.md#run-tests

timsolovev:
Please wait with the merge.

Suddenly realised that there might be a nasty bug related to resource quota filtering in `get_offers()` in `AWSCompute` class since Capacity Blocks - unlike Capacity Reservations - do not count into an account-wide on-demand instance quota. This might lead to reserved compute instances to be filtered out and fail with `No offers available`.

To validate the fix more thoroughly, I will test it on one of our upcoming Capacity Blocks next Monday. Sorry for the trouble.

r4victor:
@solovyevt, what's the status of this PR?

timsolovev:
@r4victor Ready for review. 

One thing I've noticed is that the JSON schema is probably missing the `reservation` property, so the linter in Pycharm keeps marking it as incorrect field in any fleet or devenv/task/service yaml. Is that expected?
image

r4victor:
@solovyevt, thanks! We'll take a look at it ASAP, ideally before the release on Wed.

Regarding the json schema, if you have other properties working, then `reservation` should work as well with the next release – the json schema will be updated and re-uploaded. No action is required from you.

src/dstack/_internal/core/backends/aws/compute.py
```diff
@@ -706,11 +755,13 @@ def _supported_instances(offer: InstanceOffer) -> bool:
         "g4dn.",
         "g5.",
         "g6.",
+        "g6e.",
         "gr6.",
         "p3.",
         "p4d.",
         "p4de.",
         "p5.",
+        "p5e",
```
r4victor: "p5e" -> "p5e." for consistency?

Also note that p5e instances won't work with dstack now because gpuhunt doesn't include these offers – AWS stopped to return pricing for p5e in their pricing csv for some reason. Apparently, it's because p5e is only available via Capacity Blocks.

timsolovev: Right, I'll remove it from the filter then.

r4victor:
@solovyevt, if I have idle instances (not from any reservation) and specify any `reservation` in a run configuration, then the idle instances are suggested for the run (and they go first). Is it expected? Or should we only suggest offers belonging to the specified `reservation`?

I don't have a definitive stance on this, so let's go with whatever you think is appropriate.

src/dstack/_internal/core/backends/aws/compute.py
r4victor: @solovyevt, let us know if supporting "p5e" instances is critical to you. You can file an issue, and we'll prioritize it.

src/dstack/_internal/core/backends/aws/resources.py
```diff
@@ -596,3 +606,35 @@ def _is_private_subnet_with_internet_egress(
...
+    instance_types: List[str] = None,
+    is_capacity_block: bool = False,
+) -> Optional[Dict[str, Any]]:
+    filters = [{"Name": "state", "Values": ["active"]}]
+    if instance_types:
+        filters.append({"Name": "instance-type", "Values": instance_types})
+    try:
+        response = ec2_client.describe_capacity_reservations(
+            CapacityReservationIds=[reservation_id], Filters=filters
+        )
```
r4victor, line 624: If I pass `reservation_id` in invalid format (e.g. `some_res_1234`), the AWS will error

```
raise error_class(parsed_response, operation_name) 
                    botocore.exceptions.ClientError: An error occurred     
                    (InvalidCapacityReservationId.Malformed) when calling  
                    the DescribeCapacityReservations operation: Capacity   
                    Reservation ID 'res123456' is malformed
```

dstack will handle this error and won't crash, but it will be logged with error log level as unexpected AWS errors. It's recommended to catch such expected errors. They can be then re-raised as `ComputeError`.

r4victor:
@solovyevt, we'd like to make this PR into the tomorrow release. I can take care of polishing it a bit and merge if you're short on time.

timsolovev:
@r4victor Would appreciate your help, I'm not the strongest Python developer. Thanks!

r4victor:
@solovyevt, alright, I'll take care of merging it. Feel free to ask for help/assistance when you submit PRs.

But I still need your opinion on this

> if I have idle instances (not from any reservation) and specify any reservation in a run configuration, then the idle instances are suggested for the run (and they go first). Is it expected? Or should we only suggest offers belonging to the specified reservation?
I don't have a definitive stance on this, so let's go with whatever you think is appropriate.

timsolovev:
Since attributes such as `instance_types` are not used for hard filtering on the existing fleet, `reservation_id` probably shouldn't, too. Besides, `reservation_id` is a throwaway value that changes on every reservation, and updating it in task configurations in addition to fleet configurations is just another thing to trip over.

r4victor:
@solovyevt. Ok, let's keep it as is. 

Have you been able to test both Capacity Reservations and Capacity Blocks? I'm not able to test them ATM.

timsolovev:
@r4victor Good catch, it's been over a month since I've tested it with regular reservations. Patched one field check which does not exist for regular capacity reservations, but exists for capacity blocks (although [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CapacityReservation.html) explicitly mention it).

r4victor:
Ah, I'm not able to test Capacity Blocks but I can test Capacity Reservations. Let me try as well.

timsolovev:
Capacity Blocks worked perfectly fine when we used them last week. If there is something wrong - we have an upcoming block this and next week, so the issues can be patched then.

r4victor:
With some minor fixes, I got Capacity Reservation working for both run and fleet configurations. Merging it now.

@tsolovev, thanks a lot for the PR! Please test the Capacity Blocks with the release version when you'll be able to.

PR #1990, opened 2024-11-13
Add deployment section with vLLM, TGI and NIM. Remove alignment handbook

Bihan:
Add Deployment Section

List vLLM,TGI and NIM in deployment. Remove alignment handbook from fine-tuning section

Remove alignment-handbook directory from docs


docs/examples.md
```diff
@@ -13,6 +13,37 @@ hide:
...
+            Deploy Llama with vLLM
+        </p>
+    </a>
+    <a href="/examples/deployment/tgi" 
+       class="feature-cell">
+       <h3>
+           TGI
+       </h3>
+       <p>
+            Deploy Mistral with TGI
```
peterschmidt85: I'd suggest use Llama 3.1 instead of Mistral...

PR #2004, opened 2024-11-15
Implement per-user repo creds storage

### LINKED ISSUE #1992: [Bug]: Remote repo uses credentials of the last user who executed `dstack init`
### Steps to reproduce

Prerequisites:
* a git repo
* two users with their own git credentials

1. The first user executes `dstack init` inside the locally cloned repo (→ `dstack` client submits the repo info including their credentials to the server).
2. The second user executes `dstack init` inside their local copy of the same  repo.
3. The first user runs dev-environment and executes `git` commands (`git commit`, `git push`, etc.)

### Actual behaviour

The first user's run has git credentials of the second user.

### Expected behaviour

_No response_

### dstack version

0.18.25

### Server logs

_No response_

### Additional information

_No response_

un-def:
Fixes: https://github.com/dstackai/dstack/issues/1992


src/dstack/_internal/server/models.py
```diff
@@ -246,9 +246,30 @@ class RepoModel(BaseModel):
     type: Mapped[RepoType] = mapped_column(Enum(RepoType))
 
     info: Mapped[str] = mapped_column(Text)
+
+    # `creds` is deprecated, for newly initialized repos per-user `RepoCredsModel` should be used
+    # instead. As of 0.18.25, there is no plan to remove this field, it's used as a fallback when
+    # `RepoCredsModel` associated with the user is not found.
     creds: Mapped[Optional[str]] = mapped_column(String(5000))
```
jvstme, line 253: Maybe remove the field now? Apart from helping avoid legacy, this might be important from the security standpoint. To my understanding, the current solution will allow users to exploit older repos to access other project members' personal SSH keys. Removing the field will prevent it.

Ideally, there could be a data migration that moves the credentials to `repo_creds` for single-user projects, and drops them for multi-user projects, forcing users to re-run `dstack init`.

src/tests/_internal/server/services/test_repos.py
jvstme, line 1: Maybe run these tests against PostgreSQL too?

jvstme (approved):

src/dstack/_internal/server/models.py
un-def: We discussed this with the team, and it was decided to keep it and do nothing. We may drop this field in the future though. From a security standpoint, it's possible to wipe all repos (or their creds only) via REST API.

src/tests/_internal/server/services/test_repos.py
un-def: There is a module-global `pytestmark` for it

PR #2032, opened 2024-11-22
[Feature] Add alias `-R` for `--reuse` with `dstack apply` 

### LINKED ISSUE #2030: [Feature] Add alias `-R` for `--reuse` with `dstack apply`

peterschmidt85:
Fixes #2030

### DIFF
--- a/docs/docs/concepts/fleets.md
+++ b/docs/docs/concepts/fleets.md
@@ -1,7 +1,12 @@
 # Fleets
 
-Fleets enable efficient provisioning and management of clusters and instances, both in the cloud and on-prem. Once a
-fleet is created, it can be reused by dev environments, tasks, and services.
+By default, when you run `dstack apply` with a dev environment, task, or service,
+`dstack` reuses `idle` instances from an existing [fleet](concepts/fleets.md).
+If no `idle` instances matching the requirements, it automatically creates a new fleet 
+using backends.
+
+For more control over configuration and lifecycle management, or to set up fleets out of on-prem servers, use fleets
+directly.
 
 ## Define a configuration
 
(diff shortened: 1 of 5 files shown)


docs/docs/concepts/fleets.md
```diff
@@ -193,13 +193,30 @@ the hosts meet the requirements (see above).
...
+
+<div class="termy">
+
+```shell
+$ dstack apply -R -f examples/.dstack.yml
+```
+
+</div>
+
+Alternatively, set [`creation_policy`](reference/dstack.yml/dev-environment.md#creation_policy) to `reuse` in the task configuration.
```
jvstme: I'm not sure this section is supposed to be here, as creation policies are relevant for runs and this page is about fleets. Every paragraph is a bit confusing :point_down: 

> By default, `dstack apply` reuses `idle` instances from an existing \[fleet](../concepts/fleets.md),

This link references the current page itself.
It's also not clear that `dstack apply` here refers to applying run configurations, not fleet configurations discussed in this page.

> pass `-R` (or `--reuse`) to `dstack apply`.

Not clear that `--reuse` is an option when applying run configurations, not fleet configurations discussed in this page.

> Alternatively, set \[`creation_policy`](reference/dstack.yml/dev-environment.md#creation_policy) to `reuse` in the task configuration.

Not only in the task configuration, it can be set in service and dev-environment configurations too.

jvstme (approved):

docs/docs/concepts/fleets.md
peterschmidt85: Thank you! Fixed.

PR #1978, opened 2024-11-09
[UX]: live table when provisioning dstack configuration runs

Tob-iee:
fixes #1951

PR #2036, opened 2024-11-26
[UX]: live table when provisioning dstack configuration runs #1978

### LINKED ISSUE #1951: [UX] Show a live table in `dstack apply for provisioning runs (similar to fleets, volumes, gateways, etc)
**Problem:**

Currently, when I provision a run, `dstack apply` only shows job names:

Screenshot 2024-10-07 at 16 46 16

As a user, I want to see more information, esp. including chosen backend, region, price, resources, etc – all what `dstack ps` shows.

**Solution:**

Do the same as we do for other configurations such as volumes, gateways, and fleets – display the live `dstack ps` table:

Screenshot 2024-10-07 at 16 45 33

Tob-iee:
fixes #1951 @jvstme, yeah, you are right I didn't reflect the changes after merging dstackai:master into my branch. I re-opened it. Can you have another look? it should work fine


src/dstack/_internal/cli/services/configurators/run.py
```diff
@@ -221,6 +225,9 @@ def apply_configuration(
                         " Check `dstack ps` to see if it's done or failed."
                     )
                     exit(1)
+                    break
+                time.sleep(LIVE_TABLE_PROVISION_INTERVAL_SECS)
+                run.refresh()
```
jvstme: These lines seem to be redundant:
- `break` will never be reached after `exit(1)` on line 227
- `time.sleep` is redundant because the decision to reattach is already made on line 216
- `run.refresh` is redundant because the run was already refreshed on line 213

jvstme:
@Tob-iee, thanks for reopening the PR. This version looks and works great. I've made some minor tweaks and will merge the PR soon. Thanks for contributing!

PR #2048, opened 2024-12-02
Remove padding in `dstack apply` live tables

### LINKED ISSUE #2047: Remove padding in `dstack apply` live tables
### Current

Since a spinner was added in #2045, live tables in `dstack apply` have a 3-space left padding.

Screenshot From 2024-12-02 16-57-10

### Proposed

Remove the extra padding. All tables in `dstack` are padded with a single space. Also consider removing the empty line between the spinner and the table.

jvstme:
Fixes #2047

video

### DIFF
--- a/src/dstack/_internal/cli/utils/rich.py
+++ b/src/dstack/_internal/cli/utils/rich.py
@@ -1,8 +1,12 @@
 import logging
 from datetime import datetime
-from typing import TYPE_CHECKING, Callable, Iterable, List, Optional, Union
+from types import TracebackType
+from typing import TYPE_CHECKING, Callable, Iterable, List, Optional, Type, Union
 
+from rich.console import Group
+from rich.live import Live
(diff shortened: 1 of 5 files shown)


un-def (approved):

src/dstack/_internal/cli/utils/rich.py
```diff
@@ -122,3 +126,31 @@ def prepend_path(self, message, record):
...
+    def update(self, *renderables: "RenderableType") -> None:
+        self._live.update(renderable=Group(self._spinner, *renderables))
+
+    def __enter__(self) -> "MultiItemStatus":
+        self._live.start()
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
```
un-def: BTW, since we dropped 3.8 support, we can use stdlib types (`type` instead of `Type`) in new code

Issue #1535, opened 2024-08-11
Drop all custom docker image requirements


jvstme:
### Current

`dstack`  allows running custom Docker images by specifying them in the [`image`](https://dstack.ai/docs/reference/dstack.yml/dev-environment/#docker-image) property. However, not all images can be used. These are some of the image requirements:
- The software in the image should allow running as root
- The image should have either `apt-get` or `yum`
- The image should have `/bin/sh`
- etc.

### Proposed

Drop all image requirements and support all valid Docker images, including images built `FROM scratch`.

### Implementation notes

The main source of requirements seems to be the installation and configuration of the OpenSSH server. Possible solutions to dropping the requirements related to the OpenSSH server include:
- Shipping a statically-linked OpenSSH server binary that would allow running without root privileges and would not need a package manager for installation.
- Using an alternative SSH server implementation in Go, so that the server could be part of the `dstack-runner` binary.

jvstme:
Some examples of images that don't work and their respective errors:
- `nvcr.io/nim/meta/llama3-8b-instruct:latest` (or any other images with a non-root user) when run on **RunPod** or **Vast.ai** - never starts, killed by provisioning timeout
- `prom/prometheus` - `Error: Distribution not supported`
- `fedora` - `sed: can't read /root/.profile: No such file or directory`
- `gcr.io/etcd-development/etcd:v3.4.34` - `exec:
"/bin/sh": stat /bin/sh: no such file or directory: unknown`
- `bitnami/thanos` - `unable to find user root: no matching entries in passwd file`

un-def:
### Action Plan (WIP)

Completing the plan would allow to (at least):

* run non-root images on backends where we cannot override the container user (e.g., NIM on RunPod);
* run non-`deb`/`rpm`-based images.

#### Keep the default image user

* [x] Get `USER` from the Docker image, as it's already done for `ENTRYPOINT` and `CMD` (see `JobConfigurator`), store it as `JobSpec.user`.
* [x] (Optional) Add a new `user` property to the run configurations to override the default image user.
* [ ]  (Optional) If the `user` property is set and not equal to the default image user, exclude offers from backends where we cannot override the container user (RunPod, Vast.ai).
* [x] Start the container as `root` (if possible) to ensure that both the runner and the SSH server have sufficient permissions.
* [x] [runner] Execute the job with `Cmd.SysProcAttr.Credential.{Uid,Gid}` set according to the `JobSpec.user`.
[...]

PR #2055, opened 2024-12-03
Add `user` property to run configurations

un-def:
* If not set, use the default user from the image (if it, in turn, is not set either, Docker uses `root` as a default value)
* The container user is still set to `root`, as we need root privileges, at least to install sshd, but the runner executes the job (shell script with `commands` from the run configuration) as `user`.
* If the `user` is not root, it gets its own copy of `~/.ssh/authorized_keys` and `~/.ssh/environment`, making it possible to `ssh user@run-name` (the default user is still `root`, that is, `ssh run-name` logs in as root)
* `~/.ssh/environment` is now generated by the runner, not the outer shell script (container entrypoint), and includes all the same variables as the job env (including `DSTACK_*` vars and vars from the `env` property of the run configuration)

Part-of: https://github.com/dstackai/dstack/issues/1535


r4victor:
@un-def, could you clarify if containers running as non-root work with volumes? I'd assume that host directories would need to have appropriate permissions for the non-root user in order for volumes to work.

un-def:
@r4victor 

> if containers running as non-root work with volumes?

Good catch! FS permissions are addressed in https://github.com/dstackai/dstack/pull/2055/commits/728fbca5492f559252e77b101521cd330e3041c0

r4victor (approved):

PR #2057, opened 2024-12-04
[chore]: Move `run_async` to `_internal/utils`

jvstme:
Part of an upcoming refactoring in #1595, moved to a separate PR for a smaller diff and to avoid merge conflicts. `run_async` will be used both in `_internal/server` and `_internal/proxy`.


un-def:
It would be great to use `run_async` here too:

https://github.com/dstackai/dstack/blob/f277126a7ab1e53449e849e05f6e5d31b50704ac/src/dstack/_internal/core/services/ssh/tunnel.py#L204
https://github.com/dstackai/dstack/blob/f277126a7ab1e53449e849e05f6e5d31b50704ac/src/dstack/_internal/core/services/ssh/tunnel.py#L217

r4victor (approved):

PR #2058, opened 2024-12-05
Move server deps to dstack[server] extra

### LINKED ISSUE #2002: Support pip-installing dstack without server dependencies
Currently, the minimal dstack installation `pip install dstack` includes both client and server dependencies. For use cases when the user installs dstack only for CLI or API, which is the case for most CI/pipeline usages, this brings many unnecessary dependencies and also leads to problems with conflicting dependencies. If dstack is to be used as a library or a separate venv is not suitable (e.g. some Airflow installations), the users may be stuck with no workaround.

The proposal is to include only client dependencies in the minimal installation `pip install dstack`. There are much less client dependencies and they are also less likely to conflict with other packages. To install the client+server, the user would do `pip install dstack[server]`.

A possible downside is that users that install `pip install dstack`, won't be able to start a server. I see it as no problem for two reasons:
* The docs already recommend to do `pip install dstack[all]` for server installation.
* `dstack server` will emit an appropriate error if there no server dependencies instructing the user to do `pip install dstack[server]` or `pip install dstack[all]`.

Note that this issue is particularly important for enterprise installations where users have a centralized dstack server.

r4victor:
Closes #2002

The PR moves out server-specific dependencies from `pip install dstack` to `pip install "dstack[server]"`.

Breaking changes
* If users installed dstack server with `pip install dstack`, they won't be able to run `dstack server` anymore. But they'll get an appropriate error message about missing server deps. This use case should not be significant since dstack documentation instructs users to do `pip install "dstack[all]"` everywhere for server installation. The docs suggests to do `pip install dstack` for installing the CLI, which should bring less dependencies and possible conflicts now.


setup.py
```diff
@@ -42,35 +42,36 @@ def get_long_description():
     "rich-argparse",
     "tqdm",
     "simple-term-menu",
+    "pydantic>=1.10.10,<2.0.0",
+    "pydantic-duality>=1.2.4",
+    "websocket-client",
+    "watchfiles",
```
un-def: Isn't it an optional uvicorn (that is, server) dependency for hot code reloading?

un-def (approved):

PR #2075, opened 2024-12-09
[Minor] If errors should be passed silently, then in pythonic way

dimitriillarionov:
Was studying your code and propose more pythonic way for passing errors silently.
Not sure if this small change requires an issue or smth else, but let me know if it does. 

Cheers from Audi :)


r4victor:
@dimitriillarionov, thanks!

Issue #1780, opened 2024-10-03
[Feature] Support GPU blocks


peterschmidt85:
Motivation and implementation notes: https://docs.google.com/document/d/1-0cInSwV3CkybQUi4w4KtsrkFw5grvYdKxGo0gqPh2w/edit?usp=sharing

PR #2078, opened 2024-12-10
[shim] Implement multi-task state

un-def:
The shim is now able to process multiple tasks at a time. The global state for the existing "legacy" API is emulated. From the API caller's point of view, the shim works exactly the same as before.

### DIFF
--- a/runner/internal/shim/docker.go
+++ b/runner/internal/shim/docker.go
@@ -81,7 +81,11 @@ func (d *DockerRunner) Run(ctx context.Context, cfg TaskConfig) error {
 		return tracerr.Errorf("task %s is already submitted", task.ID)
 	}
 
-	defer func() { _ = d.tasks.Update(task) }()
+	defer func() {
+		if ok := d.tasks.Update(task); !ok {
+			log.Printf("failed to update task %s", task.ID)
+		}
+	}()
 
 	var err error
 
--- a/runner/internal/shim/docker.go
+++ b/runner/internal/shim/docker.go
@@ -212,20 +212,21 @@ func (d *DockerRunner) Stop(force bool) {
 	if !ok {
 		return
 	}
-	if task.Status == TaskStatusPulling {
+	switch task.Status {
+	case TaskStatusPending, TaskStatusCreating, TaskStatusTerminated:
+		// nothing to do
+	case TaskStatusPulling:
 		task.cancelPull()
-		return
-	}
-
-	stopOptions := container.StopOptions{}
-	if force {
-		timeout := int(0)
-		stopOptions.Timeout = &timeout
-	}
-
-	err := d.client.ContainerStop(context.Background(), task.containerID, stopOptions)
-	if err != nil {
-		log.Printf("Failed to stop container: %s", err)
+	case TaskStatusRunning:
+		stopOptions := container.StopOptions{}
+		if force {
+			timeout := int(0)
+			stopOptions.Timeout = &timeout
+		}
+		err := d.client.ContainerStop(context.Background(), task.containerID, stopOptions)
+		if err != nil {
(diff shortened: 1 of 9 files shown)


runner/internal/shim/docker.go
```diff
@@ -169,38 +178,38 @@ func (d *DockerRunner) Run(ctx context.Context, cfg TaskConfig) error {
 		}
 	}()
 
-	d.state = Running
-	d.containerID = containerID
-	d.containerName = cfg.ContainerName
-	log.Printf("Running container, name=%s, id=%s\n", d.containerName, containerID)
+	log.Printf("Running container, name=%s, id=%s\n", task.containerName, containerID)
```
r4victor, line 185: Have you considered switching to a level-aware logging like logrus used in the runner?

r4victor: Not pushing to include it in this PR, but it seems to make sense to do such changes when refactoring.

runner/internal/shim/docker.go
```diff
@@ -35,8 +37,10 @@ import (
 const ImagePullTimeout time.Duration = 20 * time.Minute
 
 const (
+	LabelKeyPrefix = "ai.dstack.shim."
 	// Set to "true" on containers spawned by DockerRunner, used for identification.
-	LabelKeyIsRun  = "ai.dstack.shim.is-run"
+	LabelKeyIsTask = LabelKeyPrefix + "is-task"
```
jvstme, line 42: (nit) Consider adding an alias, otherwise containers created by the old shim version won't be deleted by the new shim version

runner/internal/shim/docker.go
```diff
@@ -66,21 +64,33 @@ func NewDockerRunner(dockerParams DockerParameters) (*DockerRunner, error) {
...
+	// For legacy API compatibility, since LegacyTaskID is the same for all tasks
+	if task.ID == LegacyTaskID {
+		d.tasks.Delete(task.ID)
+	}
+
+	if ok := d.tasks.Add(task); !ok {
+		return tracerr.Errorf("task %s is already submitted", task.ID)
+	}
+
+	defer func() { _ = d.tasks.Update(task) }()
```
jvstme: Would it make sense to check for `false` here and log a message in case of errors?

r4victor (approved):

runner/internal/shim/docker.go
un-def: Logging facility revamping is in the plans, will be eventually submitted as a separate PR.

runner/internal/shim/docker.go
```diff
@@ -169,38 +178,38 @@ func (d *DockerRunner) Run(ctx context.Context, cfg TaskConfig) error {
...
-	if d.state == Pulling && d.containerID == "" {
-		d.cancelPull()
+	task, ok := d.tasks.Get(LegacyTaskID)
+	if !ok {
+		return
+	}
+	if task.Status == TaskStatusPulling {
+		task.cancelPull()
 		return
 	}
```
jvstme: Is any handling needed for the pending status too? Otherwise shim will attempt to stop a container that does not exist yet (line 222)

jvstme (approved):

runner/internal/shim/docker.go
un-def: I don't think backward compatibility is the issue here, since:

1. The only way to get a new `shim` version is to recreate the fleet.
2. Currently, fleet instances are destroyed and created from scratch when the fleet is recreated, thus it doesn't affect cloud instances at all.
3. There is normally only one (or zero) dead container from the last run. In the case of SSH fleet instances, if that dead container is a problem, it can be removed manually, it's a one-time action.

jvstme (approved):

PR #2087, opened 2024-12-12
Support gateway features in dstack-proxy

jvstme:
This commit adds all gateway features to dstack-proxy: registry API, config API, stats API, Nginx as reverse proxy, automatic certificate issuance. This is in preparation for running dstack-proxy on gateways, which will be enabled later in a separate PR. So far, the changes do not affect the use of dstack in any way.

Most of the code is taken from dstack-gateway and adapted to the dstack-proxy architecture. Some notable changes compared to dstack-gateway:
- Services are identified by their project name and run name, not run id. This allows the gateway and in-server environments to share more code. The registry and stats APIs are changed accordingly.
- All state is concentrated in the repo - singletons like Nginx or StatsCollector don't have persistent state.
- Since the Nginx singleton is now stateless, Nginx configs are constructed from the service state on each change, rather than keeping the current config version in memory and applying changes to it.
- Unit test coverage.

The commit also includes adjustments to the dstack-proxy architecture influenced by the addition of new features:
- Auth checks are moved from BaseProxyRepo to a separate dependency, BaseAuthProvider.
- BaseProxyRepo implementations are renamed to ServerProxyRepo and GatewayProxyRepo.
- The dependency injection mechanism is adjusted to allow gateway-only API routes to depend directly on GatewayProxyRepo. This allows to define gateway-specific repo methods in GatewayProxyRepo only. API routes common to both the gateway and in-server environments still depend on BaseProxyRepo.

The following is not part of this commit and will be added later:
- Actually running dstack-proxy on gateways.
- Restoring the application state after restarts.
- Server-side adjustments to accommodate for the registry and stats API changes.

Part of #1595


jvstme:
Sorry for the huge PR. Most of the diff is tests and things copied from dstack-gateway. I will comment on big files that were copied without many changes, hope this will help with the review a bit.

src/dstack/_internal/proxy/resources/nginx/00-log-format.conf
jvstme, line 1: All Nginx configs were copied as-is, only some variables were renamed

src/dstack/_internal/proxy/routers/auth.py
jvstme, line 1: Things in `proxy/routers` and `proxy/schemas` are the presentation layer of the proxy (gateway) API. They are very similar to dstack-gateway, but adapted to new interfaces + have a few minor API changes

src/dstack/_internal/proxy/services/nginx.py
jvstme, line 1: Copied with some refactoring, see commit message

src/dstack/_internal/proxy/services/registry.py
jvstme, line 1: Taken from dstack-gateway, but many changes to adapt to new interfaces. 88% test coverage

src/dstack/_internal/proxy/services/server_client.py
jvstme, line 1: Copied as-is

src/dstack/_internal/proxy/services/stats.py
jvstme, line 1: Copied with minor refactoring and a bugfix. 95% test coverage

src/tests/_internal/proxy/routers/test_registry.py
```diff
@@ -0,0 +1,416 @@
...
+@dataclass
+class Mocks:
+    reload_nginx: AnyMock
+    run_certbot: AnyMock
+    open_conn: AnyMock
+    close_conn: AnyMock
+
+
+@pytest.fixture
+def system_mocks() -> Generator[None, None, None]:
```
r4victor: Generator[None, None, None] => Generator[Mocks, None, None]

src/tests/_internal/proxy/routers/test_registry.py
```diff
@@ -0,0 +1,416 @@
...
+                https=False,
+                auth=False,
+                client_max_body_size=1024,
+            ),
+        )
+        assert resp.status_code == 200
+        assert resp.json() == {"status": "ok"}
+        conf = (tmp_path / "443-test-run.gtw.test.conf").read_text()
+        # general
+        system_mocks.reload_nginx.call_count == 1
```
r4victor: Should be assert or a comment?

src/tests/_internal/proxy/routers/test_registry.py
```diff
@@ -0,0 +1,416 @@
...
+        assert "server_name test-run.gtw.test;" in conf
+        assert "client_max_body_size 1024;" in conf
+        assert "listen 80;" in conf
+        # no https
+        system_mocks.run_certbot.call_count == 0
```
r4victor: Should be assert or a comment?

src/tests/_internal/proxy/routers/test_registry.py
r4victor: There are some more similar expressions in the file

src/dstack/_internal/proxy/main.py
```diff
@@ -0,0 +1,17 @@
...
+    formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+    handler = logging.StreamHandler()
+    handler.setFormatter(formatter)
+
+    logger = logging.getLogger("dstack")
+    logger.setLevel(level)
+    logger.addHandler(handler)
+
+
+configure_logging(logging.DEBUG)
```
r4victor: An env var for log level?

src/dstack/_internal/proxy/app.py
```diff
@@ -0,0 +1,93 @@
...
+
+    app.include_router(auth_router, prefix="/api/auth")
+    app.include_router(config_router, prefix="/api/config")
+    app.include_router(model_proxy_router, prefix="/api/models")
+    app.include_router(registry_router, prefix="/api/registry")
+    app.include_router(stats_router, prefix="/api/stats")
+
+    @app.get("/")
+    def get_info():
+        return {"version": __version__}
```
r4victor: Consider using `dstack._internal.DSTACK_VERSION` instead of `__version__`.

src/dstack/_internal/proxy/main.py
jvstme: Currently there is no way to deliver env variables to the gateway, so the logging level is the same as in the [old app](https://github.com/dstackai/dstack/blob/9d3f947299227a594718e00b1ce6b1938045130c/gateway/src/dstack/gateway/main.py#L44)

src/dstack/_internal/proxy/app.py
jvstme: `DSTACK_VERSION` is based on an env variable and there is no way to deliver env variables to the gateway, so I'd use `__version__` directly to prevent misunderstanding

jvstme:
Won't merge yet, still looking into ways to improve the architecture

r4victor (approved):

PR #2093, opened 2024-12-13
[shim] Rework resource management

un-def:
* Collect GPU resource identifiers: UUIDs for NVIDIA, DRI render nodes for AMD
* Request GPUs by identifiers when creating a container
* Keep track of busy/idle GPUs
* Move host-related code to a separate package
* Don't `panic()`


r4victor:
@un-def, how does requesting GPUs work with TPUs? Will a task have `gpu_count: 0`? No need to support GPU blocks for TPUs since it doesn't seem TPU Pods can be sliced further. Just want to ensure nothing breaks.

r4victor (approved):

un-def:
> how does requesting GPUs work with TPUs?

shim knows nothing about TPUs, for shim TPU instances are CPU instances, that is, gpuVendor is "none". 

> Will a task have gpu_count: 0

Yes

> it doesn't seem TPU Pods can be sliced further

Yes. As for now, we have no plans to allow TPU slices

Issue #1551, opened 2024-08-14
[Bug]: `dstack` marks instance as `terminated` without terminating it


jvstme:
### Steps to reproduce

```shell
> cat .dstack.yml 
type: dev-environment
ide: vscode

> dstack apply --spot-auto -b runpod -y
```
Then wait until the run is `running` and switch off the network on `dstack-server`'s host.

### Actual behaviour

The run is marked `failed`, the instance is marked `terminated`. However, the instance actually still exists in RunPod and the user is billed for it.

### Expected behaviour

The instance is not marked `terminated` until it is actually deleted in RunPod.

### dstack version

master

### Server logs

```shell
[09:21:18] DEBUG    dstack._internal.core.services.ssh.tunnel:73 SSH tunnel failed: b'ssh: connect to host 194.68.245.18 port 22056: Network is                
                    unreachable\r\n'                                                                                                                           
I0000 00:00:1723620079.263387 1605744 work_stealing_thread_pool.cc:320] WorkStealingThreadPoolImpl::PrepareFork
[09:21:19] DEBUG    dstack._internal.core.services.ssh.tunnel:73 SSH tunnel failed: b'ssh: connect to host 194.68.245.18 port 22056: Network is                
                    unreachable\r\n'                                                                                                                           
           WARNING  dstack._internal.server.background.tasks.process_running_jobs:259 job(e3ec13)polite-starfish-1-0-0: failed because runner is not available 
                    or return an error,  age=0:03:00.121137                                                                                                    
           INFO     dstack._internal.server.background.tasks.process_runs:338 run(5dd434)polite-starfish-1: run status has changed RUNNING -> TERMINATING      
[09:21:21] DEBUG    dstack._internal.server.services.jobs:238 job(e3ec13)polite-starfish-1-0-0: stopping container                                             
           INFO     dstack._internal.server.services.jobs:269 job(e3ec13)polite-starfish-1-0-0: instance 'polite-starfish-1-0' has been released, new status is
                    TERMINATING                                                                                                                                
           INFO     dstack._internal.server.services.jobs:286 job(e3ec13)polite-starfish-1-0-0: job status is FAILED, reason: INTERRUPTED_BY_NO_CAPACITY       
[09:21:22] INFO     dstack._internal.server.services.runs:932 run(5dd434)polite-starfish-1: run status has changed TERMINATING -> FAILED, reason: JOB_FAILED   
[...]

PR #2132, opened 2024-12-23
Add Vultr Support

Bihan:


src/dstack/_internal/server/services/config.py
```diff
@@ -428,6 +429,13 @@ class VastAIConfig(CoreModel):
     creds: Annotated[AnyVastAICreds, Field(description="The credentials")]
 
 
+class VultrConfig(CoreModel):
+    type: Annotated[Literal["vultr"], Field(description="The type of backend")] = "vultr"
+    regions: Optional[List[str]] = None
```
r4victor: Add Annotated – otherwise the field won't be shown in the docs reference.

src/dstack/_internal/server/migrations/versions/27d3e55759fa_add_pools.py
```diff
@@ -83,6 +83,7 @@ def upgrade() -> None:
                 "NEBIUS",
                 "TENSORDOCK",
                 "VASTAI",
+                "VULTR",
```
r4victor: You do not edit existing migrations since there will be no effect if the migrations are applied  – you need to add a new migration.

In this specific case, the migrations have to be generated for postgres. You need to set up a local postgres database, configure dstack to use it (e.g. `export DSTACK_DATABASE_URL="postgresql+asyncpg://postgres:postgres@localhost:5432/postgres"`), and then generate migrations with alembic (`alembic revision --autogenerate` from the server dir).

r4victor: Remove migrations edits and try adding a new migration. If you think you need help, let me know – I'll add the postgres migrations before merging.

Bihan: Got it. I will do the needful and push.

Bihan: @r4victor I've added a new migration. Please share your feedback or suggestions. If further updates are needed, I’ll make the necessary changes and push again.

Bihan:
@jvstme **Comments about Vultr VPC(Virtual Private Cloud) implementation.**


1. VPC Lifecycle During Instance Creation
Vultr allows the creation and attachment of a Virtual Private Cloud (VPC) during instance creation. However, if the instance creation fails due to no capacity, the created VPC is not automatically terminated. To address this, the implementation first creates the instance and, upon success, proceeds to create and attach the VPC to ensure proper cleanup and resource management.

2. Retry Logic for Attaching VPC
Vultr's API for attaching a VPC to an instance does not always succeed on the first attempt. To handle this limitation, a retry mechanism is implemented using a while loop that repeatedly attempts the VPC attachment until it is successful.

3. Unsupported VPC for Some Bare Metal Instances
Certain Bare Metal plans on Vultr do not support VPC attachment. In such cases, an exception is raised when the VPC attachment fails. As a fallback, instances are created without a VPC, and users are notified via a warning log: "Plan does not support private networking."

4. Workaround for Unreliable VPC Detachment API
Vultr's API for detaching VPCs from instances is unreliable, with detachment sometimes failing even after 15 minutes of retry. To mitigate this, a workaround checks for the successful termination of the instance before deleting the associated VPC. This is necessary because Vultr does not allow VPC deletion without either detaching it or terminating the instance. 

**Important:
To handle above limitations and constraints, I have implemented hardcoded error handling as below:**

```
def get_vpc_id(self, instance_id: str, plan_type: str) -> Optional[str]:
        try:
            if plan_type == "bare-metal":
                response = self._make_request("GET", f"/bare-metals/{instance_id}/vpcs")
            else:
                response = self._make_request("GET", f"/instances/{instance_id}/vpcs")

            vpcs = response.json().get("vpcs", [])
            if vpcs:
                return vpcs[0].get("id")
            return None
        except BackendError as e:
            # The above endpoints throws 404, when instance is not available.
            # Propagating this error blocks the execution, therefore need to handle here.
            if "Invalid instance-id" in str(e):
                logger.info(f"Instance {instance_id} not found.")
                return None
            else:
                raise
```

```
 def attach_vpc(self, vpc_id: str, instance_id: str, plan_type: str):
        data = {"vpc_id": vpc_id}
        try:
            if plan_type == "bare-metal":
                self._make_request("POST", f"/bare-metals/{instance_id}/vpcs/attach", data=data)
            else:
                self._make_request("POST", f"/instances/{instance_id}/vpcs/attach", data=data)
        except BackendError as e:
            # Sometimes instance is unable to attach VPC and throws 500 error.
            if "Unable to attach VPC" in str(e):
                logger.info("Unable to attach VPC")
            else:
                raise
```

```
    def delete_vpc(self, vpc_id: str):
        try:
            self._make_request("DELETE", f"/vpcs/{vpc_id}")
        except BackendError as e:
            if "The following servers are attached to this VPC network" in str(e):
                logger.info(f"Instance is connected to VPC {vpc_id}. Unable to delete VPC.")
                return None
            else:
                raise
```

```
def create_instance():
    ...
    ...
        while not self.api_client.get_vpc_id(instance_id, plan_type) and vpc_id is not None:
            time.sleep(1)
            # Vultr's limitation is that we cannot attach VPC without multiple attempts.
            logger.info("Attempting to attach instance to VPC")
            try:
                self.api_client.attach_vpc(vpc_id, instance_id, plan_type)
            except BackendError as e:
                if "plan does not support private networking" in str(e):
                    logger.warning("Plan does not support private networking.")
                    # delete created vpc
                    self.api_client.delete_vpc(vpc_id=vpc_id)
                    vpc_id = None
                else:
                    raise
```

In above functions, I am checking specific string in error message like "Invalid instance-id" or "plan does not support private networking". This has chances of breakage if Vultr changes the error message.

I am implementing such checks to ensure that only specific errors are handled within these functions and others are raised.

Please suggest, Thank you!

src/dstack/_internal/core/backends/vultr/api_client.py
```diff
@@ -28,6 +31,75 @@ def get_instance(self, instance_id: str, plan_type: str):
...
+    def attach_vpc(self, vpc_id: str, instance_id: str, plan_type: str):
+        data = {"vpc_id": vpc_id}
+        try:
+            if plan_type == "bare-metal":
+                self._make_request("POST", f"/bare-metals/{instance_id}/vpcs/attach", data=data)
+            else:
+                self._make_request("POST", f"/instances/{instance_id}/vpcs/attach", data=data)
+        except BackendError as e:
+            # Sometimes instance is unable to attach VPC and throws 500 error.
+            if "Unable to attach VPC" in str(e):
```
Bihan: Here I am checking specific error string, but doing so has a risk of breakage if Vultr changes it. Basically, I wanted to handle only the specific case of "Unable to attach VPC" in `attach_vpc` function and raise for others.

src/dstack/_internal/core/models/backends/vultr.py
```diff
@@ -0,0 +1,43 @@
...
+from pydantic.fields import Field
+from typing_extensions import Annotated, Literal
+
+from dstack._internal.core.models.backends.base import ConfigElement, ConfigMultiElement
+from dstack._internal.core.models.common import CoreModel
+
+
+class VultrConfigInfo(CoreModel):
+    type: Literal["vultr"] = "vultr"
+    project_id: str
```
jvstme: Is this field relevant to Vultr? Looks like it's not used in the code and I can't find anything about projects in Vultr docs.

Please remove the field if it is not needed. The same field is also defined in `VultrConfigInfoWithCredsPartial`, `VultrConfigValues`, and `VultrConfig`.

src/dstack/_internal/server/background/tasks/process_instances.py
```diff
@@ -911,4 +911,6 @@ def _get_instance_timeout_interval(
         return timedelta(seconds=1200)
     if backend_type == BackendType.OCI and instance_type_name.startswith("BM."):
         return timedelta(seconds=1200)
+    if backend_type == BackendType.VULTR:
+        return timedelta(seconds=1800)
```
jvstme: (nit) Do you know if such long timeouts are needed for all instances or only some specific ones? If it is only needed for some instances, consider narrowing down the `if` condition, e.g. the way it's done for Oracle bare metal instances above.
Same in `process_running_jobs.py`.

src/dstack/_internal/server/services/backends/configurators/vultr.py
```diff
@@ -0,0 +1,91 @@
...
+REGIONS = [
+    "no-luster-1",
+    "se-smedjebacken-1",
+    "gb-london-1",
+    "se-stockholm-1",
+    "us-newyork-1",
+    "us-santaclara-1",
+]
+
+DEFAULT_REGION = "no-luster-1"
```
jvstme: I don't think Vultr has such regions, these seem to be from CUDO.

I also don't think we need to have default Vultr regions at all. If the user doesn't specify `regions` in the backend configuration, just keep `None` or `[]` in the field and allow using all regions available in Vultr.

See TensorDock and Vast.ai as examples of backends without default regions.

src/dstack/_internal/core/backends/vultr/compute.py
```diff
@@ -0,0 +1,161 @@
...
+        self.config = config
+        self.api_client = VultrApiClient(config.creds.api_key)
+
+    def get_offers(
+        self, requirements: Optional[Requirements] = None
+    ) -> List[InstanceOfferWithAvailability]:
+        offers = get_catalog_offers(
+            backend=BackendType.VULTR,
+            requirements=requirements,
+        )
```
jvstme: Make sure to enforce `regions` if they are specified in the backend configuration.
```suggestion

        offers = get_catalog_offers(
            backend=BackendType.VULTR,
            locations=self.config.regions or None,
            requirements=requirements,
        )
```

jvstme: Please also add `extra_filter` to filter out unsupported offers, including those that may appear in `gpuhunt` in the future.

Offers worth excluding:
- Spot offers (`offer.instance.resources.spot == True`). I think Vultr may add spot instances in the future, in which case we wouldn't want to break existing `dstack` versions that don't support spot instances by releasing a new `gpuhunt` version that does.
- The `vc2-1c-0.5gb-free` - I believe only one such instance is available and it can only be run with the Free Tier. If most users won't be able to run it, I'd suggest to just filter it out.
- The GH200.
- As an alternative to blacklisting the two instances above, the preferred solution would actually be a whitelist of supported instances or "families" ([AWS example](https://github.com/dstackai/dstack/blob/87fe51fbbe36a1527236b21510f810dab4c29822/src/dstack/_internal/core/backends/aws/compute.py#L750)). Please see if it's possible to compose a concise whitelist for Vultr. If not, a blacklist will do.

src/dstack/_internal/core/backends/vultr/compute.py
```diff
@@ -0,0 +1,161 @@
...
+        try:
+            instance_id = self.api_client.launch_instance(
+                region=instance_offer.region,
+                label=instance_config.instance_name,
+                plan=instance_offer.instance.name,
+                startup_script=shim_commands,
+                public_keys=public_keys,
+            )
+        except Exception:
+            raise
```
jvstme: (nit) redundant

src/dstack/_internal/core/backends/vultr/api_client.py
```diff
@@ -0,0 +1,233 @@
...
+            else:
+                raise
+
+    def launch_instance(
+        self, region: str, plan: str, label: str, startup_script: str, public_keys: List[str]
+    ):
+        # Fetch or create startup script ID
+        script_id: str = self.get_startup_script_id(startup_script)
+        # Fetch or create SSH key IDs
+        sshkey_ids: List[str] = self.get_sshkey_id(public_keys)
```
jvstme: Vultr seems to support cloud-init - both [`POST /baremetals`](https://www.vultr.com/api/#tag/baremetal/operation/create-baremetal) and [`POST /instances`](https://www.vultr.com/api/#tag/instances/operation/create-instance) can accept the `user_data` property.

Do you think we could use cloud-init instead of Vultr's custom API for SSH keys and startup scripts? We are already using it for other backends, see [AWS](https://github.com/dstackai/dstack/blob/87fe51fbbe36a1527236b21510f810dab4c29822/src/dstack/_internal/core/backends/aws/compute.py#L229) for example.

I think using cloud-init would have many advantages:
- No keys or startup scripts will be stored in the Vultr account and shown in the web console
- Less API requests during instance launch
- Less code
- Identical to other backends, hence easier to maintain

src/dstack/_internal/core/backends/vultr/api_client.py
```diff
@@ -0,0 +1,233 @@
...
+                requests.codes.forbidden,
+                requests.codes.unauthorized,
+            ):
+                raise BackendInvalidCredentialsError(e.response.text)
+            if e.response is not None and e.response.status_code in (
+                requests.codes.bad_request,
+                requests.codes.internal_server_error,
+                requests.codes.not_found,
+            ):
+                raise BackendError(e.response.text)
```
jvstme, line 98: (nit) Maybe just use `BackendError` regardless of the status code? I can see that `bad_request` and `not_found` are documented in the API docs, but `internal_server_error` isn't, so who knows what other codes the API may decide to return

src/dstack/_internal/core/backends/vultr/api_client.py
```diff
@@ -0,0 +1,233 @@
...
+        elif "vcg" in plan:
+            # Ubuntu 22.04 will be installed. For gpu VMs, docker is preinstalled.
+            data = {
+                "region": region,
+                "plan": plan,
+                "label": label,
+                "os_id": 1743,
+                "script_id": script_id,
+                "sshkey_id": sshkey_ids,
+            }
```
jvstme: (nit)

> For gpu VMs, docker is preinstalled

Does this mean there will be some conflicts if we'll try using `"image_id": "docker"` for GPU instances? If not, we could use the same `image_id` and even the same request payload regardless of the instance type.

src/dstack/_internal/core/models/backends/vultr.py
Bihan: Removed Project ID

src/dstack/_internal/server/services/backends/configurators/vultr.py
Bihan: Done

jvstme:
@Bihan, I've just published gpuhunt 0.0.18 with Vultr support. I'd be great if you could update the gpuhunt version in `setup.py` and test this PR with the new version

Bihan:
> @Bihan, I've just published gpuhunt 0.0.18 with Vultr support. I'd be great if you could update the gpuhunt version in `setup.py` and test this PR with the new version.

Tested. It is working

src/dstack/_internal/server/background/tasks/process_instances.py
Bihan: Long timeouts are needed only for bare-metals. I update like oracle bare-metal as below:

 ```
if backend_type == BackendType.VULTR and instance_type_name.startswith("vbm"):
        return timedelta(seconds=1800)
```

File: src/dstack/_internal/core/backends/vultr/api_client.py (hunk 11 above)
Comment (Bihan) [reply src/dstack/_internal/core/backends/vultr/api_client.py]: @jvstme Yes, for GPU instances `"image_id": "docker"` is not accepted with below error.

```
{
    "error": "Server add failed: This Marketplace App is unable to be deployed to a GPU Workstation Instance. The vendor of this app has specified that this application can not be deployed to a GPU Workstation Instance.",
    "status": 400
}
```

**Comment by jvstme** [comment]
Thanks. Then please push the updated `setup.py` to this PR later with the rest of the changes

File: src/dstack/_internal/core/backends/vultr/api_client.py (hunk 10 above)
Comment (Bihan) [reply src/dstack/_internal/core/backends/vultr/api_client.py:98]: Attaching VPC to instance, sometimes throws 500 error. To handle this in `attach_vpc` as below, I included requests.codes.internal_server_error in BackendError.

May be later during VPC support, I can directly handle internal_server_error in `attach_vpc` without including it in BackendError type.

For now I will delete it.


```
def attach_vpc(self, vpc_id: str, instance_id: str, plan_type: str):
        data = {"vpc_id": vpc_id}
        try:
            if plan_type == "bare-metal":
                self._make_request("POST", f"/bare-metals/{instance_id}/vpcs/attach", data=data)
            else:
                self._make_request("POST", f"/instances/{instance_id}/vpcs/attach", data=data)
        except BackendError as e:
            # Sometimes instance is unable to attach VPC and throws 500 error.
            if "Unable to attach VPC" in str(e):
                logger.info("Unable to attach VPC")
            else:
                raise
                
```

File: src/dstack/_internal/core/backends/vultr/compute.py [hunk 12]
```diff
@@ -0,0 +1,350 @@
...
+        "vc2-2c-2gb",
+        "vc2-2c-2gb-sc1",
+        "vc2-2c-4gb",
+        "vc2-2c-4gb-sc1",
+        "vc2-4c-8gb",
+        "vc2-4c-8gb-sc1",
+        "vc2-6c-16gb",
+        "vc2-6c-16gb-sc1",
+        "vc2-8c-32gb",
+        "vc2-8c-32gb-sc1",
```
jvstme: Do you think we could use this shorter list of regexes instead? It's based on your full list.
```python
for family in [
    # Bare Metal - GPU
    r"vbm-\d+c-\d+gb-\d+-(a100|h100|l40|mi300x)-gpu",
    # Bare Metal - AMD CPU
    r"vbm-\d+c-\d+gb-amd",
    # Bare Metal - Intel CPU
    r"vbm-\d+c-\d+gb(-v\d+)?",
    # Cloud GPU
    r"vcg-(a16|a40|l40s|a100)-\d+c-\d+g-\d+vram",
    # Cloud Compute - Regular Performance
    r"vc2-\d+c-\d+gb(-sc1)?",
    # Cloud Compute - High Frequency
    r"vhf-\d+c-\d+gb(-sc1)?",
    # Cloud Compute - High Performance
    r"vhp-\d+c-\d+gb-(intel|amd)(-sc1)?",
    # Optimized Cloud Compute
    r"voc-[cgms]-\d+c-\d+gb-\d+s-amd(-sc1)?",
]:
    if re.fullmatch(family, offer.instance.name):
        return True
```

These regexes will allow some future instances to appear in dstack automatically if their ID looks similar to those that we already support, e.g. if Vultr adds a new CPU/RAM/Disk configuration for CPU instances.

Yet if Vultr adds something entirely different (ARM, new GPUs, etc) that may require testing and code changes, it will probably have an ID that won't match any of these regexes and won't appear in dstack before we test it and add to the list manually.

src/dstack/_internal/server/background/tasks/process_instances.py
```diff
@@ -911,4 +911,6 @@ def _get_instance_timeout_interval(
         return timedelta(seconds=1200)
     if backend_type == BackendType.OCI and instance_type_name.startswith("BM."):
         return timedelta(seconds=1200)
+    if backend_type == BackendType.VULTR and instance_type_name.startswith("vbm"):
+        return timedelta(seconds=1800)
```
jvstme: 30 minutes is a huge timeout already, but apparently it's not enough - in my testing, some bare metal instances took almost 40 minutes to start. And I haven't tested H100 and MI300X yet, so maybe some instances will take even longer. Additionally, the timeout (at least the one in `process_running_jobs`) should account for the pulling of the Docker image, which can also take a few minutes for large images.

Not sure if we can do anything about it but increase the timeout even more. It's important to give the instance enough time to start, otherwise dstack will terminate it, but the user will still be charged for the wasted provisioning time.

I can suggest 55 minutes. Not 60 to avoid being charged for the second hour, since Vultr charges hourly.

src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -656,4 +656,6 @@ def _get_runner_timeout_interval(backend_type: BackendType, instance_type_name:
         return timedelta(seconds=1200)
     if backend_type == BackendType.OCI and instance_type_name.startswith("BM."):
         return timedelta(seconds=1200)
+    if backend_type == BackendType.VULTR:
+        return timedelta(seconds=1800)
```
jvstme: Could you add the `instance_type_name.startswith("vbm")` check here too?

And also increase the timeout here if you will increase it in `process_instances.py`

src/dstack/_internal/core/backends/vultr/api_client.py
```diff
@@ -224,10 +151,4 @@ def _make_request(self, method: str, path: str, data: Any = None) -> Response:
                 requests.codes.unauthorized,
             ):
                 raise BackendInvalidCredentialsError(e.response.text)
-            if e.response is not None and e.response.status_code in (
-                requests.codes.bad_request,
-                requests.codes.internal_server_error,
-                requests.codes.not_found,
-            ):
-                raise BackendError(e.response.text)
```
jvstme: Since this error handling is now removed, I sometimes see these errors during provisioning:
```
  File "/src/dstack/_internal/core/backends/vultr/compute.py", line 72, in create_instance
    instance_id = self.api_client.launch_instance(
  File "/src/dstack/_internal/core/backends/vultr/api_client.py", line 64, in launch_instance
    resp = self._make_request("POST", "/instances", data)
  File "/src/dstack/_internal/core/backends/vultr/api_client.py", line 85, in _make_request
    response.raise_for_status()
  File "/venv/lib64/python3.9/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://api.vultr.com/v2/instances
```
In particular, this happens when there is no capacity in a specific Vultr region. Since this is an expected case, it shouldn't result in exceptions and trigger Sentry alerts.

So I guess we'll have to bring back this error handling. Maybe just without handling `internal_server_error` as it was only needed for VPCs. Raising `BackendError` is fine, it will be handled by dstack and logged as a warning.

jvstme (commented):
@Bihan, I've tested the PR locally and everything works really well! I've only encountered a couple of small issues with some edge cases, please see my comments.

We are almost ready to merge the PR, so it'd also be great if you could resolve the merge conflict in `setup.py` by rebasing onto master or merging master into the PR branch.

src/dstack/_internal/server/background/tasks/process_instances.py
Bihan: @jvstme Just thinking out loud. Should we provide warning to users when provisioning bare-metals? May be we can warn client in the logs that "Vultr might take more than 30 mins for setup".

src/dstack/_internal/core/backends/vultr/compute.py
Bihan: This is much better implementation. Thank you

src/dstack/_internal/server/background/tasks/process_instances.py
jvstme: Yes, I think a message in `dstack apply` would be informative. Although this seems a bit tricky to implement at the moment. In general, we don't know which instance will be assigned to the job before provisioning starts, it may end up not being bare metal or Vultr. And even after the instance is assigned, the CLI can't tell if it is actually being provisioned or just reused from an existing fleet.

jvstme (approved):
@Bihan, thank you for the update, the PR looks good to me.

I found one more issue though - quite often instance termination fails if you cancel the provisioning with Ctrl+C shortly after it has started:

```
[22:32:37] ERROR    dstack._internal.server.background.tasks.process_instances:801 Failed to terminate instance cloud-0: BackendError('{"error":"Unable to destroy server: This subscription is not currently active, you cannot destroy it.","status":500}')
```
... or shortly after it has finished:
```
[22:28:07] ERROR    dstack._internal.server.background.tasks.process_instances:801 Failed to terminate instance cloud-0: BackendError('{"error":"Unable to destroy server: Unable to remove VM: Server is currently locked","status":500}')
```
This is pretty critical, since dstack then forgets about the instance while it keeps running in Vultr, possibly causing unexpected charges for the user.

However, I think this should be fixed by adding termination retries in the dstack server code, not in the backend implementation. So I'd suggest to address this issue separately as part of #1551. We'll discuss this with the rest of the team on Monday.

jvstme:
Will add termination retries in #1551

PR #2141, opened 2024-12-25
[shim] Implement Future API

un-def:
* Added Future API endpoints `/api/tasks{/...}`
* Added a mechanism to restore shim state on restarts
* Reworked error reporting —  methods now return `error` instead `bool` or (`bool`, `error`) with error category (`ErrNotFound`, `ErrInternal`, etc.) wrapped.
* Simplifed API routing — a new `Router` struct with a simpler interface added
* CPU and memory resource limits added

### DIFF
--- a/runner/internal/shim/api/server.go
+++ b/runner/internal/shim/api/server.go
@@ -10,11 +10,16 @@ import (
 )
 
 type TaskRunner interface {
-	Run(context.Context, shim.TaskConfig) error
-	GetState() (shim.RunnerStatus, shim.JobResult)
-	Stop(bool)
+	Submit(context.Context, shim.TaskConfig) error
+	Run(ctx context.Context, taskID string) error
+	Terminate(ctx context.Context, taskID string, timeout uint, reason string, message string) error
+	Remove(ctx context.Context, taskID string) error
 
 	Resources() shim.Resources
+	TaskIDs() []string
+	TaskInfo(taskID string) shim.TaskInfo
+
+	GetState() (shim.RunnerStatus, shim.JobResult)
 }
 
 type ShimServer struct {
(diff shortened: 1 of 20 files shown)


runner/internal/shim/api/server.go
```diff
@@ -27,25 +32,33 @@ type ShimServer struct {
...
-	mux.HandleFunc("/api/pull", api.JSONResponseHandler("GET", s.PullGetHandler))
-	mux.HandleFunc("/api/stop", api.JSONResponseHandler("POST", s.StopPostHandler))
+	r.AddHandler("GET", "/api/healthcheck", s.HealthcheckHandler)
+
+	// Future API
+	r.AddHandler("GET", "/api/tasks", s.TaskListHandler)
+	r.AddHandler("GET", "/api/tasks/{id}", s.TaskInfoHandler)
+	r.AddHandler("PUT", "/api/tasks", s.TaskSubmitHandler)
+	r.AddHandler("POST", "/api/tasks/{id}/terminate", s.TaskTerminateHandler)
+	r.AddHandler("DELETE", "/api/tasks/{id}", s.TaskRemoveHandler)
```
r4victor: Suggestion: switch from REST-like API to RPC (replace with POSTs) since it's more flexible for internal APIs and doesn't carry extra semantics.

runner/internal/shim/task.go
```diff
@@ -4,17 +4,19 @@ import (
...
 )
 
 type TaskStatus string
 
 const (
-	// pending -> pulling -> creating -> running -> terminated
-	//    |         |           |
-	//    v         v           v
-	// terminated terminated terminated
+	// pending -> preparing -> pulling -> creating -> running -> terminated
```
r4victor, line 14: @un-def, could you elaborate why preparing status is needed?

r4victor (approved):

runner/internal/shim/task.go
un-def: The idea was to split `Run()` method into two — submission and run operations, `Submit()`, which should work as fast as possible, only puts a task into storage with `pending` status and returns an error on conflict, which is returned to the server via HTTP API. `Run()`, in turn, performs some preparation steps before it can start a container (puts authorized_keys, prepares volumes, etc.). Previously, these steps were executed with `pending` status (and with legacy API — `pulling` state), which is not quite right semantically.

In short, `penging` has not `Run()` yet, `preparing` — just first status in the `Run()` sequence (prepare, pull image, create container, run, mark terminated)

PR #2143, opened 2024-12-25
Fix certbot process getting stuck in dstack-proxy

jvstme:
**Problem**: if obtaining a certificate times out, `subprocess.run` cancels the certbot command with
SIGKILL. However, the command is run with sudo, so SIGKILL actually kills the sudo process, while its child certbot process is adopted by init and keeps running indefinitely. This breaks adding or
updating services on the gateway, since certbot
refuses to run if another certbot process exists.

Attempts to implement graceful cancelling by
sending SIGTERM to the sudo process were not
successful - for some reason sudo ignores SIGTERM
originating from its parent.

The solution in this commit uses the `timeout`
shell command to set the timeout on the actual
certbot process instead of the sudo process.

#1595

### DIFF
--- a/src/dstack/_internal/proxy/gateway/services/nginx.py
+++ b/src/dstack/_internal/proxy/gateway/services/nginx.py
@@ -14,7 +14,8 @@ from dstack._internal.proxy.lib.errors import ProxyError, UnexpectedProxyError
 from dstack._internal.utils.common import run_async
 from dstack._internal.utils.logging import get_logger
 
-CERTBOT_TIMEOUT = 30
+CERTBOT_TIMEOUT = 40
+CERTBOT_2ND_TIMEOUT = 5
 CONFIGS_DIR = Path("/etc/nginx/sites-enabled")
 logger = get_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


un-def (approved):

un-def:
I think the core reason why `subprocess.run()` won't terminate the whole process subtree is that on timeouts it calls `Popen.kill()`, which, as the name implies, sends `SIGKILL`, and `sudo` has no chance to propagate the signal to its children: https://github.com/python/cpython/blob/067145177975eadd61a0c907d0d177f7b6a5a3de/Lib/subprocess.py#L554-L558

`man 8 sudo` (*emphasis* added): 

>  Signal handling
>
> When the command is run as a child of the sudo process, sudo will relay signals it receives to the command.  [...]  *Some signals, such as SIGSTOP and SIGKILL, cannot be caught and thus will not be relayed to the command*. [...]

`Popen()` + `communiate()` + `terminate()` instead of the wrapper (`run()`) should fix the issue without external tools, I believe, but I'm not insisting that is should be fixed in that way, just providing some context; the current solution is fine.

jvstme:
I tried it first, but it didn't work for some reason.
> Attempts to implement graceful cancelling by
sending SIGTERM to the sudo process were not
successful - for some reason sudo ignores SIGTERM
originating from its parent.

I can terminate the sudo process manually by sending SIGTERM from a shell, but when the proxy application was trying to do the same thing with `terminate()` nothing would happen.

PR #2146, opened 2024-12-26
Optimize `dstack-gateway` installation time

jvstme:
Add the `dstack[gateway]` extra, so that the
gateway doesn't install `dstack[server]`.
`dstack[gateway]` still has some unnecessary
dependencies from BASE_DEPS, but the current
optimization is enough for now.

Also increase the gateway installation timeout
just in case.

#1595


gateway/pyproject.toml
```diff
@@ -11,7 +11,7 @@ requires-python = ">=3.10"
 dynamic = ["version"]
 dependencies = [
     # release builds of dstack-gateway depend on a PyPI version of dstack instead
-    "dstack[server] @ https://github.com/dstackai/dstack/archive/refs/heads/master.zip",
+    "dstack[gateway] @ https://github.com/dstackai/dstack/archive/refs/heads/master.zip",
```
r4victor, line 14: Seems like a fragile solution since there are no tests that run with `dstack[gateway]`.

r4victor (approved):

gateway/pyproject.toml
jvstme: I will merge for now and think how we can test it.
Although we also don't test CLI or individual backends without `[all]`, so it is not much more fragile than the existing extras.

PR #2156, opened 2024-12-29
Add API v2 compat support to ShimClient

un-def:
From the caller's point of view, ShimClient interface stays the same, but internally it's now able to work with both API v1 and v2, emulating API v1 logic via API v2 methods if v2 is supported by shim

### DIFF
--- a/src/dstack/_internal/server/schemas/runner.py
+++ b/src/dstack/_internal/server/schemas/runner.py
@@ -1,4 +1,5 @@
 from base64 import b64decode
+from enum import Enum
 from typing import Dict, List, Optional, Union
 
 from pydantic import Field, validator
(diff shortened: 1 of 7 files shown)


src/dstack/_internal/server/schemas/runner.py
```diff
@@ -94,13 +95,55 @@ class ShimVolumeInfo(CoreModel):
...
+    gpu: int
+    cpu: float
+    memory: int
+    shm_size: int
+    volumes: list[ShimVolumeInfo]
+    volume_mounts: list[VolumeMountPoint]
+    instance_mounts: list[InstanceMountPoint]
+    host_ssh_user: str
+    host_ssh_keys: list[str]
+    container_ssh_keys: List[str]
```
r4victor:
```suggestion
    container_ssh_keys: list[str]
```

r4victor (approved):

Issue #2150, opened 2024-12-27
Show deleted runs in the UI


r4victor:
Viewing deleted runs is useful for auditing and debugging, especially for global admins. Currently, if a run is deleted or overridden, it's not returned by the `/api/runs/list` endpoint and, thus, not shown in the UI. The API should return such runs, and the UI should show them ~~with status Deleted~~ as regular runs with their final status (deleted runs always have final statuses). So basically, the UI should not differentiate between deleted and non-deleted runs, and the Delete button can be also removed.

To support viewing deleted runs details, the UI should be able to get a deleted run. Currently, it's not possible because the /get endpoint returns only active runs. The `/get` endpoint can be changed to accepts a run id. In this case, the UI can include a run id in the routes (e.g. `projects/my-project/runs/slimy-rabbit-1?id=0d0f42c5-a242-4b35-aa0e-862ef21d7b28`).

For other resources, the UI doesn't show detailed pages, but their `/get` endpoints can be adjusted in a similar manner.

Implementation plan:
1. (The server) Support getting deleted runs by run id.
2. (The UI) Support viewing deleted runs details (route runs by id). The old name-based routing can remain for non-deleted runs.
3. (The server) Include deleted runs in `/api/runs/list`.
4. (The UI) Remove the Delete button.

peterschmidt85:
See https://github.com/dstackai/dstack/pull/2157#issuecomment-2565944058

I get if I run a new CLI with an old server:

```
Server validation error: 
{'detail': [{'loc': ['body', 'id'],
             'msg': 'extra fields not permitted',
             'type': 'value_error.extra'}]}
```

PR #2157, opened 2024-12-30
Add API support to get runs by id

r4victor:
Part of #2150 

This PR:
* Extends `/api/project/{project.name}/runs/get` to return runs by id (included deleted runs).
* Add `deleted` field to run API response.


peterschmidt85:
@r4victor Is this backward compatible? Try running the new CLI with an old server

PR #2158, opened 2024-12-30
Return deleted runs in /api/runs/list

r4victor:
Part of #2150 

The PR changes `/api/runs/list` behavior to include deleted runs by default. If `only_active: true` is set, deleted runs are filtered out as before because deleted runs always have finished statuses. 

The change also affects the CLI: `dstack ps -a` now includes deleted (overridden) runs as well:

```
✗ dstack ps -a
 NAME                  BACKEND       RESOURCES    PRICE     STATUS      SUBMITTED   
 my-dev                aws           2xCPU, 8GB,  $0.0447   running     3 mins ago  
                       (eu-west-1)   100.0GB                                        
                                     (disk),                                        
                                     SPOT                                           
 my-dev                aws           2xCPU, 8GB,  $0.0447   terminated  4 mins ago  
                       (eu-west-1)   100.0GB                                        
                                     (disk),                                        
                                     SPOT                                                            
```

Before merging this PR, the UI should be update to show run details page for deleted runs based on id (now different runs within a project can have the same name because deleted runs are returned).


peterschmidt85:
@r4victor is the task pending on @olgenn ?

r4victor:
@peterschmidt85, yes: https://github.com/dstackai/dstack/issues/2150

PR #2167, opened 2024-12-30
Replace `termination_policy` and `termination_idle_time` with `idle_duration: int|str|off`

### LINKED ISSUE #1154: Replace `termination_policy` and `termination_idle_time` with `idle_duration: int|str|off`
I suggest that we drop `termination_policy` but rename `termination_idle_time` to `idle_duration`. 
`termination_idle_time` is an important fieled but currently too long to be convenient to use. Also, I suggest that we use duration everywhere instead of somewhere `time` and somewhere `duration`.

peterschmidt85:
Fixes #1154


src/dstack/_internal/core/models/fleets.py
```diff
@@ -172,18 +172,33 @@ class InstanceGroupParams(CoreModel):
         Optional[float],
         Field(description="The maximum instance price per hour, in dollars", gt=0.0),
     ] = None
+
+    idle_duration: Annotated[
+        Optional[Union[Literal["off"], str, int]],
+        Field(
+            description="Time to wait before terminating idle instances. Defaults to `5m` for runs and `3d` for fleets. Use `off` for unlimited duration"
```
un-def, line 179: Not sure if `off` is better than `-1`. For me, `off` is ambiguous: does it mean “no duration” (= unlim, = -1, = don't destroy) or “no idle” (= 0, = destroy immediately after use). I personally would prefer `-1` for “never destroy”, `0` for “destroy immediately” and `> 0 ` for “destroy if idle for N days (minutes, hours, …)”

src/dstack/api/server/__init__.py
```diff
@@ -143,6 +143,8 @@ def _request(
                     code = kwargs.pop("code")
                     raise _server_client_errors[code](**kwargs)
             if resp.status_code == 422:
+                print(path)
+                print(body)
```
un-def: debug prints

src/dstack/api/server/_fleets.py
```diff
@@ -73,6 +73,9 @@ def _get_fleet_spec_excludes(fleet_spec: FleetSpec) -> Optional[_ExcludeDict]:
     if not fleet_spec.configuration.reservation:
         configuration_excludes["reservation"] = True
         profile_excludes.add("reservation")
+    if not fleet_spec.configuration.idle_duration:
```
un-def: Both `None` and `0` are falsey. Shouldn't this check be `is not None`?

src/dstack/api/server/_runs.py
```diff
@@ -132,6 +132,9 @@ def _get_run_spec_excludes(run_spec: RunSpec) -> Optional[dict]:
     if not configuration.reservation:
         configuration_excludes.add("reservation")
         profile_excludes.add("reservation")
+    if not configuration.idle_duration:
```
un-def: Same here

src/dstack/_internal/core/models/fleets.py
peterschmidt85: In general I agree about `-1` being potentially better than `off`. On the other hand:
a) We already support `off` with `max_duration`
b) `-1` is a bit too technical
Perhaps, we could later support t`-1` too as a synonym for `off`. Okay?

un-def: a) Ah, I see
b) Agree, but, as far as I understand, `-1` is already works the same way as `off`, it's just not documented. We can leave it as is for now

src/dstack/api/server/__init__.py
peterschmidt85: Removed as body can contain sensitive information

src/dstack/api/server/_fleets.py
un-def: UPD seems not to be the case, yaml ints are casted to strings, `idle_duration: 0` in YAML → `'0'` in Pydantic model.

un-def (approved):

src/dstack/api/server/_runs.py
un-def: UPD seems not to be the case

src/dstack/api/server/_fleets.py
peterschmidt85: Fixed

src/dstack/api/server/_runs.py
peterschmidt85: Fixed

PR #2176, opened 2025-01-06
[shim] Revamp logging and CLI

un-def:
* Migrate to context-aware logging facility
* Write to `shim.log` internally (don't rely on systemd/shell std{out,err} redirects, preserves shim logs in journald)
* Add `--shim-log-level`
* Rename `--home` to `--shim-home` for consistency
* Drop `docker` command (it is _the only_ command)

### DIFF
--- a/runner/internal/shim/api/server.go
+++ b/runner/internal/shim/api/server.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	"net"
 	"net/http"
 	"sync"
 
@@ -15,7 +16,7 @@ type TaskRunner interface {
 	Terminate(ctx context.Context, taskID string, timeout uint, reason string, message string) error
 	Remove(ctx context.Context, taskID string) error
 
-	Resources() shim.Resources
+	Resources(context.Context) shim.Resources
 	TaskIDs() []string
 	TaskInfo(taskID string) shim.TaskInfo
 }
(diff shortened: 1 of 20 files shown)


runner/internal/shim/task.go
```diff
@@ -45,19 +46,19 @@ type Task struct {
 
 // Lock is used for exclusive operations, e.g, stopping a container,
 // removing task data, etc.
-func (t *Task) Lock() {
+func (t *Task) Lock(ctx context.Context) {
 	if !t.mu.TryLock() {
-		log.Fatalf("task %s already locked!", t.ID)
+		log.Fatal(ctx, "already locked!", "task", t.ID)
```
r4victor, line 51: log.Fatal calls exit(1) and defers do not run, i.e. basically panics. Is it intended here?

un-def: > log.Fatal calls exit(1) and defers do not run

Didn't know that, but in this specific case it's fine I think, this should never happen, fatal here indicates a serious  programmatic error (which otherwise would lead to a deadlock, as mutexes are not reentrant) and should abort execution, even if not gracefully

runner/internal/shim/api/server.go
```diff
@@ -29,12 +30,13 @@ type ShimServer struct {
...
-func NewShimServer(address string, runner TaskRunner, version string) *ShimServer {
+func NewShimServer(ctx context.Context, address string, runner TaskRunner, version string) *ShimServer {
 	r := api.NewRouter()
 	s := &ShimServer{
 		HttpServer: &http.Server{
-			Addr:    address,
-			Handler: r,
+			Addr:        address,
+			Handler:     r,
+			BaseContext: func(l net.Listener) context.Context { return ctx },
```
r4victor, line 39: Does it mean the same context will be used for all requests? I assume it will be a problem if we need to store per-request values in context.

r4victor (approved):

runner/internal/shim/api/server.go
un-def: If I understand correctly, base context (as its name implies) is not used directly but to derive a new per-request context: https://github.com/golang/go/blob/194de8fbfaf4c3ed54e1a3c1b14fc67a830b8d95/src/net/http/server.go#L3318-L3328

PR #2191, opened 2025-01-15
Implement bridge network mode for jobs

un-def:
Allows running multiple containers per host

shim:

* Add `network_mode`: `host`|`bridge` to submit request
* Extract port mapping in `bridge` mode, expose via API as `ports`
* Add `--runner-ssh-port` (see below about hardcoded ports)
* Allow terminated->terminatad transition (fixes server flood in logs)

server:

* Add `JobRuntimeData` structure holding various runtime-specific job info (resource shares, network mode, port mapping)
* Use port mapping retrieved from shim to access runner from server and services from gateway
* `ShimClient`: implement v2 API, drop v1-over-v2 API emulation
* Replace hardcoded numeric ports with constants, deliver port settings to shim via `shim.env` file (don't rely on defaults in Go codebase)


src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -631,6 +675,12 @@ def _submit_job_to_runner(
         instance_env = remote_info.env
     else:
         instance_env = None
+
+    runner_client = client.RunnerClient(port=ports[DSTACK_RUNNER_HTTP_PORT])
+    resp = runner_client.healthcheck()
+    if resp is None:
+        return True  # runner is not available yet
```
r4victor: Shouldn't it return False? Otherwise, it seems like the job will never be terminated due to WAITING_RUNNER_LIMIT_EXCEEDED.

r4victor (approved):

src/dstack/_internal/server/background/tasks/process_running_jobs.py
un-def: Right, it's a bug introduced during refactoring. In one case (`_process_provisioning_no_shim`) we should return False, and True in the other (`_process_pulling_with_shim`):

https://github.com/dstackai/dstack/blob/5c6891ada3a2444f741053e386fcc8ad6af3605a/src/dstack/_internal/server/background/tasks/process_running_jobs.py#L370-L371
https://github.com/dstackai/dstack/blob/5c6891ada3a2444f741053e386fcc8ad6af3605a/src/dstack/_internal/server/background/tasks/process_running_jobs.py#L518-L519

Fixed with parametrization: https://github.com/dstackai/dstack/pull/2191/commits/15c37e67fe1a238809a5d8a15c7f2db3d198d47a

PR #2197, opened 2025-01-17
Implement offers cache

### LINKED ISSUE #2195: Optimize get_offers with cache
Currently, Compute.get_offers() gets called for every run plan requests and for every job/instance provisioning. If I submit 20 runs, dstack would request offers 40 times (!) in a very short period of time. Besides the inefficiency, it sometimes causes cloud rate limiting. AWS, for example, may stop accepting requests for 10-20m if there are too many requests to list service quotas. Since dstack does not implement batch provisioning, this issue is crucial for provisioning large cloud clusters.

I suggest using a simple TTL cache for get_offers with a short TTL like 1 minute. It should be enough to avoid requesting the offers multiple times for the same run.

Notes:
* The cache maxsize should be small since offers may eat memory.
* It's probably better not to cache empty results so that if a user didn't setup something, they could fix it and see the results immediately. On the other hand, updating a backend will invalidate the cache anyway, so maybe caching empty results is ok as well.

r4victor:
Closes #2195 

The PR implements a simple TTL 30s cache for Compute.get_offers().

Some other improvements:
* AzureCompute.get_offers() is optimized to get quotas for different regions in parallel – reduced get_offers time for all regions/offers from ~30s to ~8s.
* SQLalchemy pool_size and max_overflow are made configurable with bigger default (10 vs 5). max_instances for background tasks is reduced to 2. This is all to prevent a possible SQLalchemy's queue limit overflow – I got it when provisioning > 30 instances at once – process_instances tasks held all the db connections for too long since it's slow when provisioning.

### DIFF
--- a/src/dstack/_internal/server/settings.py
+++ b/src/dstack/_internal/server/settings.py
@@ -26,6 +26,10 @@ LOG_FORMAT = os.getenv("DSTACK_SERVER_LOG_FORMAT", "rich").lower()
 ALEMBIC_MIGRATIONS_LOCATION = os.getenv(
     "DSTACK_ALEMBIC_MIGRATIONS_LOCATION", "dstack._internal.server:migrations"
 )
+# Users may want to increase pool size to support more concurrent resources
+# if their db supports many connections
+DB_POOL_SIZE = int(os.getenv("DSTACK_DB_POOL_SIZE", 10))
+DB_MAX_OVERFLOW = int(os.getenv("DSTACK_DB_MAX_OVERFLOW", 10))
 
 SERVER_CONFIG_DISABLED = os.getenv("DSTACK_SERVER_CONFIG_DISABLED") is not None
 SERVER_CONFIG_ENABLED = not SERVER_CONFIG_DISABLED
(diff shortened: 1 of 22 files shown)


src/dstack/_internal/server/settings.py
```diff
@@ -26,6 +26,10 @@
 ALEMBIC_MIGRATIONS_LOCATION = os.getenv(
     "DSTACK_ALEMBIC_MIGRATIONS_LOCATION", "dstack._internal.server:migrations"
 )
+# Users may want to increase pool size to support more concurrent resources
+# if their db supports many connections
+DB_POOL_SIZE = int(os.getenv("DSTACK_DB_POOL_SIZE", 10))
+DB_MAX_OVERFLOW = int(os.getenv("DSTACK_DB_MAX_OVERFLOW", 10))
```
un-def, line 32: Not sure if these vars should be documented, but there is a dedicated section “Internal environment variables” https://dstack.ai/docs/reference/environment-variables/#server

un-def (approved):

src/dstack/_internal/server/settings.py
r4victor: I know. It makes sense to document some internal environment variables for developers but not all of them. These are "technical" atm, maybe we'll document later.

Issue #2133, opened 2024-12-23
[UI] New UI for fleets and instances


jvstme:
### Problem

It is not possible to see a list of all fleets or all instances. It is not possible to manage fleets and instances.

### Solution

Implement three pages with management controls: Fleets list, Fleet details, Instances list.
Details: https://docs.google.com/document/d/1_7Kc8bNUyYRhu29jpuLphUY4T0oF79h5AaKwoKgx-eM

This issue can be done in multiple stages. One way to split into stages:
- [x] #2042
- [x] Add a basic Instances list page - as described in the document, but without instance deletion and links to fleet details
- [x] Update the Fleets list page as described in the document, but without links to fleet details
- [x] Implement instance deletion on the Instances list page
- [x] Add fleet details page and links to it from Fleets list and Instances list

PR #2199, opened 2025-01-20
Add `/api/instances/list`

jvstme:
Add an API method for listing instances with filtering and pagination. This method will be used in the new Instances page in the UI. It is similar to the deprecated `/api/pools/list_instances` method, except it allows filtering by fleet and not by pool.

Also add the `fleet_id` and `fleet_name` fields to the instance model so that the Instances page in the UI can display fleet names and provide links to fleet details.

Part of #2133


src/dstack/_internal/server/services/pools.py
```diff
@@ -542,7 +567,8 @@ async def list_pools_instance_models(
 async def list_user_pool_instances(
     session: AsyncSession,
     user: UserModel,
-    project_name: Optional[str],
+    project_names: Optional[Container[str]],
+    fleet_ids: Optional[Iterable[uuid.UUID]],
```
r4victor, line 571: Not sure if there is a good reason to use abstract types if lists are sufficient. Depending on abstract types puts extra limits on the implementation (e.g. may need to convert to list). I'd not use abstract types unless there is a specific reason.

src/dstack/_internal/server/services/pools.py
```diff
@@ -223,6 +243,8 @@ def instance_model_to_instance(instance_model: InstanceModel) -> Instance:
         id=instance_model.id,
         project_name=instance_model.project.name,
         name=instance_model.name,
+        fleet_id=instance_model.fleet_id,
+        fleet_name=instance_model.fleet.name if instance_model.fleet else None,
```
r4victor, line 247: What if InstanceModel.fleet is not loaded? It seems `instance_model.fleet` would raise sqlalchemy lazy load error when load_instance_fleets = False.

src/dstack/_internal/server/services/pools.py
jvstme: Conversely, depending on specific types puts extra limits on the caller, e.g. the caller may need to convert to list when in fact a set or a generator could work just as well.

It's hard to say which is more likely: changing the implementation in such a way that the abstract types are no longer sufficient, or adding new callers that prefer to use other specific types. I'd bet on the latter.

src/dstack/_internal/server/services/pools.py
jvstme: It's the responsibility of the caller to load `instance_model.fleet` prior to calling `instance_model_to_instance`. Not sure if there is a better solution considering we need the fleet name in `instance_model_to_instance` but can't load fleets efficiently within the function.

Currently, all callers do load the fleet.

src/dstack/_internal/server/services/pools.py
r4victor: > Conversely, depending on specific types puts extra limits on the caller,

Agree, but in this specific case we talk about a service function that's written specifically for a router, so a router will likely be the only client.

Don't you think that following the approach you suggest with defaulting to abstract types we'll end up with lots of inconsistencies?

src/dstack/_internal/server/services/pools.py
r4victor: Ok then. I first got an impression that some callers may not load the attribute.

r4victor (approved):

src/dstack/_internal/server/services/pools.py
jvstme: > we talk about a service function that's written specifically for a router, so a router will likely be the only client

It is also likely that `Container` and `Iterable` will be enough for the purposes of this function. And who knows, we may find other uses for this function as well.

jvstme: > following the approach you suggest with defaulting to abstract types we'll end up with lots of inconsistencies

Do you mean actual type inconsistencies, when the declared type doesn't match the runtime type? This should be prevented by a type checker.

r4victor: I don't mean type correctness. I mean some functions will accept concrete types, some abstract types, and some a mixture of both. This will likely happen if you default to use abstract / generic types from now on.

jvstme: Then I don't think it's an issue. We will just trust the function's signature when calling it - if it expects a list, we pass a list; if expects an iterable, we pass any iterable.

> if you default to use abstract types from now on

I've been using them all the time actually) And I see them in commits from other teammates as well.

r4victor: Ok, I don't insist on changing the types

PR #2200, opened 2025-01-20
Allow getting by ID in `/api/project/_/fleets/get`

jvstme:
Part of #2133

### DIFF
--- a/src/dstack/_internal/server/services/fleets.py
+++ b/src/dstack/_internal/server/services/fleets.py
@@ -180,10 +180,15 @@ async def list_project_fleet_models(
 
 
 async def get_fleet(
-    session: AsyncSession, project: ProjectModel, name: Optional[str], id: Optional[uuid.UUID]
+    session: AsyncSession,
+    project: ProjectModel,
+    name: Optional[str],
+    fleet_id: Optional[uuid.UUID],
 ) -> Optional[Fleet]:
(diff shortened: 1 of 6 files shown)


src/dstack/_internal/server/services/fleets.py
```diff
@@ -179,17 +179,37 @@ async def list_project_fleet_models(
     return list(res.unique().scalars().all())
 
 
-async def get_fleet_by_name(
-    session: AsyncSession, project: ProjectModel, name: str
+async def get_fleet(
+    session: AsyncSession, project: ProjectModel, name: Optional[str], id: Optional[uuid.UUID]
```
r4victor: `id` local variable shadows a Python's built-in. Alternatives: `fleet_id` or `id_`

r4victor:
I suppose it should also handle backward compatibility for new clients with old servers similar to this fix https://github.com/dstackai/dstack/pull/2174

src/dstack/_internal/server/services/fleets.py
jvstme: The `id` built-in is rarely used, and using `id` as a parameter name allows for a more concise function signature. But I think some IDEs and linters don't like it when built-ins are shadowed, so I'll rename it.

r4victor (approved):

PR #2204, opened 2025-01-21
Add termination reason and message to the runner API

### LINKED ISSUE #1701: [Feature]: Use a dedicated status when exceeding `max_duration`
### Problem

In the server console logs, it's unclear that a run was terminated due to max_duration. Attaching image showing when a run was started and the messaging shown 6 hours later (default max_duration). 

image

### Solution

_No response_

### Workaround

_No response_

No

### LINKED ISSUE #2202: Add run termination reason to the runner API
Currently, the runner API only returns the job state (running/failed/terminating/terminated/done). It should also be possible to return the task termination reason / message similar to the shim API. This is needed, for example, for the sever to know when the run was stopped due to max duration (#2163) and possibly to differentiate between different failed reasons.

r4victor:
Closes #2202
Closes #1701

This PR:
* Adds termination_reason and termination_message to the runner API.
* Returns "Max duration exceeded" in termination_message when job is terminated due to max duration.
* Adds max_duration_exceeded termination reason but does not use it yet for backward compatibility with old clients.
* Makes shim/runner return lowercase termination reasons (termination reason values) for consistency with the server API.

The CLI error when max duration exceeded:

```
Installing extension 'ms-toolsai.jupyter'...
Run failed with error code CONTAINER_EXITED_WITH_ERROR.
Error: Max duration exceeded
```


r4victor:
I was also trying to switch to enum for termination reason in shim but it don't think it's feasible. The shim can accept a termination reason from the server and enumerating possible values set by the server will make backward compatibility challenging.

un-def:
Also here: https://github.com/dstackai/dstack/blob/80e9fae29d5b50ebd6b7cd7be3a7c5de7d13ed41/src/dstack/_internal/server/services/jobs/__init__.py#L309

`.name` → `.value` for a lowercase literal

un-def (approved):

PR #2217, opened 2025-01-23
Update migration `ffa99edd1988`

jvstme:
- Rename the migration file to match the revision ID.
- Uncomment the docstring so that alembic displays the revision message.


r4victor (approved):
Let's merge it! Idk how it happened.

PR #2226, opened 2025-01-24
[chore]: Refactor gateway modules

jvstme:
Move service-specific logic out of gateway-specific modules, since gateways are no longer the only way to run services.

The naming of modules may be a bit confusing because the word "service" can refer to both `type: service` runs and the `services` architectural layer. This leads to module names such as `dstack._internal.server.services.services`. The PR attempts to avoid confusion by adding module docstrings.

Part of #1595


r4victor (approved):
Nice!

PR #2236, opened 2025-01-27
[UI] New UI for fleets and instances

### LINKED ISSUE #2133: [UI] New UI for fleets and instances

olgenn:
Close  #2133


jvstme (commented):
@olgenn, thanks for the PR! I tested the new pages and they work as expected except for a few issues. I'll list them below.

Major:
1. The Instances list page fails to load if there are any pending instances. Console error: `React Router caught the following error during render TypeError: item.instance_type is null`.

   https://github.com/user-attachments/assets/02c782ab-0967-478c-8c4b-c174f98ab4c7


1. The pagination seems to be off in the Instances list and Fleets list pages. If I open the second page and then return to the first page, it looks different than when I first opened it. If I then go through the pages again, it shows three pages, although at first there were only two.


   https://github.com/user-attachments/assets/f39e015f-c904-447e-b0bd-030c4cd5315f



Not major, can be finished after merging this PR:
1. On the Fleets list and Fleet details pages, we won't need the `Backend`, `Region`, and `Spot` fields. The `Price` field should be calculated as the sum of non-terminated instance prices. And we also need the `Instances` field with a summary of non-terminated fleet instances. Please see the design document for a detailed specification of the desired fields.
1. We'd also like to show terminated fleets on the Fleets list page. Maybe add an "Active fleets" trigger?
1. The document also requests a "Delete" button on the Fleet details page.
1. The document also requests links to fleet details in the "Fleet" field of the Instances list page.
1. Can we also add links to project settings in the "Project" field of the Instances list page?
1. If the instance costs 0, its price on the Instances list page is shown without a dollar sign. Can we add the dollar sign?
   zero-price

jvstme (approved):
:+1:

PR #2242, opened 2025-01-29
Implement volumes force detach

### LINKED ISSUE #2218: Handle cases when a volume get stuck in the `detaching` state
In certain cases, a volume may get stuck in the `detaching` state. In this case, the run is marked as `stopped`, but the instance remains in the inconsistent state–preventing its deletion or reuse. Also, such a volume cannot be attached to other instances.

While the best approach is to prevent the cases when a volume cannot detach, as a workaround, it's important to handle such cases properly:

1. Keep the run in the `stopping` state until all the volumes are successfully detached.
2. If the user initiates `abort`, apply "force-detach".
3. Allow to configure the `max_stop_duration` to allow for automatic abort if this duration is exceeded. Not set to `off`, do not apply.

r4victor:
Closes #2218

The PR:
* Refactors job termination logic. Before, `process_terminating_job` could be called from `process_terminating_run` (e.g. for non-running provisioned jobs) (evidently to safe a processing iteration). Now, `process_terminating_run` only marks jobs as TERMINATION, and only `process_termination_jobs` background task performs the jobs termination. This simplifies locking management significantly and overall logic.
* Fixes a locking bug when an instance volumes may not be detached on job termination.
* Adds backend method to check volume attach status.
* Adds job terminating logic that keeps checking volume attach status and force detaches volumes if they are stuck.
* Adds `stop_duration` to run configuration/profile to control max duration a job waits until force terminating.
* Fixes `*_duration` parameters parsing. E.g. `"off"` was handled for `max_duration` but not `off` and vice versa for `idle_duration`. `true` was allowed but parsed as 1.
* Other small fixes.

Next:
* Fix `dstack stop` CLI command to wait until the run enters the terminal status. Currently, the API is async meaning that run may be terminating for some time after the API returns.
* Handle attaching/detaching volumes for gpu blocks.


src/dstack/_internal/cli/utils/run.py
```diff
@@ -25,6 +25,8 @@ def print_run_plan(run_plan: RunPlan, offers_limit: int = 3):
     props.add_column(no_wrap=True)  # key
     props.add_column()  # value
 
+    print(job_plan.job_spec.stop_duration)
```
un-def: debug print?

un-def: Nevermind, outdated commit

un-def (approved):

PR #2253, opened 2025-01-31
Implement GPU `blocks` property

un-def:
* Add `blocks` property to cloud and SSH fleet
* Reverse instance->job relation, switch from 1-to-1 to many-to-1 (many jobs to one instance)
* Add InstanceModel.shared_info structure to hold total and currenty used number of blocks
* Generate virtual shared offers (fraction of a real offer) on the fly
* Update CLI and API to display shared resources and instances
* Keep track of volumes used by jobs


src/dstack/_internal/core/models/fleets.py
```diff
@@ -61,6 +61,17 @@ class SSHHostParams(CoreModel):
     ] = None
     ssh_key: Optional[SSHKey] = None
 
+    blocks: Annotated[
+        Union[Literal["auto"], int],
+        Field(
+            description=(
+                "The amount of blocks to split the instance into, a number or `auto` (e.g., `4`)."
```
r4victor: Maybe elaborate what `auto` means?

src/dstack/_internal/core/models/instances.py
```diff
@@ -110,6 +110,11 @@ def get_public_keys(self) -> List[str]:
         return [ssh_key.public.strip() for ssh_key in self.ssh_keys]
 
 
+class InstanceSharedInfo(CoreModel):
+    total_blocks: Union[Literal["auto"], int]
+    busy_blocks: int = 0
```
r4victor: Not sure about storing `busy_blocks` in json vs in a separate column. Any particular reasons to do so? I'm generally inclined to keep static data in json and changing fields like this as separate columns.

Maybe using json is more flexible if we take a complete different approach instead of relying on `blocks`, Idk.

src/dstack/_internal/server/services/offers.py
```diff
@@ -92,4 +100,48 @@ async def get_offers_by_requirements(
...
+        else:
+            divisor = blocks
+        if len(gpus) % divisor or resources.cpus % divisor:
+            # gpus or cpus are not divisible by blocks
+            continue
+        shareable_offers.append((backend, offer))
+    return shareable_offers
+
+
+def generate_shared_offer(offer: InstanceOfferWithAvailability, blocks: int, total_blocks: int):
```
r4victor: missing return type

src/dstack/_internal/core/models/runs.py
```diff
@@ -416,7 +421,7 @@ def _error(cls, values) -> Dict:
 
 class JobPlan(CoreModel):
     job_spec: JobSpec
-    offers: List[InstanceOfferWithAvailability]
+    offers: List[Union[InstanceSharedOffer, InstanceOfferWithAvailability]]
```
r4victor: Do we really need to introduce different types of offers to the API? Can't every InstanceOfferWithAvailability be InstanceSharedOffer with total_blocks/blocks = 1?

src/dstack/_internal/server/services/fleets.py
```diff
@@ -395,6 +398,9 @@ async def create_fleet_instance_model(
 ) -> InstanceModel:
     profile = spec.merged_profile
     requirements = _get_fleet_requirements(spec)
+    blocks: Optional[Union[Literal["auto"], int]] = spec.configuration.blocks
+    if isinstance(blocks, int) and blocks < 2:
+        blocks = None
```
r4victor: Setting shared_info to None when `blocks: 1` is very confusing since you need to know that total_blocks cannot be 1 (e.g. when shared offers are generated). I'd reverse it: if blocks is None, then assume `blocks: 1`.

r4victor: Again, with my solution, it would be more appropriate to normalize shared_info to keep total/used blocks in separate columns so that it's easier to migrate.

src/dstack/_internal/core/models/runs.py
```diff
@@ -239,6 +240,10 @@ class JobRuntimeData(CoreModel):
     # None if data is not yet available (on vm-based backends and ssh instances)
     # or not applicable (container-based backends)
     ports: Optional[dict[int, int]] = None
+    # List of volumes used by the job
+    volume_names: Optional[list[str]] = None  # None for backward compalibility
+    # Virtual shared offer. None if the instance is not shared.
+    offer: Optional[InstanceSharedOffer] = None
```
r4victor: Maybe it's more useful to always store the offer used by the job, not just when it was shared?

r4victor: I think it's crucial now to add description of JobRuntimeData and what should go there so that it does not become a catch-all struct for everything. Probably makes sense to add a similar description to JobProvisioningData to differentiate it with JobRuntimeData.

src/dstack/_internal/server/background/tasks/process_instances.py
```diff
@@ -557,6 +589,18 @@ async def _create_instance(session: AsyncSession, instance: InstanceModel) -> No
...
+            if blocks == "auto":
+                blocks = len(instance_offer.instance.resources.gpus)
+                if blocks > 1:
+                    shared_info.total_blocks = blocks
+                else:
+                    shared_info = None
+            if shared_info is not None:
+                instance.shared_info = shared_info.json()
+            else:
+                instance.shared_info = None
```
r4victor: The `shared_info` updating is convoluted. I believe it can be created with `total_blocks: "auto"` and then it's overridden. I don't see much value in that. I'd keep user's configured `blocks` as a static property somewhere and wound't touch it. `total_blocks` can then be resolved when instance resources are known. No need to allow for `auto`.

src/dstack/_internal/server/services/offers.py
```diff
@@ -92,4 +100,48 @@ async def get_offers_by_requirements(
...
+            # TPU instances cannot be shared
+            continue
+        if blocks == "auto":
+            divisor = len(gpus)
+        else:
+            divisor = blocks
+        if len(gpus) % divisor or resources.cpus % divisor:
+            # gpus or cpus are not divisible by blocks
+            continue
+        shareable_offers.append((backend, offer))
```
r4victor: I think this code can be nicely refactored into a separate func.

src/dstack/_internal/server/services/pools.py
```diff
@@ -494,6 +511,53 @@ def filter_pool_instances(
...
+            shared_offer = generate_shared_offer(offer, blocks, total_blocks)
+            catalog_item = offer_to_catalog_item(shared_offer)
+            if gpuhunt.matches(catalog_item, query_filter):
+                if blocks <= idle_blocks:
+                    shared_offer.availability = InstanceAvailability.IDLE
+                else:
+                    shared_offer.availability = InstanceAvailability.BUSY
+                if shared_offer.availability == InstanceAvailability.IDLE or not idle_only:
+                    instances_with_offers.append((instance, shared_offer))
+                break
```
r4victor, line 543: So we show only the smallest matching offer? Is it intended. Asking because the doc showed all possible offers.

r4victor (approved):
Tested with cloud fleets. Everything works as expected.

I'd reconsider how `shared_info` is stored/updated (mainly to make the code more comprehendible) and if InstanceSharedOffer/InstanceOfferWithAvailability should be differentiated in the API. Otherwise, looks good.

src/dstack/_internal/server/migrations/versions/1338b788b612_reverse_job_instance_relationship.py
```diff
@@ -0,0 +1,69 @@
...
+    """)
+
+    with op.batch_alter_table("instances", schema=None) as batch_op:
+        batch_op.drop_constraint("fk_instances_job_id_jobs", type_="foreignkey")
+        batch_op.drop_column("job_id")
+
+
+def downgrade() -> None:
+    with op.batch_alter_table("instances", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("job_id", sa.UUID(), autoincrement=False, nullable=True))
```
r4victor: sa.UUID() fails with sqlite

change to:

```
batch_op.add_column(sa.Column("job_id", sqlalchemy_utils.types.uuid.UUIDType(binary=False), nullable=True))
```

src/dstack/_internal/core/models/runs.py
un-def: https://github.com/dstackai/dstack/pull/2253/commits/3342a34f1c371d3ff58ac5e41118e56c28a63059

src/dstack/_internal/core/models/fleets.py
un-def: https://github.com/dstackai/dstack/pull/2253/commits/5d5434ba82c86bec7c8865bc7d04e55b2d9cfb39

src/dstack/_internal/core/models/instances.py
un-def: Replaced with two integer columns https://github.com/dstackai/dstack/pull/2253/commits/8b555f7a6ad65afbed244e3d1e76359bdbbcbbf9

src/dstack/_internal/server/services/offers.py
un-def: Fixed

src/dstack/_internal/core/models/runs.py
un-def: https://github.com/dstackai/dstack/pull/2253/commits/a6dcde20a4a673cb0b458a24389cfecc68a058c7

src/dstack/_internal/server/services/pools.py
un-def: Yes, IMO there is no reason to show all virual offers as they won't be picked anyway

src/dstack/_internal/server/migrations/versions/1338b788b612_reverse_job_instance_relationship.py
un-def: https://github.com/dstackai/dstack/pull/2253/commits/a6dcde20a4a673cb0b458a24389cfecc68a058c7

src/dstack/_internal/server/services/fleets.py
un-def: Refactored as part of https://github.com/dstackai/dstack/pull/2253/commits/8b555f7a6ad65afbed244e3d1e76359bdbbcbbf9

src/dstack/_internal/core/models/runs.py
un-def: Changed in https://github.com/dstackai/dstack/pull/2253/commits/8b555f7a6ad65afbed244e3d1e76359bdbbcbbf9

src/dstack/_internal/server/background/tasks/process_instances.py
un-def: Replaced with NULL -> calculated auto blocks in https://github.com/dstackai/dstack/pull/2253/commits/8b555f7a6ad65afbed244e3d1e76359bdbbcbbf9

NULL as "info is not yet available" seems to fit there

src/dstack/_internal/server/services/offers.py
un-def: Refactored in https://github.com/dstackai/dstack/pull/2253/commits/8b555f7a6ad65afbed244e3d1e76359bdbbcbbf9

r4victor (approved):

PR #2266, opened 2025-02-05
Allow configuring availability zones 

### LINKED ISSUE #2107: [Feature]: Availability zones in configurations
### Problem

Currently, it's not possible to create a resource in a specific AZ. Given that some backends require the instances and volumes to belong not only to the same region, but also the same AZ, it can be a problem. For example, with AWS backend I cannot reliably create such a volume that would work with the existing fleet instance.

### Solution

* Add `availability_zones: array[string]` property to the [fleet](https://dstack.ai/docs/reference/dstack.yml/fleet/#root-reference) configuration
* Add `availability_zone: string` property to the [volume](https://dstack.ai/docs/reference/dstack.yml/volume/#root-reference) configuration
* Display AZ in `dstack fleet` (or `dstack fleet -v`) output
* Display AZ in `dstack volume` (or `dstack volume -v`) output

### Workaround

_No response_

No

r4victor:
Closes #2107

The PR:
* Adds `availability_zones` parameter to profiles, run, and fleet configurations.
* Adds `availability_zone` parameter to volume configurations.
* Supports custom availability zones for AWS, GCP, OCI.
* Shows availability zones in `dstack fleet -v` and `dstack volumes -v`.
* Fixes missing backward compatibility handling (run_spec excludes) for /runs/apply – led to broken in-place updates.


src/dstack/_internal/cli/utils/fleet.py
```diff
@@ -72,6 +78,10 @@ def get_fleets_table(
             row += [
                 str(instance.instance_num),
                 backend,
+            ]
+            if verbose:
+                row.append(region)
```
jvstme: Error trying to view the verbose table with pending or failed instances:
```
UnboundLocalError: local variable 'region' referenced before assignment
```

jvstme: Btw, I can suggest using `add_row_from_dict` for tables with conditional columns, so that there are less `if`s and such errors are easier to detect.

Same in `get_volumes_table`.

src/dstack/_internal/core/backends/aws/compute.py
```diff
@@ -161,10 +161,12 @@ def create_instance(
...
-        availability_zones = None
-        if instance_config.availability_zone is not None:
-            availability_zones = [instance_config.availability_zone]
-
+        zones = aws_resources.get_availability_zones(ec2_client, instance_offer.region)
+        instance_config_zones = instance_config.get_availability_zones()
+        if instance_config_zones is not None:
+            zones = [z for z in zones if z in instance_config_zones]
+            if len(zones) == 0:
+                raise ComputeError(f"No valid availability zones: {instance_config_zones}")
```
jvstme: I think this filtering should happen for all backends in`get_offers_by_requirements`, not in `create_instance` of `BackendCompute` implementations. I.e. `availability_zones` (or one `availability_zone`) could be a property of `InstanceOfferWithAvailabiity`, not of `InstanceConfiguration`.

Otherwise offers aren't filtered, which seems unexpected.

### Case 1

```yaml
type: fleet
regions: ["eu-north-1"]
availability_zones: ["us-west4-a"]
nodes: 1
spot_policy: auto
```

#### Expected

No offers because of conflicting `regions` and `availability_zones`.

#### Actual

Shows offers from eu-north-1, starts provisioning, logs 80+ warnings in server logs before finally failing.

### Case 2

```yaml
type: fleet
availability_zones: ["us-west4-a"]
nodes: 1
resources:
  gpu: 1
```

#### Expected

Only shows offers from AWS us-west4-a.

#### Actual

Shows offers from all backends and regions, provisions in Vultr.

src/dstack/_internal/core/backends/aws/resources.py
```diff
@@ -370,7 +370,9 @@ def get_subnets_ids_for_vpc(
     return subnets_ids
 
 
-def get_availability_zone(ec2_client: botocore.client.BaseClient, region: str) -> Optional[str]:
+def get_any_availability_zone(
```
jvstme: Unused. Delete?

src/dstack/_internal/core/backends/gcp/resources.py
```diff
@@ -31,7 +31,7 @@
 ]
 
 
-def get_availability_zone(
+def get_any_availability_zone(
```
jvstme: Unused. Delete?

src/dstack/_internal/core/backends/aws/compute.py
r4victor: It doesn't make sense to make an offer per availability_zone since offers are not tied to availability_zone. But I agree it would be useful to include availability_zones in the offer so that we can do offers filtering before submission. I think this would require changing the gpuhunt API to statically include availability_zones.

I'd ignore the cases above for now since they are edge cases. Case 2 can be covered by requiring regions when availability_zones are specified.

src/dstack/_internal/cli/utils/fleet.py
r4victor: thanks!

src/dstack/_internal/core/backends/aws/resources.py
r4victor: seemed potentially useful to leave just in case

src/dstack/_internal/core/backends/aws/compute.py
jvstme: > offers are not tied to availability_zone

But the offer's availability is tied to availability zones. So `InstanceOfferWithAvailability.availability_zones` could hold a list of zones in which this offer is available. This property can be populated in `BackendCompute.get_offers`.
- In OCI, we already check availability in `get_offers`, so `availability_zones` is easy to populate.
- In GCP, the zone is already part of `gpuhunt.CatalogItem`, so `availability_zones` is easy to populate too.
- In AWS, we don't know about zones in `get_offers`, so populating `availability_zones` will require an extra `describe_availability_zones` API request.

src/dstack/_internal/core/backends/aws/resources.py
jvstme: Well, this is a very trivial function, I'm sure we can easily re-create it if we'll ever need it

src/dstack/_internal/core/backends/oci/compute.py
```diff
@@ -197,3 +200,22 @@ def _supported_instances(offer: InstanceOffer) -> bool:
     if "Flex" in offer.instance.name:
         return False
     return any(map(offer.instance.name.startswith, SUPPORTED_SHAPE_FAMILIES))
+
+
+def _get_availability_domain_or_error(
+    availabile_domains: List[str],
```
jvstme: (nit) Same typo in a few places
```suggestion
    available_domains: List[str],
```

r4victor:
@jvstme, added InstanceOfferWithAvailability.availability_zones and updated get_offers for AWS, GCP, OCI as you suggested. OCI get_offers went for me from ~14s to ~20s. Other backends' times are not affected.

src/dstack/_internal/core/backends/oci/resources.py
```diff
@@ -171,7 +171,7 @@ def check_availability_in_domain(
     availability_domain_name: str,
     client: oci.core.ComputeClient,
     compartment_id: str,
-) -> Set[str]:
+) -> List[str]:
```
jvstme: This function actually still returns a set. Same in `check_availability_per_domain`.

The code works as expected with sets though.

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -373,30 +375,31 @@ def run_job(
...
                     ]
-                instance_config.availability_zones = [
-                    az
-                    for az in instance_config.availability_zones
-                    if az == volume.provisioning_data.availability_zone
+                instance_offer.availability_zones = [
+                    z
+                    for z in instance_offer.availability_zones
+                    if z == volume.provisioning_data.availability_zone
                 ]
```
jvstme: (nit) Looks like this filtering is no longer backend-specific and can now be moved to `get_offers_by_requirements`?

Same in AWS.

src/dstack/_internal/core/backends/gcp/compute.py
```diff
@@ -371,22 +375,32 @@ def run_job(
...
     ) -> JobProvisioningData:
+        # TODO: run_job is the same for vm-based backends, refactor
         instance_config = InstanceConfiguration(
             project_name=run.project_name,
             instance_name=get_instance_name(run, job),  # TODO: generate name
             ssh_keys=[
                 SSHKey(public=project_ssh_public_key.strip()),
             ],
             user=run.user,
-            volumes=volumes,
```
jvstme, line 381: Is this change intentional? It looks like `InstanceConfiguration.volumes` is no longer set anywhere

src/dstack/_internal/core/backends/oci/compute.py
```diff
@@ -111,11 +117,13 @@ def create_instance(
...
-        availability_domain = resources.choose_available_domain(
+        available_domains = resources.get_available_domains(
             instance_offer.instance.name, self.shapes_quota, region, self.config.compartment_id
         )
-        if availability_domain is None:
-            raise NoCapacityError("Shape unavailable in all availability domains")
+        availability_domain = _get_availability_domain_or_error(
+            available_domains=available_domains,
+            offer_domains=instance_offer.availability_zones,
+        )
```
jvstme: This now seems redundant and uses 1-3 extra API calls. Just use one of the `instance_offer.availability_zones`.

```python
if not instance_offer.availability_zones:
    raise NoCapacityError("Shape unavailable in all availability domains")
availability_domain = instance_offer.availability_zones[0]
```

`resources.get_available_domains` and `_get_availability_domain_or_error` can then be deleted.

jvstme (approved):

src/dstack/_internal/core/backends/gcp/compute.py
r4victor: Copied AWS implementation that missed volumes=volumes. It's currently used for TPUs only.

src/dstack/_internal/core/backends/oci/compute.py
r4victor: I was assuming the code may rely on the availability check and instance_offer.availability_zones may not be relevant, so left another check just in case. Happy to remove it.

src/dstack/_internal/core/backends/gcp/compute.py
r4victor: It's actually backend-specific since only backends can now if volumes are regional or zonal. It just happens that regional volumes are not supported and all use the same implementation. It'd leave it on the backend level for now and make a separate PR to refactor run_job for VM-based backends to avoid duplication.

PR #2271, opened 2025-02-06
Full EFA attachment for non-public IPs

### LINKED ISSUE #2270: [Feature]: AWS: Full EFA support for non-public IPs
### Problem

A small update for EFA attachment, which attaches the max possible number of EFA interfaces. This only covers only non-public IP nodes due to limitation on interface attachment if public IP is used.

### Solution

_No response_

### Workaround

_No response_

Yes

timsolovev:
Closes #2270

### DIFF
--- a/src/dstack/_internal/core/backends/aws/resources.py
+++ b/src/dstack/_internal/core/backends/aws/resources.py
@@ -140,6 +140,7 @@ def create_instances_struct(
     allocate_public_ip: bool = True,
     placement_group_name: Optional[str] = None,
     enable_efa: bool = False,
+    max_efa_interfaces: int = 0,
     reservation_id: Optional[str] = None,
     is_capacity_block: bool = False,
 ) -> Dict[str, Any]:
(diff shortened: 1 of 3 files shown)


src/dstack/_internal/core/backends/aws/resources.py
```diff
@@ -199,9 +200,44 @@ def create_instances_struct(
...
+                for i in range(1, max_efa_interfaces):
+                    struct["NetworkInterfaces"].append(
+                        {
+                            "AssociatePublicIpAddress": allocate_public_ip,
+                            "NetworkCardIndex": i,
+                            "DeviceIndex": 1,
+                            "SubnetId": subnet_id,
+                            "Groups": [security_group_id],
+                            "InterfaceType": "efa-only",  # Set specifically to efa-only to keep the interfaces exclusively for GPU-to-GPU communications and not mix with IP traffic
+                        }
```
r4victor: It seems "InterfaceType" is the only difference between "p5.48xlarge" and other instance types? Then the code can be simplified significantly:

```python
            for i in range(1, max_efa_interfaces):
                # Set efa-only to keep the interfaces exclusively for GPU-to-GPU communications and not mix with IP traffic
                interface_type = "efa-only"
                if instance_type == "p5.48xlarge":
                    # EFA configuration for P5 instances:
                    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html#efa-for-p5
                    interface_type = "efa" if i % 4 == 0 else "efa-only"
                struct["NetworkInterfaces"].append(
                    {
                        "AssociatePublicIpAddress": allocate_public_ip,
                        "NetworkCardIndex": i,
                        "DeviceIndex": 1,
                        "SubnetId": subnet_id,
                        "Groups": [security_group_id],
                        "InterfaceType": interface_type,
                    }
                )
```

timsolovev: Sorry, pushed the choppy version without refactoring. Fixed.

r4victor:
@solovyevt, thanks for the PR, looks good. I'm merging it.

I also pushed a commit to update the docs.

Issue #2284, opened 2025-02-10
[Feature]: Dynamic CLI autocompletion


timsolovev:
### Problem

An initial attempt at CLI autocompletion with dynamic resource retrieval. Currently implemented only for `dstack stats` to validate the approach. Other things that need to be taken care of:

- [x] A separate command `dstack completion` to generate bash/zsh completions
- [x] Autocomplete common prefix for matching options (e.g. `tes` → Tab → `test-` if options are `test-0` and `test-1`)
- [x] Update documentation for autocompletion, something along the lines of `dstack completion > /etc/bash_completion.d/dstack` or `echo 'dstack completion' >> ~/.bashrc`, same for `.zshrc`
- [x] Backward compatibility
- [ ] (If GETs are too intensive) Server- or client-side caching, better be implemented in a separate PR
- [ ] Server-side prefix processing

Would appreciate early feedback, would prefer to get the plumbing right before proceeding to other commands.

If you want to try it locally - use `eval "$(dstack completion bash)"`.

### Would you like to help us implement this feature by sending a PR?

Yes

r4victor:
@solovyevt, sounds good. I think we're all in favor of adding autocompletion to the dstack CLI. Happy to assist on this.

PR #2285, opened 2025-02-10
Dynamic CLI completion

timsolovev:
WIP #2284


src/dstack/_internal/cli/utils/completion.py
```diff
@@ -0,0 +1,63 @@
...
+    from the parsed arguments, otherwise falls back to the DSTACK_PROJECT environment variable).
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def get_api(self, parsed_args):
+        # TODO: Feedback needed: Refactor this to avoid duplication with APIBaseCommand._register().
+        # The client has to be explicitly initialized here since argcomplete does not trigger APIBaseCommand._command(),
+        # and APIBaseCommand().api is not being set. This is a temporary hack.
```
r4victor: I don't see a better way to init the api client since we need parsed_args to support `--project`, and it's only passed in `__call__`. I think the current solution is ok. `Client.from_config()` is cheap (does not involve network calls), so it should be fine to call it on every `__call__`.

src/dstack/_internal/cli/utils/completion.py
```diff
@@ -0,0 +1,63 @@
...
+        debug("Retrieving API client")
```
r4victor: Maybe replace `debug` with `argcomplete.debug`? Just my preference to keep generic names in namespaces.

src/dstack/_internal/cli/utils/completion.py
r4victor, line 1: It's best to move utils/completion.py to services/completion.py since utils are not supposed to import from other `dstack` modules and completion.py may need to import (e.g. `from dstack.api import Client`)

timsolovev:
I've quickly ran a profiler, and - as expected from a naive  argcomplete integration  - most of the delays for completions are caused by cascading imports, not by the server requests. This could be handled either by reworking the imports or by precalculating and caching of the completions, kind of like it's done in Heroku CLI: https://github.com/heroku/cli/tree/main/packages/cli/src/lib/autocomplete

src/dstack/_internal/cli/utils/completion.py
```diff
@@ -0,0 +1,63 @@
...
+    def __call__(self, prefix, parsed_args, **kwargs):
+        api = self.get_api(parsed_args)
+        debug("Fetching run completions")
+        try:
+            runs = api.runs.list()
+            completions = [run.name for run in runs if run.name.startswith(prefix)]
+            return completions
+        except Exception as e:
+            debug("Error fetching run completions: " + str(e))
+            return []
```
r4victor: If I try to `dstack stats <TAB HIT>` when the server is unavailable/shutdown, the shell gets stuck for some time. In debug mode I see
```
Error fetching run completions: Failed to connect to dstack server http://127.0.0.1:8000
```
I think the errors should be printed with `argcomplete.warn` so that users can see that autocomplete does not work and why.

Autocomplete appears to get stuck because for some reason it gets the exception three times:

```
Error fetching run completions: Failed to connect to dstack server http://127.0.0.1:8000

Error fetching run completions: Failed to connect to dstack server http://127.0.0.1:8000

Error fetching run completions: Failed to connect to dstack server http://127.0.0.1:8000
```

We do retry server requests three times, but it should not cause the exception being propagated three times. It seems something calls RunNameCompleter three times.

src/dstack/_internal/cli/utils/completion.py
```diff
@@ -0,0 +1,63 @@
...
+    def get_api(self, parsed_args):
+        # TODO: Feedback needed: Refactor this to avoid duplication with APIBaseCommand._register().
+        # The client has to be explicitly initialized here since argcomplete does not trigger APIBaseCommand._command(),
+        # and APIBaseCommand().api is not being set. This is a temporary hack.
+        debug("Retrieving API client")
+        project = getattr(parsed_args, "project", os.getenv("DSTACK_PROJECT"))
+        return Client.from_config(project_name=project)
+
+
+class RunNameCompleter(BaseProjectCompleter):
```
r4victor: This RunNameCompleter can be parametrized to suggest all runs or only active runs (see api.runs.list(all=False)), e.g. for commands like `dstack stop`/`dstack stats`, it makes sense to pass only active runs, for other commands like `dstack logs`, it should return all runs.

src/dstack/_internal/cli/main.py
```diff
@@ -72,9 +75,12 @@ def main():
     StatsCommand.register(subparsers)
     StopCommand.register(subparsers)
     VolumeCommand.register(subparsers)
+    CompletionCommand.register(subparsers)
+
+    argcomplete.autocomplete(parser, always_complete_options=False)
 
     args, unknown_args = parser.parse_known_args()
-    args.unknown = unknown_args
```
r4victor, line 77: `dstack apply` uses `args.unknown`

setup.py
```diff
@@ -49,6 +49,7 @@ def get_long_description():
     "filelock",
     "psutil",
     "gpuhunt>=0.0.19,<0.1.0",
+    "argcomplete==3.5.3",  # TODO: Version range
```
r4victor: Min supported version like `argcomplete>=3.5.0` would be sufficient.

src/dstack/_internal/cli/main.py
timsolovev: Just noticed and restored.

src/dstack/_internal/cli/utils/completion.py
timsolovev: Fixed

src/dstack/_internal/cli/utils/completion.py
timsolovev: Fixed

r4victor:
> I've quickly ran a profiler, and - as expected from a naive argcomplete integration - most of the delays for completions are caused by cascading imports, not by the server requests. This could be handled either by reworking the imports or by precalculating and caching of the completions, kind of like it's done in Heroku CLI: https://github.com/heroku/cli/tree/main/packages/cli/src/lib/autocomplete

I'd guess cache completion is mainly to get rid of server requests. I can see cache invalidation being tricky in this approach. Maybe we can optimize imports. Can you share the profiler results?

solovyevt:
@r4victor DMd in Discord, cannot attach .pstat files here.

r4victor:
For me, `dstack st` doesn't feel slow. `dstack stats` feels ok with localhost but sluggish with a remote server. Optimizing cascading imports would help all cases but I don't see how we can do it. Maybe caching is more straightforward after all. Anyway, I think we can first release CLI completion without fancy optimizations.

src/dstack/_internal/cli/utils/completion.py
timsolovev: Good suggestion, added that

src/dstack/_internal/cli/utils/completion.py
timsolovev: Interesting, that seems to be caused by argcomplete iterating through different parsers if a completer returns an empty list of completions.

Another thing I've noticed - connection timeouts are not handled well due to the missing timeouts in the client. I think a timeout of 5 to 10 seconds would be a reasonable choice. https://github.com/solovyevt/dstack/blob/d03ac2557d45af7c725ae525f87024a45013db88/src/dstack/api/server/__init__.py#L127

r4victor:
@solovyevt, what are you plans on this PR's scope / completion MVP? Do you want to implement name completion for all commands? Any help needed?

timsolovev:
@r4victor Basic completion for all commands and some fields, like `--project`. I think it currently covers the cases I had in mind, and I would appreciate your help with the backward compatibility.

Besides that, I will try to resolve this issue with repeated completion calls, it might be tricky.

r4victor:
Seems like backward compatibility should not be impacted. If you don't change the server API, the new CLI version can get working completions with older servers.

src/dstack/_internal/cli/utils/completion.py
timsolovev: Could not find a way to circumvent this with native mechanisms, so as a quick fix a completer returns a single empty string if API connection fails - no repeated calls now. The only issue is that now it automatically concatenates an additional space to an empty string, couldn't find a way to disable `append_space` exclusively for custom parsers.

timsolovev:
@r4victor I think it is basically complete, only missing the documentation. Please let me know if I missed something.

UPD: discovered an issue with the project completion, the logic doesn't make sense as the user likely expects the completions of the projects defined in `.dstack/config.yml`, not on the server. There is no suitable function in ConfigManager to list the projects.

src/dstack/_internal/cli/utils/completion.py
r4victor: Are you sure the repeated calls are caused by argcomplete? I couldn't find any logic related to this. Maybe it's shell that does the repeated calls (I'm on zsh)? There are still repeated calls if the server is available but returns empty results. Or if the server is not available and you continue `dstack stats a`. (3 retries in all cases.) Returning an empty string seems to help only with `dstack stats` for some reason.

timsolovev: @r4victor You are right, the empty string mitigates repeated calls only for `dstack stats `.

Upon further inspection, when I tried to reproduce this issue both in a fresh MacOS VM and in a container with a plain `zsh` and `autoload -U +X compinit && compinit && eval "$(dstack completion zsh)"`, this issue does not occur at all. Same with `bash`. 

My current guess is that it might be related to zsh configuration frameworks s.a. `oh-my-zsh` and their fiddling with completions. 

UPD: Confirmed, `oh-my-zsh` in the same fresh environments causes repeated completions.

r4victor: Thanks for the investigation! This is no blocker then, although we should figure out what to do with oh-my-zsh eventually.

r4victor:
I'll test it again early next week and merge it. @solovyevt would you be willing to update the docs with installation instructions in this or separate PR? I think it can be an expandable section in https://github.com/dstackai/dstack/blob/c4ddee1896fa93fbaef57df235f67c8ed14afc4c/docs/docs/installation/index.md#L63

src/dstack/_internal/cli/utils/completion.py
timsolovev: One more detail for future reference - the repeated calls seem to be triggered by a non-zero return code from [`_describe`](https://github.com/kislyuk/argcomplete/blob/f3443c002a15b3e9c0c56b0ca5f09ae027b0daac/argcomplete/shell_integration.py#L53) on empty completions list, which - as far as I can guess - triggers some sort of fallback completion mechanism introduced by `oh-my-zsh`.

r4victor: I filed an issue in argcomplete: https://github.com/kislyuk/argcomplete/issues/523. Seems like it can be fixed on their side.

r4victor:
@solovyevt, fixed typing with minor refactoring, LGTM. Please confirm if it's ok to merge on your side.

timsolovev:
@r4victor Added completion docs, check two previous commits. I've started with a more detailed version, then reduced it to match the [Installation](https://dstack.ai/docs/installation/) page. 

Nothing else on my behalf, feel free to merge if lgty.

docs/docs/installation/index.md
```diff
@@ -94,6 +94,37 @@ Configuration is updated at ~/.dstack/config.yml
...
+
+`dstack` supports shell autocompletion for `bash` and `zsh`.
+
+=== "bash"
+    
+    <div class="termy">
+    
+    ```shell
+    $ eval "$(dstack completion bash)"
+    $ echo "$(dstack completion bash)" >> ~/.bashrc
```
r4victor: I'm not sure it's a good idea to write directly to ~/.bashrc or ~/.zshrc. Let me fix it.

r4victor:
I reverted to a more detailed version, recommended not writing to shell profile directly, and added commands for oh my zsh. Merging it.

r4victor:
@solovyevt, thanks for the contribution. We really appreciate it. Feel free to open more PRs from new branches if it needs more changes.

Issue #2277, opened 2025-02-10
[Bug]: Provision Failing with Runpod


Bihan:
### Steps to reproduce

Run the example configuration with runpod backend.

`dstack apply -f ./examples/.dstack.yml -b runpod`

### Actual behaviour

dstack apply -f ./examples/.dstack.yml -b runpod

 ```
Project          bihan                                        
 User             admin                                        
 Configuration    examples/.dstack.yml                         
 Type             dev-environment                              
 Resources        2..xCPU, 8GB.., 1xGPU (24GB), 100GB.. (disk) 
 Max price        -                                            
 Max duration     6h                                           
 Spot policy      auto                                         
 Retry policy     -                                            
 Creation policy  reuse-or-create                              
 Idle duration    5m                                           
 Reservation      -                                            

```
 ```
#  BACKEND  REGION    INSTANCE          RESOURCES                                    SPOT  PRICE   
 1  runpod   EUR-IS-2  NVIDIA L4         12xCPU, 201GB, 1xL4 (24GB), 100.0GB (disk)   yes   $0.22   
 2  runpod   CA-MTL-1  NVIDIA RTX A5000  9xCPU, 48GB, 1xA5000 (24GB), 100.0GB (disk)  yes   $0.22   
 3  runpod   EU-SE-1   NVIDIA RTX A5000  9xCPU, 50GB, 1xA5000 (24GB), 100.0GB (disk)  yes   $0.22   
    ...                                                                                             
 Shown 3 of 18 offers, $0.69 max

Finished run vscode already exists.
Override the run? [y/n]: y
 NAME    BACKEND            INSTANCE          RESOURCES                    RESERVATION  PRICE  STATUS  SUBMITTED  ERROR                           
 vscode  runpod (CA-MTL-1)  NVIDIA RTX A5000  9xCPU, 48GB, 1xA5000                      $0.22  failed  06:36      JOB_FAILED                      
                                              (24GB), 100.0GB (disk),                                             (WAITING_RUNNER_LIMIT_EXCEEDED) 
                                              SPOT                                                                                                

vscode provisioning completed (failed)
Run failed with error code WAITING_RUNNER_LIMIT_EXCEEDED.
Check CLI, server, and run logs for more details.
```

### Expected behaviour

Provision should happen successfully.

### dstack version

Latest Master

### Server logs

```shell
[...]

PR #2286, opened 2025-02-10
Support non-root Docker images in RunPod

jvstme:
RunPod recently added an option to override the
container user. This allows us to run
dstack-runner as root regardless of the Docker
image, thus supporting Docker images with a
non-root default user, such as NVIDIA NIM.

Related to #1535
Same as #2185


Bihan:
@jvstme With this branch, when I tested provisioning container with runpod using our default example `dstack apply -f ./examples/.dstack.yml -b runpod` the container is not starting. The issue is similar to https://github.com/dstackai/dstack/issues/2277


But when I add the change `env={"RUNPOD_POD_USER": "0"},` in latest master and test there is no issue.

jvstme:
@Bihan, this is probably because master has a newer dstack-runner version that this branch is not compatible with. I merged master into this branch, this should make it work. Would you like to retest?

Bihan:
> @Bihan, this is probably because master has a newer dstack-runner version that this branch is not compatible with. I merged master into this branch, this should make it work. Would you like to retest?

Yes, I will do a quick test and inform you.

Bihan:
@jvstme when I tried to run `NIM` example with runpod as below

`dstack apply -f ./examples/deployment/nim/.dstack.yml -b runpod` 

I am able to get offers only when I remove the `volumes` from the YAML file and run below config

```
type: service
name: llama31

image: nvcr.io/nim/meta/llama-3.1-8b-instruct:latest
env:
  - NGC_API_KEY
  - NIM_MAX_MODEL_LEN=4096
registry_auth:
  username: $oauthtoken
  password: ${{ env.NGC_API_KEY }}
port: 8000
# Register the model
model: meta/llama-3.1-8b-instruct

# Uncomment to leverage spot instances
#spot_policy: auto

# Cache downloaded models
# volumes:
#   - /root/.cache/nim:/opt/nim/.cache

resources:
  gpu: 24GB
  # Uncomment if using multiple GPUs
  #shm_size: 24GB
```

jvstme:
@Bihan, this is expected because instance volumes are not supported in RunPod.
https://dstack.ai/docs/guides/troubleshooting/#cause-7-feature-support

I think we could try to improve our examples so that they work in all backends without configuration changes, e.g. by using [optional instance volumes](https://github.com/dstackai/dstack/releases/tag/0.18.40). I'd do this in a separate PR though, since this PR only fixes the issue of running non-root images in RunPod.

Bihan:
> @Bihan, this is expected because instance volumes are not supported in RunPod. https://dstack.ai/docs/guides/troubleshooting/#cause-7-feature-support
> 
> I think we could try to improve our examples so that they work in all backends without configuration changes, e.g. by using [optional instance volumes](https://github.com/dstackai/dstack/releases/tag/0.18.40). I'd do this in a separate PR though, since this PR only fixes the issue of running non-root images in RunPod.

@jvstme That sounds reasonable. 

BTW the NIM container ran successfully with RunPod. However, during deployment of model `llama-3.1-8b`, Runpod crashed with OOM. This must be RunPod's issue because GPU vRAM was not fully consumed but CPU RAM went 100% and OOM occurred. I will re-confirm and share issue with RunPod team. 

I think we are good to go with the merge.

Bihan (approved):

jvstme:
> GPU vRAM was not fully consumed but CPU RAM went 100% and OOM occurred

I've seen problems like this before and was able to solve them by setting a higher `memory` in `resources`. Not sure if this is expected or not.

Issue #2010, opened 2024-11-19
[Feature] Support head nodes with SSH fleets


peterschmidt85:
Currently, if you set up an SSH fleet, `dstack` requires that the `dstack` server can access each host via the provided SSH credentials. In many cases, compute hosts aren't accessible directly. Instead, they are accessible through so called "head" nodes.

Setup example:
1. A `head-host` that can access SSH into `compute-host-1` and `compute-host-2`.
2. On a server node, run `dstack server`
3. Create a fleet config:

```
type: fleet
name: private-fleet

placement: cluster

ssh_config:
  hosts:
    - compute-host-1
    - compute-host-2
   proxy_jump: # also can be defined per host; (or it can be a hostname instead of an object)
     hostname: host-node
     identity_file: ~/.ssh/private_key
```

4. The `dstack apply` and `dstack attach` will try to reuse the jump host SSH config for the specified host (`login-node`) automatically.

It means the user must have `head-node` SSH config defined.

timsolovev:
I might be reading the issue wrong, but it might potentially unlock the use of `mpirun`/`mpiexec`, which in some cases are a must-have for infrastructure benchmarking (e.g. [nccl-tests](https://github.com/NVIDIA/nccl-tests)). A critical requirement for `mpirun` is a passwordless SSH access from the head node to workers, so it seems to be adjacent to this opened issue.

Some additional docs on OpenMPI and remote shells: [OpenMPI v4.*](https://www-lb.open-mpi.org/faq/?category=rsh) and [OpenMPI v5](https://docs.open-mpi.org/en/v5.0.x/launching-apps/ssh.html).

PR #2292, opened 2025-02-12
Add head node support for SSH fleets

un-def:
* Add `proxy_jump` property
* Store configuration as part of `RemoteConnectionInfo`
* Always use a project key to connect to SSH instances, drop backward compatibility code (previously, the user-provided key was used, as the project key was not added to the SSH instance, this was fixed in https://github.com/dstackai/dstack/pull/1716)

NOTE: services are not currently supported, proxy support will be added in a separate PR.

Part-of: https://github.com/dstackai/dstack/issues/2010

### DIFF
--- a/docs/docs/concepts/fleets.md
+++ b/docs/docs/concepts/fleets.md
@@ -300,6 +300,47 @@ ssh_config:
     - 3.255.177.52
 ```
 
+#### Head node
+
+If fleet nodes are behind a head node, configure [`proxy_jump`](../reference/dstack.yml/fleet.md#proxy_jump):
+
+<div editor-title="examples/misc/fleets/.dstack.yml">
+
+    ```yaml
+    type: fleet
+    name: my-fleet
+
+    ssh_config:
+      user: ubuntu
+      identity_file: ~/.ssh/worker_node_key
+      hosts:
+        - 3.255.177.51
+        - 3.255.177.52
+      proxy_jump:
+        hostname: 3.255.177.50
+        user: ubuntu
+        identity_file: ~/.ssh/head_node_key
+    ```
+
+</div>
+
+To be able to attach to runs, both explicitly with `dstack attach` and implicitly with `dstack apply`, you must either
+add a front node key (`~/.ssh/head_node_key`) to an SSH agent or configure a key path in `~/.ssh/config`:
+
+<div editor-title="~/.ssh/config">
+
+    ```
+    Host 3.255.177.50
+        IdentityFile ~/.ssh/head_node_key
+    ```
+
+</div>
+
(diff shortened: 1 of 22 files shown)


src/dstack/_internal/server/services/pools.py
```diff
@@ -737,6 +759,8 @@ async def create_ssh_instance_model(
     port: int,
     ssh_user: str,
     ssh_keys: List[SSHKey],
+    ssh_proxy: Optional[SSHConnectionParams],
+    ssh_proxy_keys: list[SSHKey],
```
r4victor: I guess ssh_proxy_keys should be Optional?

docs/docs/concepts/fleets.md
```diff
@@ -300,6 +300,47 @@ ssh_config:
...
+      proxy_jump:
+        hostname: 3.255.177.50
+        user: ubuntu
+        identity_file: ~/.ssh/head_node_key
+    ```
+
+</div>
+
+To be able to attach to runs, both explicitly with `dstack attach` and implicitly with `dstack apply`, you must either
+add a front node key (`~/.ssh/head_node_key`) to an SSH agent or configure a key path in `~/.ssh/config`:
```
r4victor, line 328: I don't understand. Why do we require the user to configure a key path in `~/.ssh/config` if user specifies `identity_file: ~/.ssh/head_node_key`? Can't we just use it to connect to the head node?

un-def: It can be two different users, one creating a fleet has the key, but other users who run workloads may not have the key. With regular setup, the key problem is solved via shim -- shim keeps user's key on the instance while a run is running, but on the head node we don't have any dstack agent to manage authorized keys, thus we require that each user has the head node key on their machine preconfigured. We could download the key to a user's machine though, but I'm not sure if this is a good idea.

src/dstack/_internal/core/services/ssh/attach.py
```diff
@@ -88,28 +90,63 @@ def __init__(
...
                 "StrictHostKeyChecking": "no",
                 "UserKnownHostsFile": "/dev/null",
             }
-        if dockerized and not local_backend:
-            self.container_config = {
+        elif dockerized:
+            if ssh_proxy is not None:
+                # SSH instance with jump host
+                # dstack has no IdentityFile for jump host, it must be either preconfigured
+                # in the ~/.ssh/config of loaded into ssh-agent
```
r4victor:
```suggestion
                # in the ~/.ssh/config or loaded into ssh-agent
```

docs/docs/concepts/fleets.md
r4victor: I'd put this comment somewhere in the code.

r4victor (approved):

Issue #2329, opened 2025-02-20
[Bug]: Generated openapi schema has validation errors


haringsrob:
### Steps to reproduce

On the server head to `/openapi.json` on the host path.

### Actual behaviour

There is a v3.1.0 openapi schema. But when building an sdk with https://github.com/OpenAPITools/openapi-generator the resulting code (in my case rust) various validation errors occur.

Also opening it in https://editor-next.swagger.io/ shows various validation issues.

<img width="905" alt="Image" src="https://github.com/user-attachments/assets/1ef78fe2-fa23-455f-88cd-6cdd32899b3d" />

### Expected behaviour

The api should validate without issues so that sdk's can be generated.

### dstack version

0.18.42

### Server logs

```shell

```

haringsrob:
You can also see errors on the https://dstack.ai/docs/reference/api/rest/ page when you open up the following for example: 

 (scroll up after opening)

<img width="762" alt="Image" src="https://github.com/user-attachments/assets/e3ce8d48-f20c-4356-8984-a15f5958919b" />

PR #2330, opened 2025-02-20
Partially fixes openapi spec

haringsrob:
Relevant info of why this change: https://github.com/fastapi/fastapi/discussions/8449

Fixes part of https://github.com/dstackai/dstack/issues/2329 to where it produces a compilable rust sdk.


jvstme (approved):
@haringsrob, thanks for the fix!

PR #2350, opened 2025-02-24
Use unique names for backend resources

### LINKED ISSUE #1744: [Bug]: Cannot re-create GCP and Azure gateways
### Steps to reproduce

1. Create a gateway on GCP or Azure.
```shell
> cat gateways/gcp.dstack.yml 
type: gateway
name: gcp
backend: gcp
region: europe-west4
domain: example.com

> dstack apply -f gateways/gcp.dstack.yml -y
 BACKEND  REGION        NAME  HOSTNAME  DOMAIN              DEFAULT  STATUS    
 gcp      europe-west4  gcp             example.com                  submitted
```
2. Wait until the gateway is running (see server logs or `dstack gateway -w`)
3. Re-create the gateway
```shell
> dstack apply -f gateways/gcp.dstack.yml -y --force
```

### Actual behaviour

Gateway provisioning fails.
```shell
> dstack gateway
 BACKEND  REGION        NAME  HOSTNAME  DOMAIN              DEFAULT  STATUS    
 gcp      europe-west4  gcp             example.com                  failed
```

### Expected behaviour

Gateway becomes running after re-creation.

### dstack version

master

### Server logs

```shell
[23:52:33] ERROR    dstack._internal.server.background.tasks.process_gateways:145 Got exception when creating gateway compute for gateway gcp                  
                      File "/home/jvstme/git/dstack/dstack/src/dstack/_internal/server/background/tasks/process_gateways.py", line 122, in                     
                    _process_submitted_gateway                                                                                                                 
                        gateway_model.gateway_compute = await create_gateway_compute(                                                                          
                      File "/home/jvstme/git/dstack/dstack/src/dstack/_internal/server/services/gateways/__init__.py", line 127, in create_gateway_compute     
                        gpd = await run_async(                                                                                                                 
                      File "/home/jvstme/git/dstack/dstack/src/dstack/_internal/server/utils/common.py", line 23, in run_async                                 
                      File "/usr/lib64/python3.8/concurrent/futures/thread.py", line 57, in run                                                                
                      File "/home/jvstme/git/dstack/dstack/src/dstack/_internal/core/backends/gcp/compute.py", line 439, in create_gateway                     
                        operation = self.instances_client.insert(request=request)                                                                              
                      File "/home/jvstme/git/dstack/dstack/venv/lib64/python3.8/site-packages/google/cloud/compute_v1/services/instances/client.py", line 4130,
                    in insert                                                                                                                                  
                        response = rpc(                                                                                                                        
                      File "/home/jvstme/git/dstack/dstack/venv/lib64/python3.8/site-packages/google/api_core/gapic_v1/method.py", line 131, in __call__       
                        return wrapped_func(*args, **kwargs)                                                                                                   
                      File "/home/jvstme/git/dstack/dstack/venv/lib64/python3.8/site-packages/google/api_core/grpc_helpers.py", line 76, in                    
                    error_remapped_callable                                                                                                                    
                        return callable_(*args, **kwargs)                                                                                                      
                      File "/home/jvstme/git/dstack/dstack/venv/lib64/python3.8/site-packages/google/cloud/compute_v1/services/instances/transports/rest.py",  
                    line 3197, in __call__                                                                                                                     
                        raise core_exceptions.from_http_response(response)                                                                                     
                    google.api_core.exceptions.Conflict: 409 POST https://compute.googleapis.com/compute/v1/projects/dstack/zones/europe-west4-c/instances: The
                    resource 'projects/dstack/zones/europe-west4-c/instances/gcp' already exists
```

### Additional information

AWS gateways are re-created successfully.

A similar problem occurs when re-creating GCP fleets - the same exception in server logs, although it does not lead to fleets failing because dstack retries fleet provisioning.

### LINKED ISSUE #2341: Use unique names for cloud resources
Currently, dstack uses names for cloud resource that are deterministically inferred from dstack resource names. For example, a job `my-run-0` in project `my-project` creates instance named `my-project-my-run-0`. This is a major problem for backends where names must be unique (e.g. Azure, GCP). It can lead to a dstack resource failing to re-submit/re-create if the previous version of the resource is not yet deleted in the cloud (bugs like #1774). The proposal is to introduce a new approach to naming cloud resources to mitigate this class of problems.

For backends with tags support, every time dstack creates a cloud resource for a dstack resource (e.g. an instance or a volume), it generates a unique random name like `dstack-{random_suffix}` and uses it as the cloud resource name. The dstack resource name is added as a tag `dstack_name` to be able to associate the dstack resource from the cloud.

For backends without tags support, use the current naming approach when possible. Currently, a case is not handled when the dstack name cannot be used in the cloud due to name limitations. In this case, fallback to randomly generated names.

The new approach will solve conflicts problems and also allow us to lift current dstack restriction on resource names (at most 40 chars).

r4victor:
Closes #2341
Fixes #1744

This PR:
* Switches to using the same approach to naming backend resources across all backends. Resources are named `f"{prefix}-{suffix}"` where prefix is `f"dstack-{project}-{resource_name}"[:max_backend_len-9]` and suffix is a randomly generated string of length 8. This approach guarantees that resource names cannot conflict across provisioning attempts (crucial for GCP, Azure), while including the most important info in backend names, thus making it easy to identify the resources.
* Ensures backend resource names are always valid without backend-specific validation. If a part of the name is invalid (e.g. project), it's not included in the name.
* For backends that implement tags, adds `dstack_name` tag that includes full dstack resource name so that the tag can be used for identification if name prefix is truncated.

### DIFF
--- a/src/dstack/_internal/core/backends/base/compute.py
+++ b/src/dstack/_internal/core/backends/base/compute.py
@@ -234,6 +234,21 @@ def generate_unique_instance_name(
     )
 
 
+def generate_unique_instance_name_for_job(
+    run: Run,
+    job: Job,
+    max_length: int = _DEFAULT_MAX_RESOURCE_NAME_LEN,
+) -> str:
+    """
+    Generates a unique instance name for a job valid across all backends.
+    """
+    return generate_unique_backend_name(
+        resource_name=get_job_instance_name(run, job),
+        project_name=run.project_name,
+        max_length=max_length,
+    )
+
+
 def generate_unique_gateway_instance_name(
     gateway_compute_configuration: GatewayComputeConfiguration,
     max_length: int = _DEFAULT_MAX_RESOURCE_NAME_LEN,
--- a/src/dstack/_internal/core/backends/base/compute.py
+++ b/src/dstack/_internal/core/backends/base/compute.py
@@ -1,5 +1,7 @@
 import os
+import random
 import re
+import string
 import threading
 from abc import ABC, abstractmethod
 from functools import lru_cache
@@ -31,6 +33,7 @@ from dstack._internal.core.models.volumes import (
     VolumeAttachmentData,
     VolumeProvisioningData,
 )
+from dstack._internal.core.services import is_valid_dstack_resource_name
 from dstack._internal.utils.logging import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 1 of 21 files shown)


src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -209,8 +212,105 @@ def get_offers_cached(
         return self.get_offers(requirements)
 
 
-def get_instance_name(run: Run, job: Job) -> str:
-    return f"{run.project_name.lower()}-{job.job_spec.job_name}"
+def get_job_instance_name(run: Run, job: Job) -> str:
+    return job.job_spec.job_name
```
jvstme, line 216: (nit) `run` is unused now

src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -209,8 +212,105 @@ def get_offers_cached(
...
+    return generate_unique_backend_name(
+        resource_name=volume.name,
+        project_name=volume.project_name,
+        max_length=max_length,
+    )
+
+
+def generate_unique_backend_name(
+    resource_name: str,
+    project_name: Optional[str],
```
jvstme, line 282: (nit) Supporting `project_name=None` seems redundant since no callers use it

src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -209,8 +212,105 @@ def get_offers_cached(
...
 
 
-def get_instance_name(run: Run, job: Job) -> str:
-    return f"{run.project_name.lower()}-{job.job_spec.job_name}"
+def get_job_instance_name(run: Run, job: Job) -> str:
+    return job.job_spec.job_name
+
+
+_DEFAULT_MAX_RESOURCE_NAME_LEN = 60
+_CLOUD_RESOURCE_SUFFIX_LEN = 8
```
jvstme, line 220: (nit) I think it could be a bit shorter to allow more resource names to fit and still provide decent randomness. E.g. a 4-char suffix can yield 1.7M combinations with digits+lowercase or 14.8M combinations with digits+lowercase+uppercase

src/dstack/_internal/core/backends/runpod/compute.py
```diff
@@ -69,14 +74,15 @@ def run_job(
...
-            instance_name=get_instance_name(run, job),
+            instance_name=get_job_instance_name(run, job),
             ssh_keys=[
                 SSHKey(public=run.run_spec.ssh_key_pub.strip()),
                 SSHKey(public=project_ssh_public_key.strip()),
             ],
             user=run.user,
         )
 
+        pod_name = generate_unique_instance_name(instance_config, max_length=MAX_RESOURCE_NAME_LEN)
```
jvstme, line 85: (nit, unrelated) Looks like constructing `InstanceConfiguration` is redundant here. Instead, RunPod could use `generate_unique_instance_name_for_job` like the other container-based backends

src/dstack/_internal/core/backends/nebius/compute.py
jvstme, line 1: Looks like line 73 needs `generate_unique_instance_name`

src/dstack/_internal/core/backends/datacrunch/compute.py
jvstme, line 1: (nit, unrelated) Line 162 could use `get_job_instance_name` for consistency with other backends

jvstme (approved):
lgtm :+1:

src/dstack/_internal/core/backends/nebius/compute.py
r4victor: I prefer not to update nebius until we can test it. Not sure what max_length is there.

src/dstack/_internal/core/backends/base/compute.py
r4victor: I'm going to address this and other comments in a separate PR when refactoring run_job() for all vm-based backends.

PR #2369, opened 2025-02-27
[Internal]: Update backend contributing docs

jvstme:


jvstme:
@r4victor, we need to share these docs, so I'll merge now. But if you have any comments, I can address them later

Issue #2372, opened 2025-03-03
Simplify implementation of new backends


r4victor:
Currently, adding a new backend involves unnecessary complexity, largely due to the support of legacy features that are no longer needed. The proposal is to make the process of adding a new backend as simple as possible by minimizing dstack-specific knowledge required:

* [x] Move all backend files under the same dir including compute, models, configurator, etc. This should be in the `server/` dir since configurators need to import from other server modules, i.e. something like  `server/services/backends/{backend_name}`. Or even better, make backends independent of server functionality – change `Configurator` not to depend on `BackendModel`, `ProjectModel`, etc. Then all backend files can be put in `core/backends/{backend_name}`.
* [x] Remove `Configurator.get_default_configs()` – it's not used since we dropped automatic backend configuration.
* [x] Replace `Configurator.get_config_values()` with a new method that only performs validation. dstack doesn't need to work with partial configs since we dropped interactive backend configuration. Also, remove `*WithCredsPartial` and `*ConfigValues` models.
* [x] Simplify config models. `*ConfigInfo` models are used as API and configurator models and `*Config` models from `services/config.py` are used as YAML config models. There are also config models for db and Backend/Compute. For most backends, the configs will be the same. But we still need the flexibility to keep them separate. I suggest we allow defining `*Config` (yaml), `*ConfiguratorConfig`, `*BackendConfig`, `*DBConfig` models. Most backends would just define one model that will be used for all of them.
* [x] <s>Consider dropping json-based backend configuration endpoints</s>. #2389 unifies backend models so that maintaining json endpoints does not add much extra maintenance.
* [x] Consider dropping `BackendConfig` base class from base/config.py. Move configs from config.py to models.py. Clarify that `*Config` models are optional backend implementation detail.
* [x] Do not define backend features via lists like `BACKENDS_WITH_CREATE_INSTANCE_SUPPORT`. Introduce Compute properties that define features. The lists can be built dynamically if required.
* [x] Store BackendType in the db as text instead of enum to avoid migrations when adding a new backend (`InstanceModel.backend_type` and `BackendModel.type`).
* [x] <s>Return BackendType as str in the API (or as a enum/str union) to avoid breaking backward compatibility when adding a new backend.</s> Since many models are used as request and response models, replacing enum with str would require validating backend types manually in many places, which would be cumbersome. I suggest we leave it as is since backward compatibility is broken only if someone uses the new backend on the project. It's ok to force everyone to update in that case.

[...]

PR #2389, opened 2025-03-06
Simplify backend configurators and models

r4victor:
Part of #2372 

This PR refactors backends configurators and models so that all backend functionality is kept in `core/backends/{backend_name}`.

* Drops `Configurator.get_default_configs()`.
* Replaces `Configurator.get_config_values()` with `Configurator.validate_config()`.
* Adds basic `validate_config` tests for all backends.
* Makes `Configurator` interface independent of server models by introducing `BackendRecord` and `StoredBackendRecord`.
* Moves configurators to backend dirs.
* Refactors backend models. Replaces `*ConfigInfo`, `*ConfigInfoWithCreds`, and config models in `server/services/config.py` with two models: `*BackendConfig` and `*BackendConfigWithCreds`.
* Moves backend models to backend dirs.
* Moves `Backend` implementations for `__init__.py` to `backend.py` so that same-level backend modules do not import for  `__init__.py`.
* Updates Backend guide to reflect changes.

Breaking changes:
* Drops /api/backends/config_values endpoint used for interactive configuration. Affects only the HTTP API.
* The API accepts/returns azure_config["regions"] instead of azure_config["locations"] (unified with server/config.yml). Affects only the API.

Refactoring Backend and Compute is outside of this PR's scope.


src/dstack/_internal/core/backends/base/configurator.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @abstractmethod
+    def create_backend(
+        self, project_name: str, config: AnyBackendConfigWithCreds
+    ) -> BackendRecord:
+        """
+        Sets up backend given backend config and returns
+        text-encoded config and creds to be stored in the DB.
+        It may perform backend initialization, create
+        cloud resources such as networks and managed identites, and
```
jvstme:
```suggestion
        cloud resources such as networks and managed identities, and
```

contributing/BACKENDS.md
```diff
@@ -132,62 +131,38 @@ Refer to examples:
...
-[azure](https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/core/backends/azure/config.py), etc.
-
-##### 2.4.7. Import config model classes
-
-Ensure the config model classes are imported
-into [`src/dstack/_internal/core/models/backends/__init__.py`](https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/core/models/backends/__init__.py).
+The models should be added to `AnyBackendConfig*` unions in [`src/dstack/_internal/core/backends/models.py`](https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/core/backends/models.py).
 
-##### 2.4.8. Create the configurator class
+It's not required but recommended to define `*BackendStoredConfig` that extends `*BackendConfig` to be able to store extra parameters in the DB. By the same logic, it's recommended to define `*Config` that extends `*BackendStoredConfig` with creds and use it as the main `Backend` and `Compute` config instead of using `*BackendConfigWithCreds` directly.
```
jvstme, line 146: Oh, I finally understood what `*Config` models are for! If they are optional, do you think we could drop the `BackendConfig` base class? Having a base class makes it look like these models are required to be implemented and are used in a generic way.

src/dstack/_internal/core/backends/oci/auth.py
```diff
@@ -20,7 +20,3 @@ def creds_valid(creds: AnyOCICreds) -> bool:
     except any_oci_exception:
         return False
     return True
-
-
-def default_creds_available() -> bool:
-    return creds_valid(OCIDefaultCreds())
```
jvstme, line 26: `creds_valid` is also unused now and can be deleted

src/dstack/_internal/core/backends/oci/configurator.py
```diff
@@ -50,54 +46,20 @@
...
+    def validate_config(self, config: OCIBackendConfigWithCreds, default_creds_enabled: bool):
+        if is_core_model_instance(config.creds, OCIDefaultCreds) and not default_creds_enabled:
             raise_invalid_credentials_error(
                 fields=[["creds"]],
                 details="Default credentials are forbidden by dstack settings",
             )
-
         try:
-            available_regions = get_subscribed_regions(config.creds).names & SUPPORTED_REGIONS
+            get_subscribed_regions(config.creds).names
```
jvstme, line 56: Both `validate_config` and `create_backend` call `get_subscribed_regions`, which is a redundant API call.

Also, the regions validation logic implemented in `create_backend` semantically belongs in `validate_config`.

Some suggestions:
- Consider moving the regions validation logic to `validate_config` and documenting that `validate_config` is allowed to modify the config.
- Or consider documenting that `create_backend` can perform additional validation and raise `ServerClientError` if it fails.
- Or consider merging `validate_config` and `create_backend` in the base class, since they are never used independently - the `create_backend` call simply [follows](https://github.com/dstackai/dstack/blob/38ff7c21d7533f187d395bd8ec3510e722c12ac1/src/dstack/_internal/server/services/backends/__init__.py#L99) the `validate_config` call

contributing/BACKENDS.md
jvstme: Also, consider mentioning that `*Config` is defined in `config.py`. Or move it to `models.py`, I think that can reduce the complexity a bit.

jvstme (approved):

contributing/BACKENDS.md
r4victor: > drop the BackendConfig
> move it to models.py

Yes, that's the plan for a separate PR.

src/dstack/_internal/core/backends/oci/configurator.py
r4victor: Merging `validate_config` and `create_backend` would solve these problems, but I don't want to do it yet cause dry run validation can be a useful thing. For now, I clarified that `create_backend` can perform additional validation. Redundant API calls are relevant only for oci and azure, so not a big problem.

PR #2392, opened 2025-03-10
Introduce ComputeWith classes to detect compute features

r4victor:
Part of #2372

This PR moves optional `Compute` methods to `ComputeWith*` feature classes. This makes it explicit which methods are required for different features and also allows building `BACKENDS_WITH_*` lists dynamically instead of hardcoding them.

Notes:
* Compute instances now need to be type-narrowed like `assert isinstance(compute, ComputeWithPlacementGroupSupport)`. This is still better than previous implicit assumption about method being implemented.
* `BACKENDS_WITH_*` lists now contain only available backend types. This does not affect any behavior except for error messages that return supported backend types. Errors are clarified to return _available_ backend types.

### DIFF
--- a/contributing/BACKENDS.md
+++ b/contributing/BACKENDS.md
@@ -117,13 +117,8 @@ Refer to examples:
 ##### 2.4.4. Create the backend compute class
 
 Under the backend directory you've created, create the `compute.py` file and define the
-backend compute class there (should extend `dstack._internal.core.backends.base.compute.Compute`).
-
-You'll have to implement `get_offers`, `run_job` and `terminate_instance`.
-You may need to implement `update_provisioning_data`, see its docstring for details.
-
-For VM-based backends, also implement the `create_instance` method and add the backend name to
-[`BACKENDS_WITH_CREATE_INSTANCE_SUPPORT`](`https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/core/backends/__init__.py`).
+backend compute class that extends the `dstack._internal.core.backends.base.compute.Compute` class.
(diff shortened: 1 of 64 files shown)


contributing/BACKENDS.md
```diff
@@ -117,13 +117,8 @@ Refer to examples:
 ##### 2.4.4. Create the backend compute class
 
 Under the backend directory you've created, create the `compute.py` file and define the
-backend compute class there (should extend `dstack._internal.core.backends.base.compute.Compute`).
-
-You'll have to implement `get_offers`, `run_job` and `terminate_instance`.
-You may need to implement `update_provisioning_data`, see its docstring for details.
-
-For VM-based backends, also implement the `create_instance` method and add the backend name to
-[`BACKENDS_WITH_CREATE_INSTANCE_SUPPORT`](`https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/core/backends/__init__.py`).
```
jvstme, line 126: (nit) I think it's worth mentioning in the guide that VM-based backends are supposed to extend `ComputeWithCreateInstanceSupport`. This will help contributors find the correct mixin and relevant implementation examples

src/dstack/_internal/server/services/backends/__init__.py
```diff
@@ -364,3 +367,12 @@ async def get_instance_offers(
     offers = heapq.merge(*offers_by_backend, key=lambda i: i[1].price)
     # Put NOT_AVAILABLE, NO_QUOTA, and BUSY instances at the end, do not sort by price
     return sorted(offers, key=lambda i: not i[1].availability.is_available())
+
+
+def check_backed_type_available(backend_type: BackendType):
```
jvstme: `backed` -> `backend`

jvstme (approved):

PR #2397, opened 2025-03-11
Add script to generate boilerplate code for new backend

r4victor:
Part of #2372 

The PR adds a `scripts/add_backend.py` script to generate all the necessary files and classes for a new backend and updates the backend guide to use it.

### DIFF
--- a/scripts/add_backend.py
+++ b/scripts/add_backend.py
@@ -22,10 +22,12 @@ def main():
 
 
 def generate_backend_code(backend_name: str):
+    template_dir_path = Path(__file__).parent.parent.joinpath(
+        "src/dstack/_internal/core/backends/template"
+    )
     env = jinja2.Environment(
-        loader=jinja2.PackageLoader(
-            package_name="dstack",
-            package_path="_internal/core/backends/template",
+        loader=jinja2.FileSystemLoader(
+            searchpath=template_dir_path,
         ),
         keep_trailing_newline=True,
     )
--- a/scripts/add_backend.py
+++ b/scripts/add_backend.py
@@ -1,0 +1,44 @@
+import argparse
+from pathlib import Path
+
+import jinja2
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="This script generates boilerplate code for a new backend"
+    )
+    parser.add_argument(
+        "-n",
+        "--name",
+        help=(
+            "The backend name in CamelCase, e.g. AWS, Runpod, VastAI."
+            " It'll be used for naming backend classes, models, etc."
+        ),
+        required=True,
+    )
+    args = parser.parse_args()
+    generate_backend_code(args.name)
+
+
+def generate_backend_code(backend_name: str):
+    env = jinja2.Environment(
+        loader=jinja2.PackageLoader(
+            package_name="dstack",
+            package_path="_internal/core/backends/template",
+        ),
+        keep_trailing_newline=True,
+    )
+    backend_dir_path = Path(__file__).parent.parent.joinpath(
+        f"src/dstack/_internal/core/backends/{backend_name.lower()}"
+    )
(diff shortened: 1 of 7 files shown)


scripts/add_backend.py
```diff
@@ -0,0 +1,44 @@
...
+    args = parser.parse_args()
+    generate_backend_code(args.name)
+
+
+def generate_backend_code(backend_name: str):
+    env = jinja2.Environment(
+        loader=jinja2.PackageLoader(
+            package_name="dstack",
+            package_path="_internal/core/backends/template",
+        ),
```
jvstme: (nit) `_internal/core/backends/template/*.jinja2` is not actually included in the package, so this won't work in the general case. It does work for me when `dstack` is installed in editable mode, but this is probably an implementation detail of pip, so I wouldn't rely on it.

I'd suggest using `jinja2.FileSystemLoader` instead.

src/dstack/_internal/core/backends/template/configurator.py.jinja
```diff
@@ -0,0 +1,70 @@
...
+    {{ backend_name }}Config,
+    {{ backend_name }}Creds,
+    {{ backend_name }}StoredConfig,
+)
+from dstack._internal.core.models.backends.base import (
+    BackendType,
+)
+
+# TODO: Add all supported regions and default regions
+REGIONS = []
```
jvstme, line 23: Hardcoded regions are not needed for most backends, so I wouldn't include this in the template.

I guess they used to be needed for interactive setup, but now they are only needed for backends with custom-built VM images that are not available in all regions. For other backends, hardcoded regions are rather harmful, as they prevent users from using newly added regions

src/dstack/_internal/core/backends/template/compute.py.jinja
```diff
@@ -0,0 +1,72 @@
...
+):
+    def __init__(self, config: {{ backend_name }}Config):
+        super().__init__()
+        self.config = config
+
+    def get_offers(
+        self, requirements: Optional[Requirements] = None
+    ) -> List[InstanceOfferWithAvailability]:
+        # If the provider is added to gpuhunt, you'd typically get offers
+        # using `get_catalog_offers()` and extend them with availability info.
```
jvstme, line 47: (nit) Consider adding a sample call to `get_catalog_offers`, as contributors can [forget](https://github.com/dstackai/dstack/pull/2132#discussion_r1905947967) to pass important arguments, such as `locations` or `configurable_disk_size`.

```python

offers = get_catalog_offers(
    backend=BackendType.{{ backend_name|upper }},
    requirements=requirements,
    configurable_disk_size=...,  # TODO: set in case of boot volume size limits
)
```

src/dstack/_internal/core/backends/template/models.py.jinja
```diff
@@ -0,0 +1,58 @@
+from typing import Annotated, List, Literal, Optional, Union
+
+from pydantic import Field
+
+from dstack._internal.core.models.common import CoreModel
+
+
+# The template uses "api_key" creds as a most popular creds type.
```
jvstme:
```suggestion
# The template uses "api_key" creds as the most popular creds type.
```

jvstme (approved):
:+1:

src/dstack/_internal/core/backends/template/configurator.py.jinja
r4victor: Regions don't need to be hardcoded but they need to be validated, and hardcoding them seems to be the best option for most GPU clouds (have a few regions, don't add new regions often, unlikely to have an API to get all the regions).

r4victor: Anyway, feel free to adjust the template on regions.

PR #2403, opened 2025-03-13
Transition high-level Python runs API to new methods

### LINKED ISSUE #2402: Update Python API
The current high-level Python API is not fully transitioned to the new plan/apply approach. The major pain point is that [`get_plan()` and `submit()` methods](https://github.com/dstackai/dstack/blob/09992a0f5ffedabc5e9c0ee9a21953eb2ae06558/src/dstack/api/_public/runs.py#L391) accept profile parameters as arguments. They are hard to maintain and are not kept up-to-date. Introduce new methods with better signatures to make it closer to CLI and HTTP API (they should accept single profile parameter).

r4victor:
Closes #2402

The PR:
* Replaces high-level Python runs API `get_plan`, `exec_plan`, `submit` methods with new methods  `get_run_plan`, `apply_plan`, `apply_configuration`. Old methods are deprecated.
* Makes Python API docs up-to-date.
* Deprecates the `/api/project/{project_name}/runs/submit` endpoint.
* Removes some old client code to handle backward compatibility with old servers.


src/dstack/api/_public/runs.py
```diff
@@ -403,6 +402,117 @@ def __init__(
...
+        """
+        Apply the run configuration.
+        Use this method to apply configurations without getting a run plan first.
+
+        Args:
+            configuration (Union[Task, Service, DevEnvironment]): The run configuration.
+            repo (Union[LocalRepo, RemoteRepo, VirtualRepo]): The repo to mount to the run.
+            profile: The profile to use for the run.
+            configuration_path: The path to the configuration file. Omit if the configuration is not loaded from a file.
+
```
un-def, line 505: `reserve_ports` description is missing

PR #2414, opened 2025-03-17
Restore removed models fields for compatibility with 0.18 clients

r4victor:
The PR restores recently removed models fields to keep the server compatible with 0.18 clients that continue to send those fields as `null`. It also excludes the fields so that the new clients do not send them and the fields can be removed in the future.

### DIFF
--- a/src/dstack/_internal/core/models/envs.py
+++ b/src/dstack/_internal/core/models/envs.py
@@ -102,10 +102,11 @@ class Env(BaseModel):
     def copy(self, **kwargs) -> Self:
         # Env.copy() is tricky because it copies only the hidden top-level {"__root__": {...}}
         # structure, not the actual nested dict representing the env itself.
-        # To avoid possible bugs, we prohibit shallow copying altogether.
+        # So we copy __root__ explicitly in case of a shallow copy.
+        new_copy = super().copy(**kwargs)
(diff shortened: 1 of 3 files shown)


src/dstack/_internal/core/models/envs.py
```diff
@@ -102,10 +102,11 @@ def __setitem__(self, item, value):
     def copy(self, **kwargs) -> Self:
         # Env.copy() is tricky because it copies only the hidden top-level {"__root__": {...}}
         # structure, not the actual nested dict representing the env itself.
-        # To avoid possible bugs, we prohibit shallow copying altogether.
+        # So we copy __root__ explicitly in case of a shallow copy.
+        new_copy = super().copy(**kwargs)
         if not kwargs.get("deep", False):
-            raise TypeError(f"shallow copying of {self.__class__.__name__} is prohibited")
-        return super().copy(**kwargs)
+            new_copy.__root__ = new_copy.__root__.copy()
```
r4victor, line 108: @un-def, I had to implement shallow copy for Env somehow because if a model defines excludes/includes, pydantic performs shallow copy for fields internally when calling copy(deep=True). Yes, copy(deep=True) may call copy(deep=False).

un-def (approved):

PR #2415, opened 2025-03-17
[Examples] Update nccl-tests

un-def:
* Replace `un1def/aws-efa-test` with `dstackai/efa`
* Use `DSTACK_GPUS_NUM`


peterschmidt85:
@un-def please review finally!

PR #2422, opened 2025-03-17
[Docker] Add dstackai/efa image

un-def:


jvstme (approved):
@un-def, this is image is for the [NCCL Tests](https://github.com/dstackai/dstack/tree/2b3e95e428b3f2d060acd441f13825cf234627c3/examples/misc/nccl-tests) example, right?

un-def:
@jvstme, the image is mainly for EFA stack tests (thus the name), nccl-tests are included as a sample workload, but yes, I'll update the [NCCL Tests](https://github.com/dstackai/dstack/tree/2b3e95e428b3f2d060acd441f13825cf234627c3/examples/misc/nccl-tests) example to use this image once it's ready.

PR #2434, opened 2025-03-18
Add new metrics

un-def:
* `memory_total_bytes`
* `gpu_memory_total_bytes`
* `cpus_detected_num`


src/dstack/_internal/server/services/metrics.py
```diff
@@ -131,6 +149,14 @@ def _calculate_job_metrics(points: Sequence[JobMetricsPoint]) -> JobMetrics:
     return JobMetrics(metrics=metrics)
 
 
+def _make_constant_metric(name: str, timestamps: list[datetime], value: float) -> Metric:
+    return Metric(
+        name=name,
+        timestamps=[timestamps[0], timestamps[-1]] if len(timestamps) > 1 else [timestamps[0]],
```
r4victor: I think we need to document the format for constant metrics. Also unsure if it's worth having a special case for constant metrics at all to save some bytes. I'd return all the timestamps.

r4victor (approved):

src/dstack/_internal/server/services/metrics.py
un-def: Removed the special case, constant metrics now use the same timestamps as regular metrics.

PR #2444, opened 2025-03-22
Add TensorrRT-LLM Example

Bihan:
I would like your suggestions on below listed points.

For @peterschmidt85 
1. The latest NGC container `nvcr.io/nvidia/tritonserver:25.02-trtllm-python-py3` does not support Deepseek-R1. So either user needs to build image using main branch or may be we  provide the image as below. I have already built the image during benchmarks, so just need to push into dstackai docker respository.
```
# Use Docker image built using https://github.com/NVIDIA/TensorRT-LLM.git main branch
image: dstackai/tensorrt_llm:latest 
```
For @jvstme 
1. The streaming response from `tritonserver` is as follows:
`data: {"id":"cmpl-b65ad9a7-063f-11f0-9720-654fce3b6072","choices":[{"delta":{"content":"","function_call":null,"role":"assistant"},"logprobs":null,"finish_reason":null,"index":0}],"created":1742553116,"model":"ensemble","system_fingerprint":null,"object":"chat.completion.chunk"}`

2. The streaming response from `pytorch backend` is as follows:
`data: {"model":"deepseek-ai/DeepSeek-R1","choices":[{"index":0,"delta":{"content":" I"},"finish_reason":null}]}`

In the response we can see that `"system_fingerprint":null`, and our proxy throws validation error as below for the query
```
curl http://localhost:3000/proxy/models/bihan/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer 47a4b92c-3442-4f2b-a8f3-2bb52e97d7b2" \
  -d '{
  "model": "ensemble",
  "messages": [
    {
      "role": "user",
      "content": "Hello world"
    }
  ],
  "stream": true,
  "max_tokens": 512
}'
```
**Validation Errors**
1. Triton Server
`{"detail":"Invalid chunk in model stream: 1 validation error for ChatCompletionsChunkResponse\nsystem_fingerprint\n  none is not an allowed value (type=type_error.none.not_allowed)"}`
2. Pytorch Backend
`{"detail":"Invalid chunk in model stream: 2 validation errors for ChatCompletionsChunkResponse\nid\n  field required (type=value_error.missing)\ncreated\n  field required (type=value_error.missing)"}`

Therefore, I have modified types of fields of `ChatCompletionsChunk` as below
1. `system_fingerprint` from `str` to `Optional[str]`
2.  `id` from `str` to `Optional[str] = None`
3. `created` from `int` to `Optional[int] = None`

PR #2452, opened 2025-03-28
Update ChatCompletionsChunk for Deepseek-R1 response

Bihan:
This PR will update the ChatCompletionsChunk as described in https://github.com/dstackai/dstack/pull/2444. This update is removed from https://github.com/dstackai/dstack/pull/2444


peterschmidt85:
Is it ready to merge? Then go for it!

PR #2454, opened 2025-03-28
Update NIM example with DeepSeek-R1-Distill

Bihan:


docs/examples.md
```diff
@@ -38,7 +38,7 @@ hide:
            NIM
        </h3>
        <p>
-            Deploy Llama 3.1 with NIM
+            Deploy DeepSeek-R1-Distill with NIM
```
peterschmidt85: "DeepSeek R1 Distill Llama 8B"?

peterschmidt85 (approved):

docs/examples.md
Bihan: No specific reason. Before the NIM was using Llama 3.1 8B, so modified it to use similar 8B distill model.

examples/deployment/nim/.dstack.yml
```diff
@@ -1,7 +1,7 @@
 type: service
-name: qwen-nim
+name: serve-distill-deepseek
 
-image: nvcr.io/nim/qwen/qwen-2.5-7b-instruct:latest
+image: nvcr.io/nim/deepseek-ai/deepseek-r1-distill-llama-8b:1.5.2
```
peterschmidt85: Do they have `latest`?

PR #2455, opened 2025-03-28
Migrate to uv

colinjc:
Was curious about the level of effort this would take with Cursor and got bored, so here's a quick and dirty attempt to moving over to uv.


uv.lock
r4victor, line 1: I don't think uv.lock should be in the git since libraries do not usually have dependencies pinned.

r4victor:
@colinjc, We do not currently have issues with pip/setuptools so there needs to be a rationale for moving to uv. I may assume it may be beneficial for using in CI if it speeds up installation. Any numbers on that compared to dstack pip installation times?

If you gonna proceed with the PR, CONTRIBITING should be updated to use uv, as well as all GitHub actions.

r4victor:
Answering my own question: dstack installation time in CI went from 1m to 5s, which is substantial improvement.

colinjc:
> Answering my own question: dstack installation time in CI went from 1m to 5s, which is substantial improvement.

Also makes checking out the repo for the first time to contribute much faster! 

> If you gonna proceed with the PR, CONTRIBITING should be updated to use uv, as well as all GitHub actions.

Done

contributing/DEVELOPMENT.md
r4victor, line 1: dstack should be installed in editable mode for development. Need to figure out how to do it properly with uv.

Maybe https://docs.astral.sh/uv/concepts/projects/dependencies/#editable-dependencies

r4victor:
@colinjc, I don't have a time for a thorough review atm, but the PR certainly has some TODOs and is not ready for a merge. I think someone from the core team will be able to move it forward in a few weeks.

contributing/DEVELOPMENT.md
colinjc: `src/dstack` is already in "editable" mode when you run uv sync. Do you mean the other python code like the gateway should also be installed in editable? 

https://docs.astral.sh/uv/concepts/projects/sync/#editable-installation

r4victor: Ah, right, I'm not yet familiar with uv for development

r4victor (approved):

r4victor:
Made a few changes and fixes and merged. Works great for development and CI. Also tested `uv publish` on test.pypi.org. We'll see how the release workflow works with the 0.19.4rc1 release.

@colinjc thanks a lot for the initiative!

PR #2463, opened 2025-04-02
Add Nebius backend

### LINKED ISSUE #2390: [Feature]: Nebius backend
### Problem

Nebius AI Cloud is not available in `dstack`.

There is an irrelevant implementation of the `nebius` backend that does not work.

### Solution

Drop the irrelevant `nebius` backend implementation and implement the backend using the relevant Nebius AI Cloud API.

### Workaround

Use Nebius via SSH fleets.

Yes

jvstme:
Closes #2390


src/dstack/_internal/core/backends/nebius/compute.py
```diff
@@ -0,0 +1,266 @@
...
+        try:
+            instance = resources.get_instance(self._sdk, instance_id)
+        except RequestError as e:
+            if e.status.code != StatusCode.NOT_FOUND:
+                raise
+            instance = None
+        if instance is not None:
+            if instance.status.state != instance.status.InstanceState.DELETING:
+                resources.delete_instance(self._sdk, instance_id)
+                raise WorkRemains(
```
r4victor: I'd raise ComputeSpecific-errors and handle it accordingly instead of introducing generic errors. At least, it keeps the error hierarchy - Compute returns ComputeError.

r4victor (approved):

r4victor (approved):

PR #2466, opened 2025-04-02
[CLI] Divide CPU util by a number of vCPUs in `dstack metrics`

un-def:
Previously, the max value was (100 * N)% where N is a number of vCPUs, e.g., 3200% for 32-core CPU. This change normalizes CPU util range to 0..100%

### DIFF
--- a/src/dstack/_internal/cli/commands/metrics.py
+++ b/src/dstack/_internal/cli/commands/metrics.py
@@ -14,6 +14,7 @@ from dstack._internal.cli.utils.common import (
     console,
 )
 from dstack._internal.core.errors import CLIError
+from dstack._internal.core.models.instances import Resources
 from dstack._internal.core.models.metrics import JobMetrics
 from dstack.api._public import Client
 from dstack.api._public.runs import Run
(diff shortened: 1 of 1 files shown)


un-def:
On 32 vCPUs:

```
sysbench cpu run --time=3600 --threads=8
```

```
 NAME        CPU  MEMORY         GPU
 nvidia-smi  25%  102MB/31990MB  #0 1517MB/8192MB 16% Util
```

src/dstack/_internal/cli/commands/metrics.py
```diff
@@ -86,14 +87,23 @@ def _get_metrics_table(run: Run, metrics: List[JobMetrics]) -> Table:
         add_row_from_dict(table, run_row)
 
     for job, job_metrics in zip(run._run.jobs, metrics):
+        jrd = job.job_submissions[-1].job_runtime_data
+        jpd = job.job_submissions[-1].job_provisioning_data
+        resources: Optional[Resources] = None
+        if jrd is not None and jrd.offer is not None:
+            resources = jrd.offer.instance.resources
+        elif jpd is not None:
+            resources = jpd.instance_type.resources
```
jvstme, line 96: Is there a difference between getting the offer from jrd or jpd? As far as I can see, they are set at the same time in `_process_submitted_job`

jvstme (approved):

src/dstack/_internal/cli/commands/metrics.py
un-def: The jpd offer is an instance offer, while the jrd offer is a job offer, which is a fraction of an instance offer in the case of the blocks feature. New job always have the jrd offer, but we still have to fall back to jpd for backward compatibility as there can be jobs submitted before the jrd offer was introduced.

PR #2488, opened 2025-04-08
Add `fleets` property to run configurations and CLI

### LINKED ISSUE #1447: Allow to pass `--fleet NAME` to `dstack run` (and also specify from YAML)

un-def:
Closes: https://github.com/dstackai/dstack/issues/1447


peterschmidt85:
@un-def The property in the run configuration YAML file (and API) should be `fleets` and must accept an array. Is it implemented this way? Just want to confirm.

un-def:
@peterschmidt85 it accepts both a name and a list of names. I chose the singular form because I think it will be mostly used with a single fleet

peterschmidt85:
> @peterschmidt85 it accepts both a name and a list of names. I chose the singular form because I think it will be mostly used with a single fleet

I strongly suggest to use plural - similar to `regions` and `backends` and always require a list.

r4victor (approved):

PR #2492, opened 2025-04-09
Update Llama example

Bihan:
Update Llama example with Llama 4 Scout.

With TGI Llama 4 did not work. I have created an [issue](https://github.com/huggingface/text-generation-inference/issues/3153). After the issue is resolved, I will be update the example.


peterschmidt85 (approved):
Left minor changes. If all is good, feel free to merge

PR #2502, opened 2025-04-11
Update Axolotl Examples

Bihan:
1. Updated Axolotl Nvidia Example with Llama 4 Scout

2. Resolve module not found error for AMD
**Error**
`ModuleNotFoundError: No module named 'pynvml.nvml'; 'pynvml' is not a package`
**Solution**
Installed previous release of pynvml
`pip install pynvml==11.5.3`

### DIFF
--- a/examples/accelerators/amd/README.md
+++ b/examples/accelerators/amd/README.md
@@ -161,13 +161,18 @@ To request multiple GPUs, specify the quantity after the GPU name, separated by 
     
     ```yaml
     type: task
+    # The name is optional, if not specified, generated randomly
     name: axolotl-amd-llama31-train
-    
+
     # Using RunPod's ROCm Docker image
     image: runpod/pytorch:2.1.2-py3.10-rocm6.0.2-ubuntu22.04
     # Required environment variables
     env:
       - HF_TOKEN
+      - WANDB_API_KEY
+      - WANDB_PROJECT
+      - WANDB_NAME
+      - HUB_MODEL_ID
     # Commands of the task
     commands:
       - export PATH=/opt/conda/envs/py_3.10/bin:$PATH
@@ -177,6 +182,8 @@ To request multiple GPUs, specify the quantity after the GPU name, separated by 
       - cd axolotl
       - git checkout d4f6c65
       - pip install -e .
+      - pip uninstall pynvml -y
+      - pip install pynvml==11.5.3
       - cd ..
       - wget https://dstack-binaries.s3.amazonaws.com/flash_attn-2.0.4-cp310-cp310-linux_x86_64.whl
       - pip install flash_attn-2.0.4-cp310-cp310-linux_x86_64.whl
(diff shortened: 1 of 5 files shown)


examples/accelerators/amd/README.md
```diff
@@ -161,13 +161,18 @@ To request multiple GPUs, specify the quantity after the GPU name, separated by
...
     name: axolotl-amd-llama31-train
-    
+
     # Using RunPod's ROCm Docker image
     image: runpod/pytorch:2.1.2-py3.10-rocm6.0.2-ubuntu22.04
     # Required environment variables
     env:
       - HF_TOKEN
+      - WANDB_API_KEY
+      - WANDB_PROJECT
```
peterschmidt85, line 173: How is `WANDB_API_KEY` not enough?

examples/accelerators/amd/README.md
```diff
@@ -177,6 +182,8 @@ To request multiple GPUs, specify the quantity after the GPU name, separated by
       - cd axolotl
       - git checkout d4f6c65
       - pip install -e .
+      - pip uninstall pynvml -y
+      - pip install pynvml==11.5.3
```
peterschmidt85, line 187: Should we add a note or at least a comment on it?

examples/accelerators/amd/README.md
```diff
@@ -177,6 +182,8 @@ To request multiple GPUs, specify the quantity after the GPU name, separated by
       - cd axolotl
       - git checkout d4f6c65
```
peterschmidt85, line 183: Why this particular revision??

examples/accelerators/amd/README.md
Bihan: Yes. I will add.

examples/accelerators/amd/README.md
Bihan: No, we need to set WANDB_PROJECT and WANDB_NAME. 

The difference is that in our current master it is set in the [config file ](https://github.com/dstackai/dstack/blob/a39e946280a7886763e2f7b8eaeebe13aa09ebb4/examples/fine-tuning/axolotl/config.yaml#L30) and in this PR, we are sending it as an argument. When we send it as argument, we do not need to include the config.yaml in our repo.

peterschmidt85: Okay, then at least let's hardcode the value of `WANDB_NAME`, e.g. to `axolotl-amd-llama31-train`. If the user wants, they would change it.

peterschmidt85: BTW this is another user case when we could set it to `$DSTACK_RUN_NAME`

examples/accelerators/amd/README.md
Bihan: xformers is incompatible with ROCm. Axolotl suggests applying workarounds as suggested in this [link](https://docs.axolotl.ai/docs/amd_hpc.html#apply-xformers-workaround)

In the revision `d4f6c65`, workaround is implemented. This is how ROCm builds Axolotl image. [link](https://github.com/ROCm/rocm-blogs/blob/d0422efb34fc34922ad16a9a16c79b12031141b4/blogs/artificial-intelligence/axolotl/src/Dockerfile.rocm#L65)

peterschmidt85: Then a comment is needed I suppose

Bihan: Yes. I will update it accordingly.

peterschmidt85 (approved):
Please feel free to merge

PR #2516, opened 2025-04-15
Use large runners for slow workflow jobs

r4victor:
https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners


r4victor:
For some reason larger runners queue jobs so that the build workflow is slower although individual jobs are faster.

PR #2529, opened 2025-04-17
Update TGI Example with Llama 4 Scout

Bihan:


peterschmidt85 (approved):
Don't forget to update `examples.md`  to replace Llama 3.1 to Llama 4 Scout

PR #2534, opened 2025-04-18
Updated runner and shim contributing guide

peterschmidt85:
* Updated README to include instructions on how to test shim and runner remotely
* Added justfile with common reciples to build and upload runner and shim


runner/README.md
```diff
@@ -76,6 +78,29 @@ The `local` backend will submit the run to the locally started shim and runner.
...
+Build the runner and shim, and upload them to S3 automatically using `just` (see [justfile](../justfile)).
+
+```shell
+just upload
+```
+
+To use the built shim and runner with the dstack server, pass the URLs via `DSTACK_SHIM_DOWNLOAD_URL` and `DSTACK_RUNNER_DOWNLOAD_URL`:
+
+```shell
+DSTACK_SHIM_DOWNLOAD_URL=https://dstack-runner-downloads-stgn.s3.eu-west-1.amazonaws.com/0.0.0/binaries/dstack-shim-linux-amd64
```
r4victor: If multiple people use the same 0.0.0 version, they'll be overriding each other build?

justfile
r4victor, line 1: Let's put it in runner/

runner/README.md
```diff
@@ -76,6 +78,29 @@ The `local` backend will submit the run to the locally started shim and runner.
...
+
+You can also test the built shim and runner using standard backends (including SSH fleets).
+
+> [!NOTE]
+> To run with standard backends, both the runner and shim must be built for linux/amd64.
+
+Build the runner and shim, and upload them to S3 automatically using `just` (see [justfile](../justfile)).
+
+```shell
+just upload
```
r4victor, line 101: How should external contributors use it if they don't have access to dstack-runner-downloads-stgn?

justfile
peterschmidt85: I thought of it but there are two points.
1. It's very inconvenient to change the current directory. I prefer to always stay at the root. The whole value of `just` is to have a list of common recipes across the project. A good example is frontend, in order to build frontend I always have to leave the root. This is super inconvenient.
2. I'd like to have both shim and frontend recipes at the root level instead of having multiple just files.

runner/README.md
peterschmidt85: I will extract the bucket name variable and document it.

justfile
un-def: `just` supports [imports](https://github.com/casey/just?tab=readme-ov-file#imports), we can put this justfile in runner/ and import it in the root justfile

runner/.justfile
```diff
@@ -1,61 +1,58 @@
...
-# Default recipe to display available commands
 default:
     @just --list
 
 # Export version variable
+# Must be changed to a unique value to avoid conflicts with other developers using the same bucket
 export version := "0.0.0"
 
+# S3 bucket name - must be set to a bucket that the developer has access to
+export s3_bucket := "dstack-runner-downloads-stgn"
```
r4victor: Maybe allow setting via an env variable so that contributors do not need to modify a file?

https://just.systems/man/en/functions.html#environment-variables

PR #2540, opened 2025-04-21
[Feature]: Support `dstack offer` #2142

### LINKED ISSUE #2142: [Feature]: Allow listing available key resources such as `gpu`, `region`,  and `backends`
### Problem 1 (major): listing available GPUs per region and backend

As a user, sometimes, I want to know which GPUs are available within my project (and in which regions). Currently, I either have to specify a wide range in YAML and then try `dstack apply` and explore offers, or check it via specific backends' console UI (e.g. I do it often with RunPod - because they have a great and convenient UI for that). 
It would be great if I could list available GPUs per region and backend and their availability (if this information is available) via the CLI.

Also, a minor yet related issue is when I don't know how certain GPUs are named with specific backends. This issue though should be better addressed by ensuring that `dstack` uses unified GPU names across all vendors and backends. But this may not be very easy to implement. That's why this issue is also related.

Note, the SSH backend is also relevant here as I may want to see what fleets are available (if I use `dstack` heavily on-pem machines).

### Problem 2 (medium): listing available regions per backend

As a user, I may often want to see what regions are available so I can configure the backend or run accordingly, or to create volumes. Currently, I always have to either read the backend's documentation or use their console UI.

>Overall, if I have to use backend's UI to work with backends, this isn't very productive nor convenient. We strive to make dstack the most simple way to work with multiple clouds. For that, we need to ensure all is possible without using backends' console UI.

### Problem 3 (minor): listing available backends

As a user, sometimes I may want to know what backends are available. Currently, this information is only available via UI or the YAML files if you're a server admin. It would be a lot more convenient if I could access it via the CLI.

### Solution

```shell
$ dstack list gpu [--backend BACKEND] [--region REGION] [--name NAME] [--all]

| <backend> | <GPU name/memory> | GPU vendor | <region> | <availability/quota information> |
```

```shell
$ dstack list region [--backend BACKEND] 

| <backend> | <region> |
```

```shell
$ dstack list backend

| <backend> |
```

### Workaround

_No response_

Yes

peterschmidt85:


src/dstack/_internal/server/schemas/runs.py
```diff
@@ -26,6 +26,7 @@ class GetRunRequest(CoreModel):
 
 class GetRunPlanRequest(CoreModel):
     run_spec: RunSpec
+    max_offers: int = Field(50, ge=1, le=1000)
```
r4victor: Should be excluded on the client if not set explicitly to handle old servers that don't support `max_offers`.

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -0,0 +1,199 @@
...
+        conf = TaskConfiguration(commands=[":"])
+
+        configurator = OfferConfigurator(api_client=self.api)
+        configurator.apply_args(conf, args, [])
+        profile = load_profile(Path.cwd(), profile_name=args.profile)
+
+        if args.format == "json":
+            run_spec = RunSpec(
+                configuration=conf,
+                ssh_key_pub=Path(self.api.ssh_identity_file + ".pub").read_text().strip(),
```
jvstme: (nit) I think you can use a dummy value to avoid reading from disk and avoid unnecessary errors if the key is misconfigured.

Same on line 84

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -0,0 +1,199 @@
...
+                run_spec = RunSpec(
+                    configuration=conf,
+                    ssh_key_pub=Path(self.api.ssh_identity_file + ".pub").read_text().strip(),
+                    profile=profile,
+                )
+                run_plan = self.api.client.runs.get_plan(
+                    self.api.project,
+                    run_spec,
+                    max_offers=args.max_offers,
+                )
```
jvstme: (nit) To avoid code duplication, you can conditionally use `contextlib.nullcontext`. Also move `RunSpec` construction out of the context manager.
```python
run_spec = ...
if args.format == "plain":
    status = console.status("Getting offers...")
else:
    status = contextlib.nullcontext()
with status:
    run_plan = ...
```

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -0,0 +1,199 @@
...
+
+            for offer in job_plan.offers:
+                availability = None
+                if offer.availability in {
+                    InstanceAvailability.NOT_AVAILABLE,
+                    InstanceAvailability.NO_QUOTA,
+                    InstanceAvailability.IDLE,
+                    InstanceAvailability.BUSY,
+                }:
+                    availability = offer.availability.value
```
jvstme: I'd suggest dropping this and just passing `offer.availability.value` to the json output. `None` is not a very intuitive availability representation.

This condition is only needed when showing a table in the CLI, where we prefer to omit `InstanceAvailability.AVAILABLE` and `InstanceAvailability.UNKNOWN` for a shorter output

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -0,0 +1,199 @@
...
+        if args.format == "json":
+            output = {
+                "project": run_plan.project_name,
+                "user": run_plan.user,
+                "resources": job_plan.job_spec.requirements.resources.dict(),
+                "max_price": (job_plan.job_spec.requirements.max_price),
+                "spot": run_spec.configuration.spot_policy,
+                "reservation": run_plan.run_spec.configuration.reservation,
+                "offers": [],
+            }
```
jvstme: (nit) Add `total_offers`? Otherwise, there is no way to tell if all offers were returned of some were capped by `--max-offers`

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -0,0 +1,199 @@
...
+            default="plain",
+            help="Output format (default: plain)",
+        )
+        self._parser.add_argument(
+            "--json",
+            action="store_const",
+            const="json",
+            dest="format",
+            help="Output in JSON format (equivalent to --format json)",
+        )
```
jvstme, line 57: Actually, do we need to support the JSON output at all? This will be yet another place where we'll have to maintain backward compatibility, e.g. changing or dropping fields will not be as easy as in the regular CLI output. I can see where the JSON output can be useful, but in such cases users can use the API instead

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if job_plan.total_offers > len(job_plan.offers):
+            table.add_row("", "...", style="secondary")
+        console.print(table)
+        if job_plan.total_offers > len(job_plan.offers):
+            console.print(
+                f"[secondary]Shown {len(job_plan.offers)} of {job_plan.total_offers} offers, "
+                f"${job_plan.max_price:g} max[/]"
+            )
+        console.print()
```
jvstme: Please consider refactoring to avoid code duplication with the `print_run_plan` function.

There are also some discrepancies between this implementation and `print_run_plan` that I think are worth removing:

- This implementation does not include the `INSTANCE` column. I think it's worth including, especially for teams that are using one or two backends and are used to backend-specific instance types.
- This implementation does not show a warning if there are no offers. I think it's worth showing for a neater output and for the troubleshooting link.
- This implementation rounds the price to 3 digits after the decimal point, which can decrease accuracy. For example, some AWS prices have 4 digits after the decimal point.

src/dstack/_internal/cli/services/configurators/run.py
```diff
@@ -274,7 +274,7 @@ def delete_configuration(
         console.print(f"Run [code]{conf.name}[/] deleted")
 
     @classmethod
-    def register_args(cls, parser: argparse.ArgumentParser):
+    def register_args(cls, parser: argparse.ArgumentParser, default_max_offers: int = 3):
```
jvstme, line 277: (nit) This new parameter does not match the `BaseApplyConfigurator.register_args` interface and does not seem to be necessary

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -0,0 +1,199 @@
...
+        cls,
+        parser: argparse.ArgumentParser,
+    ):
+        super().register_args(parser, default_max_offers=50)
+
+
+# TODO: Support aggregated offers
+# TODO: Add tests
+class OfferCommand(APIBaseCommand):
+    NAME = "offer"
```
jvstme, line 40: I'd prefer plural `dstack offers`. We already have plural `dstack metrics`, not `dstack metric`.

For other commands, singular makes sense, because they can work with individual entities, e.g. there is `dstack volume delete`, `dstack gateway update`, etc. But `dstack offers` and `dstack metrics` are only used for listing and are unlikely to ever provide any functionality related to individual offers or metrics.

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -0,0 +1,199 @@
...
+                        ),
+                        "region": offer.region,
+                        "resources": offer.instance.resources.dict(),
+                        "spot": offer.instance.resources.spot,
+                        "price": float(offer.price),
+                        "availability": availability,
+                    }
+                )
+
+            print(json.dumps(output, indent=2))
```
jvstme, line 113: The JSON output doesn't seem to work if there are any GPUs.
```shell
> dstack offer --gpu 1.. --json
  File "/dstack/venv/bin/dstack", line 10, in <module>
    sys.exit(main())
  File "/dstack/src/dstack/_internal/cli/main.py", line 87, in main
    args.func(args)
  File "/dstack/src/dstack/_internal/cli/commands/offer.py", line 129, in _command
    print(json.dumps(output, indent=2))
  File "/usr/lib64/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib64/python3.10/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/usr/lib64/python3.10/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib64/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib64/python3.10/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/usr/lib64/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib64/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib64/python3.10/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/usr/lib64/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib64/python3.10/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/usr/lib64/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type AcceleratorVendor is not JSON serializable
```

jvstme (commented):
@peterschmidt85, thanks for the PR, I've tested it and it works well as a prototype. Except for the JSON output, which I don't think is really worth supporting at all

peterschmidt85:
> @peterschmidt85, thanks for the PR, I've tested it and it works well as a prototype. Except for the JSON output, which I don't think is really worth supporting at all

@jvstme Sorry, not sure I understood your comment about `--json`. This is going to be useful for anyone who would like to use `dstack offer` for monitoring GPUs or stats collection (instead of `gpuhunt`).

src/dstack/_internal/cli/services/configurators/run.py
peterschmidt85: Not sure I understood what you mean by `doesn't match`.  This is required so `dstack apply` uses `3` as the default.

src/dstack/_internal/cli/commands/offer.py
peterschmidt85: See my other comment on this. The main idea is to let anyone use `dstack offer` instead of `gpuhunt`.
Backward compatibility is an issue but for `dstack offer` we aren't required to promise it at this point of time. It's just enough that `dstack offer` itself is backward compatible. The rest will be managed by the version of `dstack` CLI on the client.

src/dstack/_internal/cli/commands/offer.py
peterschmidt85: See my other comment on this. The main idea is to let anyone use `dstack offer` instead of `gpuhunt`.
Backward compatibility is an issue but for `dstack offer` we aren't required to promise it at this point of time. It's just enough that `dstack offer` itself is backward compatible. The rest will be managed by the version of `dstack` CLI on the client.

Unfortunatelly our API is not convenient to use. CLI is 10-x easier.

src/dstack/_internal/cli/commands/offer.py
peterschmidt85: > The JSON output doesn't seem to work if there are any GPUs.

@jvstme How can I reproduce it?

For me, it seems to work:

```
dstack offer --gpu amd --max-offers 10 --json
{
  "project": "main",
  "user": "admin",
  "resources": {
    "cpu": {
      "min": 2,
      "max": null
    },
    "memory": {
      "min": 8.0,
      "max": null
    },
    "shm_size": null,
    "gpu": {
      "vendor": "amd",
      "name": null,
      "count": {
        "min": 1,
        "max": 1
      },
      "memory": null,
      "total_memory": null,
      "compute_capability": null
    },
    "disk": {
      "size": {
        "min": 100.0,
        "max": null
      }
    }
  },
  "max_price": null,
  "spot": null,
  "reservation": null,
  "offers": [
    {
      "backend": "runpod",
      "region": "EU-RO-1",
      "resources": {
        "cpus": 24,
        "memory_mib": 289792,
        "gpus": [
          {
            "name": "MI300X",
            "memory_mib": 196608,
            "vendor": "amd"
          }
        ],
        "spot": false,
        "disk": {
          "size_mib": 102400
        },
        "description": "24xCPU, 283GB, 1xMI300X (192GB), 100.0GB (disk)"
      },
      "spot": false,
      "price": 2.49,
      "availability": null
    }
  ]
}
```

src/dstack/_internal/cli/services/configurators/run.py
jvstme: Never mind, I didn't notice that the default is different in `dstack apply` and `dstack offer`

src/dstack/_internal/cli/commands/offer.py
jvstme: @peterschmidt85, what's your Python and pydantic version? I can reproduce on Python 3.10, 3.12, 3.13, pydantic 1.10.21

peterschmidt85: Python 3.13.3
`pydantic                       1.10.21`

peterschmidt85: @jvstme Can you paste here the non-json output?

src/dstack/_internal/cli/commands/offer.py
jvstme: > let anyone use dstack offer instead of gpuhunt

They are quite different - `dstack offer --json` requires running a dstack server and configuring all the backends, while `gpuhunt` works as a library and does not require configuring credentials for most backends.

Additionally, we don't know if anyone is using `gpuhunt`, so there doesn't seem to be any demand for this.

> CLI is 10-x easier.

I don't think using the CLI is easy for this purpose. If someone needs to get offers as JSON, they probably want to automate their monitoring and integrate it with their application. I think it is easier to make an API call from their application instead of installing the dstack CLI in some environment available to their application (the application is not necessarily written in Python) and calling shell commands programmatically.

Another advantage of the API is that it has automatically generated docs and an OpenAPI schema. If we will support `dstack offer --json`, should we also document the schema of its output? Will we maintain it manually?

> Unfortunatelly our API is not convenient to use

Then we can create a new API for this purpose. We might need to do it anyway to support aggregates. But not in this PR.

To summarize, these are my arguments against `dstack offer --json`:
- More code to maintain.
- Will need to maintain backward compatibility of the returned schema.
- Might need to document the returned schema.
- Not easy to integrate with external applications.
- No demand.

src/dstack/_internal/cli/commands/offer.py
jvstme:
```shell
> dstack offer --gpu 1.. --max-offers 3
 Project      ilya                                    
 User         admin                                   
 Resources    2..xCPU, 8GB.., 1..xGPU, 100GB.. (disk) 
 Max price    -                                       
 Spot policy  auto                                    
 Reservation  -                                       

 #  BACKEND  REGION  RESOURCES                                      SPOT   PRICE   
 1  runpod   CZ      6xCPU, 62GB, 1xA2000 (6GB), 100.0GB (disk)     yes   $0.060   
 2  runpod   TT      8xCPU, 30GB, 1xRTX3070 (8GB), 100.0GB (disk)   yes   $0.070   
 3  runpod   US      16xCPU, 31GB, 1xRTX3070 (8GB), 100.0GB (disk)  yes   $0.070   
    ...                                                                            
Shown 3 of 4277 offers, $127.816 max

> dstack offer --gpu 1.. --max-offers 3 --json
```

jvstme: Debugger on this line:
image
So `output` does contain enums (`AcceleratorVendor`), which are not json serializable

peterschmidt85: Pending on https://github.com/dstackai/gpuhunt/pull/135

peterschmidt85: Per Google Meet call:
* Agreed to give `--json` a shot as an experimental feature to see initial feedback.

src/dstack/api/server/_runs.py
```diff
@@ -96,6 +98,17 @@ def _get_apply_plan_excludes(plan: ApplyRunPlanInput) -> Optional[Dict]:
...
 
+def _get_get_plan_excludes(request: GetRunPlanRequest) -> Optional[Dict]:
+    """
+    Excludes new fields when they are not set to keep
+    clients backward-compatibility with older servers.
+    """
+    run_spec_excludes = _get_run_spec_excludes(request.run_spec)
+    if request.max_offers is None:
+        run_spec_excludes["max_offers"] = True
+    return run_spec_excludes
```
r4victor, line 109: Must return excludes for GetRunPlanRequest, not run_spec_excludes, i.e. excludes["max_offers"] = True and excludes["run_spec"] = run_spec_excludes.

src/dstack/_internal/server/schemas/runs.py
```diff
@@ -26,6 +26,7 @@ class GetRunRequest(CoreModel):
 
 class GetRunPlanRequest(CoreModel):
     run_spec: RunSpec
+    max_offers: Optional[int] = Field(description="The maximum number of offers to return")
```
r4victor: Should have limits

src/dstack/api/server/_runs.py
peterschmidt85: @r4victor `run_spec_excludes` is a dict that has `run_spec`.
The final dict has both `run_spec` and `max_offers` so all should be good.

src/dstack/_internal/server/schemas/runs.py
peterschmidt85: @r4victor Do you think 10000 upper limit is ok?

src/dstack/api/server/_runs.py
r4victor: Ah, ok

src/dstack/_internal/server/schemas/runs.py
r4victor: probably

peterschmidt85:
@r4victor @jvstme Ready to merge

src/dstack/_internal/cli/utils/run.py
```diff
@@ -76,29 +79,31 @@ def th(s: str) -> str:
...
+        props.add_row(th("Creation policy"), creation_policy)
+        props.add_row(th("Idle duration"), idle_duration)
     props.add_row(th("Reservation"), run_plan.run_spec.configuration.reservation or "-")
 
     offers = Table(box=None)
     offers.add_column("#")
     offers.add_column("BACKEND")
     offers.add_column("REGION")
-    offers.add_column("INSTANCE")
+    offers.add_column("INSTANCE TYPE")
```
jvstme, line 100: (nit) I prefer the old column name, as the new name can waste horizontal space and the second word can look like a separate column
```
 #  BACKEND  REGION          INSTANCE TYPE  RESOURCES                   SPOT  PRICE     
 1  aws      ap-southeast-1  t2.small       1xCPU, 2GB, 100.0GB (disk)  yes   $0.0039   
 2  aws      us-east-2       t2.small       1xCPU, 2GB, 100.0GB (disk)  yes   $0.004    
 3  aws      us-east-1       t2.small       1xCPU, 2GB, 100.0GB (disk)  yes   $0.0063
```

src/dstack/_internal/cli/utils/run.py
```diff
@@ -76,29 +79,31 @@ def th(s: str) -> str:
     props.add_row(th("Project"), run_plan.project_name)
     props.add_row(th("User"), run_plan.user)
     props.add_row(th("Configuration"), run_plan.run_spec.configuration_path)
```
jvstme: I assume `Configuration` and `Max duration` are not supposed to be shown in `dstack offer`
```shell
> dstack offer --gpu 1..
 Project        main                                    
 User           admin                                   
 Configuration                                          
 Resources      2..xCPU, 8GB.., 1..xGPU, 100GB.. (disk) 
 Max price      -                                       
 Max duration   -                                       
 Spot policy    auto                                    
 Reservation    -
```

jvstme (approved):

src/dstack/_internal/cli/utils/run.py
peterschmidt85: INSTANCE is misleading. we want to use `INSTANCE` too to indicate the exact instance.
In most cases, the instance type is much longer:

```
#   BACKEND     REGION             INSTANCE TYPE                   RESOURCES                                  SPOT  PRICE                   
 1   datacrunch  FIN-01             1H100.80S.30V                   30xCPU, 120GB, 1xH100 (80GB), 100.0GB      no    $2.19                   
                                                                    (disk)                                                                   
 2   datacrunch  FIN-02             1H100.80S.30V                   30xCPU, 120GB, 1xH100 (80GB), 100.0GB      no    $2.19                   
                                                                    (disk)                                                                   
 3   datacrunch  FIN-02             1H100.80S.32V                   32xCPU, 185GB, 1xH100 (80GB), 100.0GB      no    $2.19                   
                                                                    (disk)                                                                   
 4   datacrunch  ICE-01             1H100.80S.32V                   32xCPU, 185GB, 1xH100 (80GB), 100.0GB      no    $2.19                   
                                                                    (disk)                                                                   
 5   runpod      US-KS-2            NVIDIA H100 PCIe                16xCPU, 251GB, 1xH100 (80GB), 100.0GB      no    $2.39                   
```

jvstme (approved):

PR #2581, opened 2025-04-29
Implement plugins

### LINKED ISSUE #2576: Support server-side plugins
The proposal is to introduce a server-side plugin system for dstack. An MVP is to allow plugins modify configurations/specs on apply.

r4victor:
Closes #2576 

The PR adds support for server-side dstack plugins. Plugins are enabled via server/config.yml:

```yaml
plugins:
  - my_dstack_plugin
  - some_other_plugin
projects:
- name: main
```

To implement a plugin, plugin authors create a Python package with a `dstack.plugins.Plugin` subclass and export it as a package entrypoint:

```toml
[project.entry-points."dstack.plugins"]
my_dstack_plugin = "my_dstack_plugin:MyPlugin"
```

Currently, the only plugins functionality is `ApplyPolicy` that change apply specs (run/fleet/volume/gateway).

Notes:
* `ApplyPolicy.on_apply()` will usually be called twice: on getting a plan and on apply. So if `on_apply()` is non-deterministic, the spec displayed in the plan can be different from the actual applied spec. Discussed with @peterschmidt85 that this is a desired behavior as otherwise plugin authors would need to double-check the modified spec.

A plugin guide is TBD.


pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "psutil",
     "gpuhunt>=0.1.3,<0.2.0",
     "argcomplete>=3.5.0",
+    "backports.entry-points-selectable",
```
un-def: Can't it be moved to `server` extras?

docs/docs/concepts/plugins.md
```diff
@@ -0,0 +1,113 @@
...
+    def on_run_apply(self, user: str, project: str, spec: RunSpec) -> RunSpec:
+        # Forcing some limits
+        spec.configuration.max_price = 2.0
+        spec.configuration.max_duration = "1d"
+        # Setting some extra tags
+        if spec.configuration.tags is None:
+            spec.configuration.tags = {}
+        spec.configuration.tags |= {
+            "team": "my_team",
+        }
```
un-def, line 103: I would add an example of how to check configuration type:

```python
from dstack.api import Service

...
    # if spec.configuration.type == "service":  won't narrow the Python type
    if isinstance(spec.configuration, Service):
       spec.configuration.https = True  # service-specific property
```

un-def (approved):

Issue #2598, opened 2025-05-04
[Bug]: dstack CLI crashes with OSError when run in subprocess or non-TTY environment


vuyelwadr:
### Steps to reproduce

run this script 

```
import subprocess
import os

cmd = ["dstack", "ps"]
env = os.environ.copy()
env["DSTACK_NO_TTY"] = "1"
env["NO_COLOR"] = "1"

proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, env=env, universal_newlines=True)
for line in proc.stdout:
    print(repr(line))
proc.wait()
```

### Actual behaviour

The CLI crashes with the following error:
Traceback (most recent call last):
  File "/Users/x/.local/bin/dstack", line 10, in <module>
    sys.exit(main())
  File "/Users/x/.local/share/uv/tools/dstack/lib/python3.10/site-packages/dstack/_internal/cli/main.py", line 87, in main
    args.func(args)
  File "/Users/x/.local/share/uv/tools/dstack/lib/python3.10/site-packages/dstack/_internal/cli/commands/ps.py", line 44, in _command
    console.print(run_utils.get_runs_table(runs, verbose=args.verbose))
  File "/Users/x/.local/share/uv/tools/dstack/lib/python3.10/site-packages/dstack/_internal/cli/utils/run.py", line 152, in get_runs_table
    table = Table(box=None, expand=os.get_terminal_size()[0] <= 110)
OSError: [Errno 25] Inappropriate ioctl for device

### Expected behaviour

The CLI should not crash.
If [os.get_terminal_size()] fails, it should use a default width for table rendering, so that dstack can be used in scripts, automation, and CI environments.

### dstack version

0.19.7

### Server logs

```shell

```

### Additional information

This bug prevents any automation or monitoring of dstack runs from Python or other scripting languages.
The fix is to wrap all calls to [os.get_terminal_size()] in a try/except block and use a default width (e.g., 120) if the call fails.
See also: related PR with fix.

vuyelwadr:
PR https://github.com/dstackai/dstack/pull/2599

PR #2599, opened 2025-05-04
Fix: handle OSError from os.get_terminal_size() in CLI table rendering for non-TTY environments

vuyelwadr:
## Summary

This PR fixes a bug in the dstack CLI where running commands like `dstack ps` in a non-interactive environment (subprocess, PTY, or script) would crash with:

  File "/Users/x/.local/bin/dstack", line 10, in <module>
    sys.exit(main())
  File "/Users/x/.local/share/uv/tools/dstack/lib/python3.10/site-packages/dstack/_internal/cli/main.py", line 87, in main
    args.func(args)
  File "/Users/x/.local/share/uv/tools/dstack/lib/python3.10/site-packages/dstack/_internal/cli/commands/ps.py", line 44, in _command
    console.print(run_utils.get_runs_table(runs, verbose=args.verbose))
  File "/Users/x/.local/share/uv/tools/dstack/lib/python3.10/site-packages/dstack/_internal/cli/utils/run.py", line 152, in get_runs_table
    table = Table(box=None, expand=os.get_terminal_size()[0] <= 110)
OSError: [Errno 25] Inappropriate ioctl for device

The error was caused by unguarded calls to `os.get_terminal_size()` in the CLI’s table rendering code. This patch wraps those calls in a try/except block and uses a default width (120) if the call fails, ensuring the CLI works in both TTY and non-TTY environments.

## Details

- All uses of `os.get_terminal_size()` are now wrapped in try/except.
- If an OSError is raised, a default width of 120 is used.
- This prevents crashes when running dstack in automation, CI, or as a subprocess.

## Steps to Reproduce

1. Run `dstack ps` in a Python subprocess or PTY (not a real terminal).
2. Observe the crash with OSError.
3. With this patch, the command works and outputs as expected.

## Why this is needed

- Many users want to automate or monitor dstack runs from scripts or other tools.
- The CLI should not crash in non-interactive environments.

## Related Issue

- https://github.com/dstackai/dstack/issues/2598

---

**Thank you for reviewing!**

### DIFF
--- a/src/dstack/_internal/cli/utils/run.py
+++ b/src/dstack/_internal/cli/utils/run.py
@@ -95,7 +95,11 @@ def print_run_plan(
             props.add_row(th("Inactivity duration"), inactivity_duration)
     props.add_row(th("Reservation"), run_spec.configuration.reservation or "-")
 
-    offers = Table(box=None, expand=os.get_terminal_size()[0] <= 110)
+    try:
+        width = os.get_terminal_size()[0]
+    except OSError:
+        width = 120  # Default width for non-TTY
+    offers = Table(box=None, expand=width <= 110)
     offers.add_column("#")
     offers.add_column("BACKEND", style="grey58", ratio=2)
     offers.add_column("RESOURCES", ratio=4)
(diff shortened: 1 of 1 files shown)


src/dstack/_internal/cli/utils/run.py
```diff
@@ -95,7 +95,11 @@ def th(s: str) -> str:
             props.add_row(th("Inactivity duration"), inactivity_duration)
     props.add_row(th("Reservation"), run_spec.configuration.reservation or "-")
 
-    offers = Table(box=None, expand=os.get_terminal_size()[0] <= 110)
+    try:
+        width = os.get_terminal_size()[0]
```
peterschmidt85: IMO, it's better to use something like this:

```
import shutil
table = Table(box=None, expand=shutil.get_terminal_size(fallback=(120, 40)).columns <= 110)
```

peterschmidt85 (changes requested):

Issue #2591, opened 2025-05-01
[Bug]: GCP: "Placement policy is not applicable to machine type <type>" when creating multi-node fleet


un-def:
### Steps to reproduce

Apply the configuration:

```yaml
type: fleet
name: fleet-gcp-2x
placement: cluster
nodes: 2
backends: [gcp]
instance_types: [n1-standard-1]
resources:
  gpu: 1
```

### dstack version

0.19.6

### Server logs

```shell
[11:23:08] ERROR    dstack._internal.server.background.tasks.process_instances:588 Got exception when launching n1-standard-1 in gcp/us-east1
                    Traceback (most recent call last):
                      File "/home/def/dev/dstack/src/dstack/_internal/server/background/tasks/process_instances.py", line 572, in
                    _create_instance
                        job_provisioning_data = await run_async(
                                                ^^^^^^^^^^^^^^^^
                        ...<3 lines>...
                        )
                        ^
                      File "/home/def/dev/dstack/src/dstack/_internal/utils/common.py", line 20, in run_async
                        return await asyncio.get_running_loop().run_in_executor(None, func_with_args)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                      File "/home/def/.local/share/pyenv/versions/3.13.0/lib/python3.13/concurrent/futures/thread.py", line 58, in run
                        result = self.fn(*self.args, **self.kwargs)
                      File "/home/def/dev/dstack/src/dstack/_internal/core/backends/gcp/compute.py", line 319, in create_instance
                        operation = self.instances_client.insert(request=request)
                      File
                    "/home/def/.local/share/virtualenvs/dstack/lib/python3.13/site-packages/google/cloud/compute_v1/services/instances/client.
                    py", line 4288, in insert
                        response = rpc(
                            request,
                        ...<2 lines>...
                            metadata=metadata,
                        )
                      File "/home/def/.local/share/virtualenvs/dstack/lib/python3.13/site-packages/google/api_core/gapic_v1/method.py", line
                    131, in __call__
                        return wrapped_func(*args, **kwargs)
                      File "/home/def/.local/share/virtualenvs/dstack/lib/python3.13/site-packages/google/api_core/grpc_helpers.py", line 76,
                    in error_remapped_callable
                        return callable_(*args, **kwargs)
                      File
[...]

PR #2609, opened 2025-05-07
Use AS_COMPACT collocation for GCP A3 clusters

r4victor:
Similar to #2587 but only for A3 instances to avoid #2591

### DIFF
--- a/src/dstack/_internal/core/backends/gcp/compute.py
+++ b/src/dstack/_internal/core/backends/gcp/compute.py
@@ -76,6 +76,10 @@ class GCPVolumeDiskBackendData(CoreModel):
     disk_type: str
 
 
+class GCPPlacementGroupBackendData(CoreModel):
+    collocation: str
+
+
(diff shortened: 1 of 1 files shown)


jvstme (approved):

r4victor:
Although AS_COMPACT collocation improves performance for small number of nodes due to the nodes being located on the same physical host, it degrades offers availability. So I was not able to provision 16 A3 high spots with AS_COMPACT. It leads to:

```
Got GCP error when provisioning a VM: 503      
                    SERVICE UNAVAILABLE                            
                    ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS: The 
                    zone 'projects/dstack/zones/europe-west4-c'    
                    does not have enough resources available to    
                    fulfill the request.
```

COLLOCATED provisions 16 spots just fine.

r4victor:
Not merging unless we get a better understanding of how AS_COMPACT works

PR #2627, opened 2025-05-13
Introduce JOB_DISCONNECTED_RETRY_TIMEOUT

### LINKED ISSUE #2626: Do not fail jobs immediately in case of connectivity issues
Currently process_running_jobs background task fails jobs with INTERRUPTED_BY_NO_CAPACITY immediately in case ssh connections cannot be established (3 attempts with 1s interval). There should be a more graceful handling of connectivity issues (e.g. for the idle instance to be terminated it has to be unavailable for 20m). We could start by introducing a larger timeout interval before the job is marked as failed, like 1m or 2m. 

INTERRUPTED_BY_NO_CAPACITY can also be replaced for on-demand instances with a new termination reason like INSTANCE_UNREACHABLE.

r4victor:
Closes #2626 

The PR changes process_running_jobs handling of ssh connection errors for pulling/running jobs. Previously, it failed jobs with no capacity after establishing ssh connection fails. Now, it keeps job active for at least JOB_DISCONNECTED_RETRY_TIMEOUT (2 minutes). This allows surviving temporary networking issues.

A possible regression is that dstack becomes slower to replace failed jobs/replicas. IMO, not a big problem since there are no expectations/guarantees on how quick the replacement is.

Later we can make timeout smarter depending on the job (e.g. short for spots, if retry enabled, etc). For now, keeping it as constant for simplicity.

Also introduced JobTerminationReason.INSTANCE_UNREACHABLE – not yet used because of backward compatibility.


src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -299,19 +305,38 @@ async def _process_running_job(session: AsyncSession, job_model: JobModel):
...
+                    "%s: failed because shim/runner returned an error, age=%s",
+                    fmt(job_model),
+                    job_submission.age,
+                )
+                job_model.status = JobStatus.TERMINATING
+                # job will be terminated and instance will be emptied by process_terminating_jobs
+            else:
+                # No job_model.termination_reason set means ssh connection failed
+                if job_model.disconnected_at is None:
+                    job_model.disconnected_at = common_utils.get_current_datetime()
```
jvstme, line 323: Note that `runner_ssh_tunnel` also does [3 blocking retries](https://github.com/dstackai/dstack/blob/c521a57779dbbdbdc5284d88515c75f2c512cf28/src/dstack/_internal/server/services/runner/ssh.py#L25), and each can take [15 seconds](https://github.com/dstackai/dstack/blob/c521a57779dbbdbdc5284d88515c75f2c512cf28/src/dstack/_internal/core/services/ssh/tunnel.py#L20). So this `disconnected_at` can be imprecise, since we only set it after the `runner_ssh_tunnel` retries.

But I'm not sure we need to do anything about it. I thought about disabling the `runner_ssh_tunnel` retries in favor of the new `disconnected_at` retries, but having retry logic in both places may further improve stability.

jvstme (approved):

src/dstack/_internal/server/background/tasks/process_running_jobs.py
r4victor: Thanks for pointing out. Let's keep both! :)

PR #2628, opened 2025-05-13
Keep last metrics for finished jobs

### LINKED ISSUE #2618: [Feature]: Ensure metrics are shown also for finished runs
### Problem

Currently, both the CLI and UI show metrics only for active runs.

### Solution

_No response_

### Workaround

_No response_

Yes

un-def:
The retention window is 1800 seconds (last 30 minutes) by default, configurable via the `DSTACK_SERVER_METRICS_WINDOW_SECONDS` environment variable.

Closes: https://github.com/dstackai/dstack/issues/2618


peterschmidt85:
Anything required on the UI's side?

r4victor:
So the last 30 minutes of logs are stored forever after the run is finished?

@peterschmidt85 is this what you proposed or having logs for some time after runs finish would be sufficient?

peterschmidt85:
Ideally of course we could introduce a TTL but I assume we don't support it yet for logs too?

r4victor:
@peterschmidt85, sorry I meant metrics of course not logs. We already had a ttl for metrics:

https://github.com/dstackai/dstack/blob/c521a57779dbbdbdc5284d88515c75f2c512cf28/src/dstack/_internal/server/settings.py#L48

And the ttl applied to finished runs as well. So you could see metrics of finished runs for this ttl.

So what this PR does I believe is keeps last metrics of finished runs forever.

peterschmidt85:
That was TTL for metrics of a live run. For finished run, I guess it makes sense to set TTL as a week or so. We discussed that with @un-def no?

r4victor:
> That was TTL for metrics of a live run.

Well, it applied to all runs before. If the idea is to introduce different TTLs for active and finished runs, it's not what the PR does.

un-def:
> That was TTL for metrics of a live run. For finished run, I guess it makes sense to set TTL as a week or so. We discussed that with @un-def no?

I've just replaced the retention window with a simpler solution using two separate TTL settings for running and finished jobs.

By default:

* `DSTACK_SERVER_METRICS_RUNNING_TTL_SECONDS` — 1 hour
* `DSTACK_SERVER_METRICS_FINISHED_TTL_SECONDS` — 1 week

r4victor (approved):

PR #2631, opened 2025-05-13
Add REST plugin for user-defined policies

### LINKED ISSUE #2630: [Feature]: Support REST-based plugins
### Problem

There is currently support for plugins, however it requires building plugin code into the Dstack Docker image, . In addition, in some cases users may need to add their own pre-apply validations and policies (that are not applicable to all dstack users). They cannot do this without having to commit the code into Dstack's codebase.

### Solution

Implement a plugin that makes REST calls to a configurable server. This server could be used to offload pre-apply policies or any other customer-specific pre-apply hooks to mutate or reject requests. Mutating requests is useful for injecting cost tracking metadata for example.

### Workaround

None

Yes

Nadine-H:
Resolves https://github.com/dstackai/dstack/issues/2630.

Key notes:
- The REST plugin is a builtin plugin, unlike how other plugins are defined and loaded in the dstack system.
- The REST plugin has to be enabled through the dstack server config if the user wishes to hook up a plugin server.


Nadine-H:
Realized that setting up this plugin as an external package is unnecessarily complicating the installation process. I'm going to change it so that it's more builtin the server code.

src/plugins/rest_plugin/src/rest_plugin.py
```diff
@@ -0,0 +1,63 @@
...
+            return spec
+        except pydantic.ValidationError as e:
+            # TODO: check response error status and report if plugin service rejected request as invalid
+            logger.exception(f"Plugin service returned invalid response:\n{response.text if response else None}")
+            logger.info("Returning original run spec")
+            return spec
+        logger.info(f"Using RunSpec from plugin service:\n{spec}")
+        return spec
+    def on_run_apply(self, user: str, project: str, spec: RunSpec) -> RunSpec:
+        return self._call_plugin_service(user, project, spec, '/runs/pre_apply')
```
r4victor: I'd suggest to name endpoints the same way Python methods are structured: `/runs/pre_apply` => `/apply_policies/on_run_apply`. Maybe it's more verbose, but we won't need to think hard how to name new endpoints in the future.

src/plugins/rest_plugin/src/rest_plugin.py
```diff
@@ -0,0 +1,63 @@
...
+from dstack._internal.core.errors import ServerError
+from dstack._internal.core.models.fleets import FleetSpec
+from dstack._internal.core.models.gateways import GatewaySpec
+from dstack._internal.core.models.volumes import VolumeSpec
+from dstack.plugins import ApplyPolicy, Plugin, RunSpec, get_plugin_logger
+from dstack.plugins._models import ApplySpec
+
+logger = get_plugin_logger(__name__)
+
+PLUGIN_SERVICE_URI_ENV_VAR_NAME = "DSTACK_PLUGIN_SERVICE_URI"
```
r4victor: If the goal of this plugin is to allow enabling http-based plugins that don't require installing python packages, then there should be a way to configure multiple endpoints using this plugin, which is not possible with one env variable. We could probably support this once we allow specifying plugin options via yaml, and users will be able to specify rest_plugin multiple times (also need to support multiple registrations of the same plugin):

```
plugins:
  - rest_plugin:
      uri: http://localhost:8004
  - rest_plugin:
      uri: http://localhost:8005
```

This is just a notice for future development, not necessarily applicable to the first plugin version.

r4victor:
> Realized that setting up this plugin as an external package is unnecessarily complicating the installation process. I'm going to change it so that it's more builtin the server code.

A good place for a built-in server plugins would be to put them in `src/dstack/_internal/server/services/plugins/builtin/rest_plugin`

May require renaming `src/dstack/_internal/server/services/plugins.py` to `src/dstack/_internal/server/services/plugins/__init__.py`

Let me know if you need help with refactoring.

r4victor:
> I'll create an openapi spec doc

@Nadine-H, Do you think of writing the spec manually or generating it automatically? Automatic generation is highly preferable because spec models are regularly updated with new properties and it won't be practical to maintain manually written spec.

src/plugins/rest_plugin/src/rest_plugin.py
Nadine-H: I like that approach, even though it doesn't really follow the REST naming convention, but I'm cool with it if you are.

Nadine-H:
> > I'll create an openapi spec doc
> 
> @Nadine-H, Do you think of writing the spec manually or generating it automatically? Automatic generation is highly preferable because spec models are regularly updated with new properties and it won't be practical to maintain manually written spec.

Automatic generation for sure. I set up a local FastAPI server to test the rest plugin and will share the `openai.json` with you shortly.

src/dstack/plugins/builtin/rest_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+        response = None
+        try:
+            response = requests.post(
+                f"{self._plugin_service_uri}{endpoint}",
+                json=spec_request.dict(),
+                headers={"accept": "application/json", "Content-Type": "application/json"},
+                timeout=PLUGIN_REQUEST_TIMEOUT,
+            )
+            response.raise_for_status()
+            spec_json = json.loads(response.text)
```
dakota-shop: How does the Plugin reject a Run from being applied? Do we want some exceptions (HTTP Status Codes) to be propagated to Dstack? Or should there be an `errors` element of the response body?

I'm leaning toward the latter since if we're monitoring the plugin server, it might be difficult to distinguish between "the plugin server thought this request was invalid" and "the plugin server could not answer the question".

Nadine-H:
Adding documentation to the guides as well

src/dstack/plugins/builtin/rest_plugin.py
Nadine-H: Sorry, I misread your comment when you shared it internally. Yes - I agree on the latter option, that way the user knows if the action is on their end or if it's a system issue (a bug in dstack or plugin server)

examples/plugins/example_plugin_server/app/main.py
```diff
@@ -0,0 +1,43 @@
...
+from app.utils import configure_logging
+
+configure_logging()
+logger = logging.getLogger(__name__)
+
+app = FastAPI()
+
+
+@app.post("/apply_policies/on_run_apply")
+async def on_run_apply(request: RunSpecRequest):
```
colinjc: Declare the return type?

src/dstack/plugins/builtin/rest_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+    spec: SpecType
+
+
+RunSpecRequest = SpecRequest[RunSpec]
+FleetSpecRequest = SpecRequest[FleetSpec]
+VolumeSpecRequest = SpecRequest[VolumeSpec]
+GatewaySpecRequest = SpecRequest[GatewaySpec]
+
+
+class CustomApplyPolicy(ApplyPolicy):
```
colinjc: @r4victor Should async policies be a thing, or are policies always running in an executor?

src/dstack/plugins/builtin/rest_plugin.py
Nadine-H: Added a pydantic model for the plugin server response that includes an `error` field [here](https://github.com/dstackai/dstack/pull/2631/commits/e757a77789072cf8c810e9c47e60bc1e9e396d84#diff-cefd5b27016f62b497ed8ccba936cd08c172b06b328ae446e99d600ab1982de3R32). Expectation is if request is rejected, dstack should a 200 response from plugin server with an `error`.

src/dstack/plugins/builtin/rest_plugin.py
r4victor: They should always be running in an executor. They are not at the moment. We'll fix it.

src/dstack/plugins/builtin/rest_plugin.py
r4victor, line 1: The rest plugin should be moved to `_internal/` since everything outside `_internal/` including  
`src/dstack/plugins/` is meant for the users of the `dstack` package. The rest plugin semantically belongs to internal server code, so as I pointed out before, a good place would be `src/dstack/_internal/server/services/plugins/builtin/rest_plugin`

r4victor: @Nadine-H, tell me if you need help with refactoring.

src/dstack/plugins/builtin/models.py
r4victor, line 1: The models seem specific to rest_plugin, so makes sense to have them in `builtin/rest_plugin/_models.py`

src/dstack/plugins/builtin/rest_plugin.py
r4victor: Ah, I see that the rest plugin models are meant to be imported from within the plugin server. So that makes sense to have rest plugin in `src/dstack/plugins/`. Still let's structure it like so

* `src/dstack/plugins/builtin/rest_plugin/_models.py` – models defined, not to be imported from here
* `src/dstack/plugins/builtin/rest_plugin/_plugin.py` – the plugin code
* `src/dstack/plugins/builtin/rest_plugin/__init__.py` – expose models and possibly other helper functions for rest plugin here

So you would import it like `dstack.plugins.builtin.rest_plugin import FleetSpecRequest`.

It's more consistent with the dstack approach to exposing public members of the package. @Nadine-H what do you think?

src/dstack/plugins/builtin/rest_plugin.py
```diff
@@ -0,0 +1,119 @@
...
+
+import requests
+from pydantic import ValidationError
+
+from dstack._internal.core.errors import ServerClientError
+from dstack._internal.core.models.fleets import FleetSpec
+from dstack._internal.core.models.gateways import GatewaySpec
+from dstack._internal.core.models.volumes import VolumeSpec
+from dstack.plugins import ApplyPolicy, Plugin, RunSpec, get_plugin_logger
+from dstack.plugins._models import ApplySpec
```
r4victor: `ApplySpec` can be imported like `from dstack.plugins import ApplySpec`.

examples/plugins/example_plugin_server/README.md
```diff
@@ -0,0 +1,30 @@
+## Overview
+
+If you wish to hook up your own plugin server through `dstack` builtin ` rest-plugin`, here's a basic example on how to do so.
```
r4victor:
```suggestion
If you wish to hook up your own plugin server through `dstack` builtin `rest_plugin`, here's a basic example on how to do so.
```

examples/plugins/example_plugin_server/README.md
```diff
@@ -0,0 +1,30 @@
...
+    uv sync 
+    ```
+
+1. Start the plugin server locally:
+
+    ```bash
+    fastapi dev app/main.py 
+    ```
+
+1. Enable `rest-plugin` in `dstack` `server/config.yaml`:
```
r4victor:
```suggestion
1. Enable `rest_plugin` in `server/config.yaml`:
```

docs/docs/guides/plugins.md
```diff
@@ -113,4 +113,9 @@ class ExamplePolicy(ApplyPolicy):
 
 </div>
 
+## Built-in Plugins
+
+### REST Plugin
+If you'd like to apply custom policies within your organization, you can set up your own plugin API server and integrate it with `dstack` via the `rest-plugin`. To get started, check out the [plugijn server example](/examples/plugins/example_plugin_server/README.md).
```
r4victor:
```suggestion
`rest_plugin` is a builtin `dstack` plugin that allows writing your custom plugins as API servers, so you don't need to install plugins as Python packages.

Plugins implemented as API servers have advantages over plugins implemented as Python packages in some cases:
* No dependency conflicts with `dstack`.
* You can use any programming language.
* If you run the `dstack` server via Docker, you don't need to extend the `dstack` server image with plugins or map them via volumes.

To get started, check out the [plugin server example](/examples/plugins/example_plugin_server/README.md).
```

r4victor:
And regarding the documentation, I think we can generate the OpenAPI schema reference from the example plugin server and put it in the Reference->Plugins->Rest plugin similar to how the server OpenAPI is generated (https://github.com/dstackai/dstack/blob/master/scripts/docs/gen_openapi_reference.py). This could probably be done in a separate PR to limit the scope.

src/dstack/plugins/builtin/rest_plugin.py
Nadine-H: @r4victor I'm interested to hear your thoughts on having `dstack` as a dependency in the plugin server. An alternative would be to guide users to generate the pydantic models in their servers from the openapi spec file, using something like [`datamodel-code-generator`](https://github.com/koxudaxi/datamodel-code-generator). If we go with this route, then I'll move the rest_plugin and its models to `_internal/` as you suggested.

Nadine-H: If I'm not mistaken, the openapi specs are not published to a repo that users can pull from, so with the current setup they'd need to generate the openapi spec file using the script that I will add later. Perhaps importing the models from the `dstack` package isn't a bad idea (at least for now).

r4victor: I think exposing Python models for the plugin server makes perfect sense, so as I mentioned in my last comment, `src/dstack/plugins/builtin/rest_plugin/` should be appropriate.

src/dstack/plugins/builtin/rest_plugin/_models.py
```diff
@@ -0,0 +1,40 @@
...
+
+RunSpecRequest = SpecApplyRequest[RunSpec]
+FleetSpecRequest = SpecApplyRequest[FleetSpec]
+VolumeSpecRequest = SpecApplyRequest[VolumeSpec]
+GatewaySpecRequest = SpecApplyRequest[GatewaySpec]
+
+
+class SpecApplyResponse(BaseModel, Generic[SpecType]):
+    spec: SpecType
+    error: str | None = None
```
dakota-shop: Should these fields be documented with a brief description of what they are and their semantics (e.g. what it means if `error` is populated)? (here and in the request object)

src/dstack/plugins/builtin/rest_plugin/_plugin.py
```diff
@@ -0,0 +1,118 @@
...
+    SpecApplyRequest,
+    SpecApplyResponse,
+    VolumeSpecRequest,
+    VolumeSpecResponse,
+)
+
+logger = get_plugin_logger(__name__)
+
+PLUGIN_SERVICE_URI_ENV_VAR_NAME = "DSTACK_PLUGIN_SERVICE_URI"
+PLUGIN_REQUEST_TIMEOUT = 8  # in seconds
```
dakota-shop:
```suggestion
PLUGIN_REQUEST_TIMEOUT_SEC = 8  # in seconds
```

Not sure if this would need to be updated elsewhere

src/dstack/plugins/builtin/rest_plugin/_plugin.py
```diff
@@ -0,0 +1,118 @@
...
+        self._plugin_service_uri = os.getenv(PLUGIN_SERVICE_URI_ENV_VAR_NAME)
+        logger.info(f"Found plugin service at {self._plugin_service_uri}")
+        if not self._plugin_service_uri:
+            logger.error(
+                f"Cannot create policy because {PLUGIN_SERVICE_URI_ENV_VAR_NAME} is not set"
+            )
+            raise ServerClientError(f"{PLUGIN_SERVICE_URI_ENV_VAR_NAME} is not set")
+
+    def _check_request_rejected(self, response: SpecApplyResponse):
+        if response.error is not None:
```
dakota-shop, line 43: Theoretically `error: ""` would cause a rejection; is that intended?

src/dstack/plugins/builtin/rest_plugin/_plugin.py
```diff
@@ -0,0 +1,118 @@
...
+                f"Could not connect to plugin service at {self._plugin_service_uri}: %s", e
+            )
+            raise ServerClientError(
+                f"Could not connect to plugin service at {self._plugin_service_uri}"
+            )
+        except requests.RequestException as e:
+            logger.error("Request to the plugin service failed: %s", e)
+            raise ServerClientError("Request to the plugin service failed")
+
+    def _on_apply(self, request_cls, response_cls, endpoint, user, project, spec):
```
r4victor: Type annotations?

r4victor (approved):
LGTM. I see some unresolved comments so if you're willing to address them, please do. I'm happy to merge once CI is passed.

src/dstack/plugins/builtin/rest_plugin/_models.py
r4victor: As shown by the CI, `str | None` is not supported for Python 3.9 which we target so please use Optional.

src/dstack/plugins/builtin/rest_plugin/_plugin.py
Nadine-H: It shouldn't, I'll fix it and add additional tests for this.

Nadine-H: @r4victor should we not allow empty strings and cause a pydantic ValidationError? or just treat it as a "no error" case?

Nadine-H: @r4victor I implemented the first option so that plugin developers are aware if they mistakenly set the error to `""`, but happy to switch to the other option if you prefer.

src/tests/plugins/test_rest_plugin.py
r4victor, line 1: Modules outside `dstack._internal.server` do not have access to server fixtures such as test_db. I'm ok with importing the required fixtures to test_rest_plugin.py. That should be the simplest solution.

Nadine-H: Done in https://github.com/dstackai/dstack/pull/2631/commits/6f1f90ec00d12963ff524eb6a9772eb4c3a44108

Issue #2642, opened 2025-05-15
Check port forwarding before switching SSH instance from `pending` to `idle`


jvstme:
### Problem

If an SSH instance forbids port forwarding with `AllowTcpForwarding no`, `dstack` does not detect this configuration error at `dstack apply` time. The instance is marked as `idle` and `dstack apply` exits successfully. The instance is then marked as `unreachable` immediately after that.

This is counterintuitive and the error can remain unnoticed by the user.

### Solution

Check port forwarding before marking the instance as `idle`. For example, replace [this health check](https://github.com/dstackai/dstack/blob/10ed8e78fc86d5b4d28f5cd1b51dd6749f8283bf/src/dstack/_internal/core/backends/remote/provisioning.py#L223) that runs curl on the instance with the [regular health check](https://github.com/dstackai/dstack/blob/10ed8e78fc86d5b4d28f5cd1b51dd6749f8283bf/src/dstack/_internal/server/background/tasks/process_instances.py#L869) that uses port forwarding and calls shim from the server.

### Would you like to help us implement this feature by sending a PR?

Yes

Issue #2640, opened 2025-05-15
[Bug]: shim server should only be bound to localhost interface


colinjc:
### Steps to reproduce

When using ssh instances we can curl the shim server from a public ip

### dstack version

0.19.8

### Server logs

```shell

```

jvstme:
SSH fleet instances should have a firewall to protect the shim port and any other ports that can be bound to by users' runs (most runs can't bind to ports on the host, but distributed tasks can). Only SSH traffic should be allowed from external networks. I will mention this in the docs.

As an extra security measure, however, we should still bind shim to localhost only.

PR #2643, opened 2025-05-15
[Docs]: Mention SSH fleet networking requirements

jvstme:
#2640
#2642

### DIFF
--- a/docs/docs/concepts/fleets.md
+++ b/docs/docs/concepts/fleets.md
@@ -254,30 +254,30 @@ Define a fleet configuration as a YAML file in your project directory. The file 
 </div>
 
 ??? info "Requirements" 
-    1.&nbsp;Hosts should be pre-installed with Docker.
+    1.&nbsp;Hosts must be pre-installed with Docker.
 
     === "NVIDIA"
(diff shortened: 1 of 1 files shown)


docs/docs/concepts/fleets.md
```diff
@@ -275,6 +275,10 @@ Define a fleet configuration as a YAML file in your project directory. The file
 
     3.&nbsp;The user specified should have passwordless `sudo` access.
 
+    4.&nbsp;The SSH server should be running and configured with `AllowTcpForwarding yes` in `/etc/ssh/sshd_config`.
+
+    5.&nbsp;The firewall should allow SSH and forbid any other connections from external networks.
```
un-def: I think we have some ambiguity in this whole block.
* Almost all “should” must actually be “must” (as defined in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119)).
* The last requirement is “The firewall _must_ allow SSH and _should_ forbid any other connections from external networks” (the second part is for security reasons, it's not a strict requirement, right?).

I would also add that, as most communication libraries use ports from the ephemeral range, in cluster fleets the firewall must allow the internal network connections in that range (the easiest way is to allow all internal connections).

un-def (approved):

PR #2649, opened 2025-05-16
Replace `conda` with `uv` in `dstackai/base` images

### LINKED ISSUE #2625: [UX]: Replace `conda` with `uv` in `dstack`'s default Docker image
### Problem

Currently, `dstack`'s default Docker image (which is very convenient to use) uses Conda as the default package manager. 
Apparently, `uv` provides a significantly better experience in both performance and compatibility compared to `conda`.
Moreover, `uv` is quickly becoming the most common and user-friendly way to install packages.

Currently, the process of installing and activating `uv` manually with the default image is unnecessarily cumbersome.

### Solution

Let's ditch Miniforge's `conda` in favor of `uv`.

Note, we could also ditch our Docker images built specifically for each major version of Python, and instead, let `uv` install any version of Python that `uv` supports.

### Workaround

_No response_

No

un-def:
Closes: https://github.com/dstackai/dstack/issues/2625

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -1,14 +1,20 @@
 FROM nvidia/cuda:12.1.0-base-ubuntu20.04
 
 ARG PYTHON
-ENV PYTHON=$PYTHON
+ARG _UV_HOME="/opt/uv"
+ENV UV_PYTHON="${PYTHON}"
+ENV UV_INSTALL_DIR="${_UV_HOME}/bin"
+ENV UV_PYTHON_INSTALL_DIR="${_UV_HOME}/python"
+ENV UV_PYTHON_BIN_DIR="${UV_PYTHON_INSTALL_DIR}/bin"
+ENV UV_MANAGED_PYTHON=1
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
-ENV PIP_ROOT_USER_ACTION=ignore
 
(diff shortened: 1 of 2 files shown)


docker/base/Dockerfile
```diff
@@ -1,30 +1,26 @@
...
-    chmod +x /opt/conda/etc/profile.d/conda.sh && \
-    echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc && \
-    /opt/conda/condabin/conda update conda --all -y && \
-    /opt/conda/condabin/conda create --name workflow python=${PYTHON} -y && \
-    /opt/conda/condabin/conda config --prepend channels "nvidia/label/cuda-12.1.0" && \
-    /opt/conda/condabin/conda config --set always_yes true && \
-    /opt/conda/condabin/conda clean --all && \
-    echo "conda activate workflow" >> ~/.bashrc
+RUN curl -LsSf https://astral.sh/uv/install.sh | INSTALLER_NO_MODIFY_PATH=1 sh && \
+    uv python install --preview --default
```
jvstme, line 26: What does `--preview` do?

jvstme (approved):

docker/base/Dockerfile
un-def:
```
$ uv python install 3.10 --default
The `--default` flag is only available in preview mode; add the `--preview` flag to use `--default`
```

un-def: https://docs.astral.sh/uv/reference/settings/#preview

> Whether to enable experimental, preview features.

PR #2652, opened 2025-05-16
[UX]: Replace conda with uv in dstack's default Docker image #2625

peterschmidt85:
* Updated CUDA to 12.1.1
* Use NVIDIA's devel image


jvstme (approved):
Since the two Dockerfiles now only differ in the `FROM` clause, I'd suggest to keep just one Dockerfile and parametrize it. Something like that should work:
```dockerfile
ARG FLAVOR
FROM nvidia/cuda:12.1.1-$FLAVOR-ubuntu20.04
```
This will allow to avoid duplication and related errors, such as forgetting to add stuff to one of the images.

.github/workflows/docker.yml
```diff
@@ -60,8 +60,8 @@ jobs:
         uses: docker/setup-qemu-action@v3
       - name: Build and upload to DockerHub
         run: |
-          docker buildx build --platform linux/amd64 --build-arg PYTHON=${{ matrix.python }} --push --provenance=false --tag dstackai/base:py${{ matrix.python }}-${{ inputs.image_version }}-cuda-12.1 -f base/Dockerfile .
-          docker buildx build --platform linux/amd64 --build-arg PYTHON=${{ matrix.python }} --build-arg VERSION=${{ inputs.image_version }} --push --provenance=false --tag dstackai/base:py${{ matrix.python }}-${{ inputs.image_version }}-cuda-12.1-devel -f base/devel.Dockerfile .
+          docker buildx build --platform linux/amd64 --build-arg FLAVOR=base --build-arg PYTHON=${{ matrix.python }} --push --provenance=false --tag dstackai/base:py${{ matrix.python }}-${{ inputs.image_version }}-cuda-12.1 -f base/Dockerfile .
+          docker buildx build --platform linux/amd64 --build-arg FLAVOR=devel --build-arg PYTHON=${{ matrix.python }} --build-arg VERSION=${{ inputs.image_version }} --push --provenance=false --tag dstackai/base:py${{ matrix.python }}-${{ inputs.image_version }}-cuda-12.1-devel -f base/devel.Dockerfile .
```
jvstme, line 64: Also drop the deleted Dockerfile from the workflow, otherwise the workflow won't work
```suggestion
          docker buildx build --platform linux/amd64 --build-arg FLAVOR=devel --build-arg PYTHON=${{ matrix.python }} --build-arg VERSION=${{ inputs.image_version }} --push --provenance=false --tag dstackai/base:py${{ matrix.python }}-${{ inputs.image_version }}-cuda-12.1-devel -f base/Dockerfile .
```

jvstme, line 64: Since the devel image no longer depends on the base image, it also makes sense to move `FLAVOR` to the job matrix (line 47), so that base and devel images can be built in parallel

peterschmidt85: Done via #2659

peterschmidt85: Done via #2659

PR #2654, opened 2025-05-17
[UX] Support `dstack ps -n NUM`

peterschmidt85:


src/dstack/api/_public/runs.py
```diff
@@ -691,28 +691,31 @@ def exec_plan(
...
             repo_id=None,
             only_active=only_active,
+            limit=limit or 100,
         )
         if only_active and len(runs) == 0:
             runs = self._api_client.runs.list(
                 project_name=self._project,
                 repo_id=None,
+                limit=limit or 100,
             )[:1]
```
jvstme: (nit) Since we only need the last run here, we can always pass `limit=1`. This may speed up the response a bit
```suggestion
            runs = self._api_client.runs.list(
                project_name=self._project,
                repo_id=None,
                limit=1,
            )
```

src/dstack/_internal/cli/commands/ps.py
```diff
@@ -36,10 +36,17 @@ def _register(self):
             help="Watch statuses of runs in realtime",
             action="store_true",
         )
+        self._parser.add_argument(
+            "-n",
+            "--last",
+            help="Show only the last N runs",
```
jvstme: (nit) It is not very intuitive that `-n` also enables displaying stopped runs, so let's at least document it.
```suggestion
            help="Show only the last N runs. Implies --all",
```

jvstme (approved):

src/dstack/api/_public/runs.py
peterschmidt85: @jvstme Active runs can be many

jvstme: This block is used for loading one stopped run if there are no active runs. It already sets the limit to one run, expect it does so after receiving the response - `[:1]`

PR #2666, opened 2025-05-20
Auto activate uv venv with pip installed

r4victor:
Addresses #2625 

#2649 ditched conda in favor of uv in dstack Docker images. The side effect of that is that `pip` is no longer available in runs. This breaks things directly dependent on pip such as Jupyter's `%pip`, `python -m venv`, and `pip` hardcoded in makefiles and other scripts. The PR makes pip available by creating a uv venv with pip installed and activating it when using dstack Docker images. This should not worsen UX for users relying on uv-based workflow, only improve UX for those who depend on pip directly.

The PR will be merged once the 0.8 Docker images with uv are built and version.base_image is updated.


src/dstack/_internal/server/services/jobs/configurators/base.py
```diff
@@ -164,6 +166,18 @@ async def _commands(self) -> List[str]:
...
         return result
 
+    def _dstack_image_commands(self) -> List[str]:
+        if (
+            self.run_spec.configuration.image is not None
+            or self.run_spec.configuration.entrypoint is not None
+        ):
+            return []
+        return [
+            f"uv venv --prompt workflow --seed {DEFAULT_REPO_DIR}/.venv > /dev/null 2>&1",
```
jvstme, line 176: I assume this `.venv` can turn up in `git status`, as it is not necessarily included in `.gitignore`. Any chance we could create it outside of the repo?

src/dstack/_internal/server/services/jobs/configurators/base.py
```diff
@@ -164,6 +166,18 @@ async def _commands(self) -> List[str]:
 
         return result
 
+    def _dstack_image_commands(self) -> List[str]:
+        if (
+            self.run_spec.configuration.image is not None
```
jvstme, line 171: (nit) This means pip won't work if our base image is explicitly specified in `image`. I've seen some users do it, possibly because we advertise it in the docs ([1](https://dstack.ai/docs/quickstart/#apply-a-configuration), [2](https://dstack.ai/docs/concepts/tasks/#docker), and more).

Maybe we could set up the venv in the Dockerfile, so that our images already come with a working pip?

src/dstack/_internal/server/services/jobs/configurators/base.py
r4victor: I considered creating a global uv venv outside the repo. The problem with that is that having an active venv outside the repo would trigger uv warnings about the active repo not being used cause it defaults to `.venv` in the project directory. Not critical, but quite confusing. And since we want to support uv-first flow, I think we should optimize for this case.

As regards to .venv in git, it will never appear in git since it's always excluded by .venv/.gitignore.

src/dstack/_internal/server/services/jobs/configurators/base.py
r4victor: The current idea is to have .venv in the repo dir so that uv does not issue warnings about the wrong venv. And I don't want to hardcode the repo path in the image since we had plans to make it configurable.

src/dstack/_internal/server/services/jobs/configurators/base.py
jvstme: Users may still be confused by a new directory in their repository, especially if they don't use uv or Python. But I guess we can live with that if there are no good alternatives

jvstme (approved):

src/dstack/_internal/server/services/jobs/configurators/base.py
r4victor: Maybe, let's see how it works

r4victor:
Tested pip vs uv for installing pytorch on GCP n1-standard-4:

```
# time pip install torch
...
real    2m26.338s
user    1m37.514s
sys     0m16.711s
```

```
# time uv pip install torch
...
real    0m32.771s
user    0m29.070s
sys     0m8.300s
```

PR #2667, opened 2025-05-20
[UX] Minor improvements of `dstack metrics`

### LINKED ISSUE #2316: [UX]: Active and finished jobs mixed in `dstack stats`
### Steps to reproduce

1. Run a service with 2 replicas.
   ```yaml
   type: service
   name: nginx
   image: nginx
   port: 80
   replicas: 2
   ```
1. Scale it down to 1 replica.
   ```yaml
   type: service
   name: nginx
   image: nginx
   port: 80
   replicas: 1
   ```
1. See `dstack stats`.

### Actual behaviour

Shows two jobs, even though the second one is no longer running.
```shell
> dstack ps
 NAME               BACKEND          RESOURCES                         PRICE    STATUS      SUBMITTED 
 nginx                                                                          running     1 min ago 
   replica=0 job=0  aws (us-west-2)  1xCPU, 2GB, 100.0GB (disk), SPOT  $0.0056  running     1 min ago 
   replica=1 job=0  aws (us-west-2)  1xCPU, 2GB, 100.0GB (disk), SPOT  $0.0056  terminated  1 min ago

> dstack stats nginx
 NAME               CPU  MEMORY       GPU 
 nginx                                    
   replica=0 job=0  0%   23MB/2048MB  -   
   replica=1 job=0  0%   20MB/2048MB  -
```

### Expected behaviour

Either:
- Only show running jobs.
- Show job statuses.

### dstack version

0.18.42

### Server logs

```shell

```

### Additional information

_No response_

peterschmidt85:
* Show run/job status
* Better-format GPU metrics
* Allow finished runs
* Show GB (instead of MB)

Fixes #2316

### DIFF
--- a/src/dstack/_internal/cli/commands/metrics.py
+++ b/src/dstack/_internal/cli/commands/metrics.py
@@ -113,7 +113,7 @@ def _get_metrics_table(run: Run, metrics: List[JobMetrics]) -> Table:
                     gpu_metrics += f"gpu={i} mem={_format_memory(gpu_memory_usage, 2)}"
                     if resources is not None:
                         gpu_metrics += (
-                            f"/{_format_memory(resources.gpus[i].memory_mib * 1024 * 1024, 0)}"
+                            f"/{_format_memory(resources.gpus[i].memory_mib * 1024 * 1024, 2)}"
                         )
                     gpu_metrics += f" util={gpu_util_percent}%"
 
--- a/src/dstack/_internal/cli/commands/metrics.py
+++ b/src/dstack/_internal/cli/commands/metrics.py
@@ -100,7 +100,7 @@ def _get_metrics_table(run: Run, metrics: List[JobMetrics]) -> Table:
         if memory_usage is not None:
             memory_usage = _format_memory(memory_usage, 2)
             if resources is not None:
-                memory_usage += f"/{_format_memory(resources.memory_mib * 1024 * 1024, 0)}"
+                memory_usage += f"/{_format_memory(resources.memory_mib * 1024 * 1024, 2)}"
         gpu_metrics = ""
         gpus_detected_num = _get_metric_value(job_metrics, "gpus_detected_num")
         if gpus_detected_num is not None:
--- a/src/dstack/_internal/cli/commands/metrics.py
+++ b/src/dstack/_internal/cli/commands/metrics.py
@@ -39,8 +39,6 @@ class MetricsCommand(APIBaseCommand):
         run = self.api.runs.get(run_name=args.run_name)
         if run is None:
             raise CLIError(f"Run {args.run_name} not found")
-        if run.status.is_finished():
-            raise CLIError(f"Run {args.run_name} is finished")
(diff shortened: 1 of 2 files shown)


jvstme (approved):
> Show GB (instead of MB)

I don't think gigabyte granularity is enough.

- It makes metrics useless for low-memory instances. Many clouds provide instances with only 1GB of RAM, some even provide 0.5GB RAM, including AWS. Such instances require precise memory consumption monitoring.
- It makes it more difficult to see metric changes with `dstack metrics -w`. A metric can be growing or falling slowly but steady, which can be easily noticeable with megabyte, but not gigabyte granularity.

I suggest showing gigabytes with 2-3 digits after the decimal point, or reverting to megabyte granularity for both RAM and VRAM.

src/dstack/_internal/cli/commands/metrics.py
```diff
@@ -100,9 +98,9 @@ def _get_metrics_table(run: Run, metrics: List[JobMetrics]) -> Table:
             cpu_usage = f"{cpu_usage:.0f}%"
         memory_usage = _get_metric_value(job_metrics, "memory_working_set_bytes")
         if memory_usage is not None:
-            memory_usage = f"{round(memory_usage / 1024 / 1024 / 1024)}GB"
+            memory_usage = _format_memory(memory_usage, 2)
             if resources is not None:
-                memory_usage += f"/{round(resources.memory_mib / 1024)}GB"
+                memory_usage += f"/{_format_memory(resources.memory_mib * 1024 * 1024, 0)}"
```
jvstme: (nit) I don't think we need to limit the total memory to an integer, as this can result in significant loss of precision — for example, 1.5 GB could be rounded to 2 GB. Instead, just allow up to two decimal digits. If they are zeros, which is the most common case, they will be trimmed in `_format_memory` either way.
```suggestion
                memory_usage += f"/{_format_memory(resources.memory_mib * 1024 * 1024, 2)}"
```

src/dstack/_internal/cli/commands/metrics.py
```diff
@@ -112,9 +110,11 @@ def _get_metrics_table(run: Run, metrics: List[JobMetrics]) -> Table:
                 if gpu_memory_usage is not None:
                     if i != 0:
                         gpu_metrics += "\n"
-                    gpu_metrics += f"gpu={i} mem={round(gpu_memory_usage / 1024 / 1024 / 1024)}GB"
+                    gpu_metrics += f"gpu={i} mem={_format_memory(gpu_memory_usage, 2)}"
                     if resources is not None:
-                        gpu_metrics += f"/{round(resources.gpus[i].memory_mib / 1024)}GB"
+                        gpu_metrics += (
+                            f"/{_format_memory(resources.gpus[i].memory_mib * 1024 * 1024, 0)}"
```
jvstme: (nit) Same as above
```suggestion
                            f"/{_format_memory(resources.gpus[i].memory_mib * 1024 * 1024, 2)}"
```

jvstme (approved):

PR #2673, opened 2025-05-20
Make local code upload size limit configurable

### LINKED ISSUE #2663: [Feature]: Make local code upload size limit configurable
### Problem

Right now there is a hardcoded limit of 2M for repo diffs which also applies to the total size of the working directory when `--local` is provided to `dstack apply`. This should be configurable by operators running their own instance of the dstack server.

### Solution

I can imagine a server-level or project-level configuration value that sets this, retaining the current default value of 2M if unset. I suspect this makes more sense as a server-level configuration item. There might be a possibility that different Fleets might be more/less sensitive to large code uploads, but I'm not sure it would be worth adding that granularity.

### Workaround

No

Yes

colinjc:
Resolves https://github.com/dstackai/dstack/issues/2663

Adds an env variable which can be used to configure the max code upload size, or disable it. Also improves error message when hitting limit by including size of diff.


src/dstack/_internal/server/settings.py
```diff
@@ -85,6 +85,7 @@
 USER_PROJECT_DEFAULT_QUOTA = int(os.getenv("DSTACK_USER_PROJECT_DEFAULT_QUOTA", 10))
 FORBID_SERVICES_WITHOUT_GATEWAY = os.getenv("DSTACK_FORBID_SERVICES_WITHOUT_GATEWAY") is not None
 
+SERVER_CODE_UPLOAD_LIMIT = int(os.getenv("DSTACK_SERVER_CODE_UPLOAD_LIMIT", 2 * 2**20))
```
dakota-shop, line 88: Would it be easy to implement this as a friendly string like `2M` or `1G`?

r4victor: LGTM, so merging it now. Feel free to open a new PR if parsing human-readable strings is needed.

r4victor:
Thanks!

src/dstack/_internal/server/settings.py
r4victor: @dakota-shop, other environment variables like `DSTACK_DEFAULT_SERVICE_CLIENT_MAX_BODY_SIZE` do not support human-readable strings as well, so probably makes sense to add the parsing to all of them if at all.

PR #2683, opened 2025-05-22
Update dstack-proxy contributing guide

### LINKED ISSUE #2252: Update the gateway contributing guide
The contributing guides need to be updated to describe the dstack-proxy architecture and other changes to gateways introduced in #1595.

jvstme:
Closes #2252


contributing/PROXY.md
jvstme, line 1: [Rendered](https://github.com/dstackai/dstack/blob/issue_2252_proxy_contributing_guide/contributing/PROXY.md)

contributing/PROXY.md
```diff
@@ -0,0 +1,194 @@
...
+   ```
+
+1. Connect to the gateway:
+
+   ```shell
+   chmod 600 /path/to/the/gateway/key
+   ssh -i /path/to/the/gateway/key ubuntu@gateway.example
+   ```
+
+1. Prepare an environment with your development branch on the gateway:
```
r4victor, line 192: I wish the recommended approach wouldn't involve pushing changes to a remote branch. Previously, the guide suggested ssh/scp. Not ideal, but do you think this one is better?

r4victor (approved):

contributing/PROXY.md
jvstme: The previous guide suggested building and uploading a wheel. I think running from source is preferable, as it allows to iterate faster, run a debugger, and possibly edit the code on the gateway.

I've tried using rsync to upload the source code, but it's difficult to set up to upload a correct copy of the repo without uploading unnecessary files, like the local venv, and without deleting the necessary remote files, like the remote venv.

But I can include an example on how to push the code via git without the GitHub repo.

PR #2687, opened 2025-05-23
Update gitignore implementation to support more cases

colinjc:
Existing implementation was missing support for a ** which was pretty critical. This updates the logic to use an existing library which already supports **.


src/dstack/_internal/core/models/repos/local.py
```diff
@@ -2,13 +2,17 @@
 from pathlib import Path
 from typing import BinaryIO, Optional
 
+import humanize
```
r4victor: Note that I dropped humanize dependency in #2692 (#2691) since for dstack purposes it can be replaced with a simple function.

Use `from dstack._internal.utils.common import sizeof_fmt` instead.

PR #2689, opened 2025-05-23
Fix runner server not respecting max code upload size

colinjc:


src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -38,6 +38,7 @@
     VolumeProvisioningData,
 )
 from dstack._internal.core.services import is_valid_dstack_resource_name
+from dstack._internal.server.settings import SERVER_CODE_UPLOAD_LIMIT
```
r4victor: backends code must not depend on server modules and cannot import from dstack._internal.server. A solution would be to make the environment variable non server specific (e.g. define DSTACK_CODE_UPLOAD_LIMIT in dstack._internal.settings. Alternatively, we can fix 413 by always setting `http.MaxBytesReader(w, r.Body, math.MaxInt64)`. It shouldn't lead to problems since the runner API is exposed in a trusted environment only and will be effectively limited by the server upload limit anyway.

r4victor: I'm inclined to http.MaxBytesReader(w, r.Body, math.MaxInt64)

PR #2695, opened 2025-05-26
Update gitignore logic to catch more cases

colinjc:
Existing implementation was missing support for a ** which was pretty critical. This updates the logic to use an existing library which already supports **.


pyproject.toml
```diff
@@ -152,19 +145,9 @@ gcp = [
...
-    "dstack[server]",
-]
-lambda = [
-    "boto3>=1.38.13",
-    "botocore",
-    "dstack[server]",
-]
+datacrunch = ["datacrunch", "dstack[server]"]
+kubernetes = ["kubernetes", "dstack[server]"]
+lambda = ["boto3>=1.38.13", "botocore", "dstack[server]"]
```
r4victor: Why formatting pyproject.toml differently? I like dependency per line as it was.

src/dstack/_internal/core/models/repos/local.py
```diff
@@ -75,6 +79,7 @@ def write_code_file(self, fp: BinaryIO) -> str:
                 arcname="",
                 filter=TarIgnore(self.run_repo_data.repo_dir, globs=[".git"]),
             )
+        logger.debug(f"Code file size: {sizeof_fmt(fp.tell())} bytes")
```
r4victor:
```suggestion
        logger.debug(f"Code file size: {sizeof_fmt(fp.tell())}")
```

sizeof_fmt should already emit B/Kib/MiB suffix so "bytes" is not necessary

pyproject.toml
colinjc: My bad, auto formatter hit it. I'll revert that bit

src/dstack/_internal/utils/ignore.py
```diff
@@ -16,77 +17,83 @@ def __init__(
...
+                        prefix = ""
+                    else:
+                        prefix = path.relative_to(self.root_dir)
 
-    def ignore(self, path: PathLike, sep="/") -> bool:
-        if not path:
+                    # Read patterns and prefix them with directory path
+                    with ignore_file.open("r", encoding="utf-8", errors="ignore") as f:
+                        for line in f:
+                            line = line.strip()
```
jvstme: Leading spaces should not be trimmed. Trailing spaces should not be trimmed if preceded by a backslash.
> Trailing spaces are ignored unless they are quoted with backslash ("`\`").
[docs](https://git-scm.com/docs/gitignore#_pattern_format)

I think we can trim the newline and leave the rest to gitignore_parser.
```suggestion
                            line = line.rstrip("\n")
```

src/dstack/_internal/utils/ignore.py
```diff
@@ -16,77 +17,83 @@ def __init__(
...
+                            if line and not line.startswith("#"):
+                                if prefix:
+                                    # Prefix patterns with directory path for subdirectories
+                                    if line.startswith("/"):
+                                        # Absolute pattern within subdirectory
+                                        patterns.append(os.path.join(prefix, line[1:]))
```
jvstme: Slashes in the middle also indicate an absolute pattern.
> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular `.gitignore` file itself.
[docs](https://git-scm.com/docs/gitignore#_pattern_format)

The previous implementation had the same flaw, but I think we can fix it now while we are at it.

```suggestion
                                    if "/" in line.rstrip("/"):
                                        # Absolute pattern within subdirectory
                                        patterns.append(os.path.join(prefix, line.lstrip("/")))
```

src/tests/_internal/utils/test_ignore.py
```diff
@@ -0,0 +1,256 @@
...
+            git_ignore = GitIgnore(test_dir)
+
+            # The ignored directory itself should be ignored
+            assert git_ignore.ignore("ignored_dir") is True
+            assert git_ignore.ignore("ignored_dir/file.txt") is True
+
+            # Patterns from .gitignore files inside ignored directories should not apply
+            # to files outside those directories
+            assert git_ignore.ignore("file.should_not_apply") is False
+            assert git_ignore.ignore("file.also_should_not_apply") is False
```
jvstme: > Test that ignored directories don't have their subdirectories processed

The test doesn't seem to actually test this. The assertions pass because `ignored_dir/.gitignore` and `ignored_dir/subdir/.gitignore` do not apply to files in the root directory, not because `ignored_dir` itself is ignored.

src/dstack/_internal/utils/ignore.py
```diff
@@ -16,77 +17,83 @@ def __init__(
...
+                                    else:
+                                        # Relative pattern within subdirectory
+                                        # Add pattern that matches files directly in the subdirectory
+                                        patterns.append(os.path.join(prefix, line))
+                                        # Add pattern that matches files in deeper subdirectories
+                                        patterns.append(os.path.join(prefix, "**", line))
```
jvstme: I assume pattern negation with the `!` prefix won't work in subdirectories. Should be easy to fix, although it's not strictly necessary to fix it in this PR, since the old implementation didn't work either

src/dstack/_internal/utils/ignore.py
jvstme, line 1: This implementation seems to impact performance. For example, packing the dstack repo now takes seven seconds on my machine, up from two.

```python
>>> from dstack._internal.core.models.repos.local import *
>>> r = LocalRepo(repo_dir="/path/to/dstack/repo")
>>> f = open("/tmp/test.tar", "wb+")
>>> r.write_code_file(f)
```

Overall, I think this is acceptable since it happens on the client, but it may be worth looking into

colinjc: That's brutal for the cli, I'll swap out to a faster library instead of just using the one with the most stars 😓

colinjc: Switched to https://github.com/borsattoz/ignore-python

jvstme: It's faster, thanks

jvstme (approved):
@colinjc, thanks for the PR. Replacing our gitignore implementation with a more mature one is a good idea.

The latest commit was building malformed unpackable archives, but I fixed it with a couple of edits. Will merge now, thank you

PR #2696, opened 2025-05-26
Generate REST plugin API docs

Nadine-H:
Follow up to https://github.com/dstackai/dstack/pull/2631


r4victor:
@Nadine-H, I made a few fixes for the openapi schema to be generated in the CI. I also made example_plugin_server an installable package by defining a build system so that we can install the package instead of importing from examples and setting PYTHONPATH.

r4victor:
I also tested the example REST plugin and all looks good.

Issue #2467, opened 2025-04-03
[UX] Simplify the use of `mpirun`


peterschmidt85:
Currently, we have to write custom logic inside `commands` to make mpirun work (e.g., waiting for all nodes, FIFO, etc.). The ideal solution would be for users to avoid writing custom code inside commands.

One possible solution is to support `mpi: true`. While users would still be able to use custom `commands`, those who prefer not to would benefit from this new property.

r4victor:
@peterschmidt85, I agree we should simplify running mpi / nccl tests but I'd suggest supporting more generic properties. For example, we could add `startup_order: master-first/workers-first` and `termination_criteria: all-done/master-done`. So for mpi, you would set 

```yaml
startup_order: workers-first
termination_criteria: master-done
```

This can cover other use cases similar to mpi but not mpi necessarily.

Introducing separate properties for each particular use case would lead to lots of top-level properties in the future that would clutter dstack simple interface and wouldn't compose as well. It may not be a problem now, but I would follow a more generic approach since we cannot keep all the potential use cases in mind.

peterschmidt85:
@r4victor Only minor comment on naming:

1. If we can use a term that we used before, I'd rather go with it instead of introducing one more. We tend to use `policy` and `duration` across different settings. So IMO here we could use `startup_policy`. But it's not critical, `startup_order` is also OK.
2. Also, the term `termination` we use in regard to fleet instances, and `stop` in regard to runs/jobs. Perhaps, we could use `stop_policy` instead of `termination_criteria`.

Otherwise, I like the plan. 

Also, a few minor, almost not related, questions:
1. What is our current policy on what happens when a worker node goes down? Do we automatically re-provision a new one? 
2. What is out current police on what happens when the master node goes down? Do we stop the run?

r4victor:
@peterschmidt85 regarding the naming, to me it feels like a misuse of `_policy` suffix to the point that everything is a policy and it doesn't add any meaning to property keys. We already have `retry_policy` and `termination_policy` deprecated, and I'd avoid using word policy unless it adds useful meaning. `spot_policy` and `stop_policy` are likely to be mistaken.

I'm ok with replacing `termination_criteria` with `stop_criteria` or even `stop_on`.

peterschmidt85:
@r4victor Agreed!

r4victor:
[...]

PR #2714, opened 2025-05-30
Implement startup_order and stop_criteria

r4victor:
A part of #2467

The PR introduces new run configuration properties:
* `startup_order` allows specifying the order in which master and workers jobs are started.
* `stop_criteria` allows specifying the criteria determining when a multi-node run should be considered finished.

They simplify running mpirun with `dstack`:

```yaml
type: task
name: nccl-tests

nodes: 2
startup_order: workers-first
stop_criteria: master-done

image: dstackai/efa
commands:
  - |
    if [ ${DSTACK_NODE_RANK} -eq 0 ]; then
      cd /root/nccl-tests/build
      : > hostfile
      for ip in ${DSTACK_NODES_IPS}; do
        echo "${ip} slots=${DSTACK_GPUS_PER_NODE}" >> hostfile
      done
      MPIRUN='mpirun --allow-run-as-root --hostfile hostfile'
      # Run NCCL Tests
      ${MPIRUN} \
        -n ${DSTACK_GPUS_NUM} -N ${DSTACK_GPUS_PER_NODE} \
        --mca btl_tcp_if_exclude lo,docker0 \
        --bind-to none \
        ./all_reduce_perf -b 8 -e 8G -f 2 -g 1
    else
      sleep infinity
    fi

```

Other multi-node tasks such as iperf may require `startup_order: master-first`. Most such as pytorch will work with the default `startup_order: any`.

TODO:
* use the new properties in the NCCL tests examples


peterschmidt85:
Do you plan to also support `DSTACK_MPI_HOSTFILE`? I guess we just need to mount this file and pass the environment variable on container start.

r4victor:
@peterschmidt85 in a separate PR

r4victor:
and then we can update the NCCL tests example

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -313,6 +313,13 @@ async def _process_active_run(session: AsyncSession, run_model: RunModel):
             termination_reason = RunTerminationReason.RETRY_LIMIT_EXCEEDED
         else:
             raise ValueError(f"Unexpected termination reason {run_termination_reasons}")
+    elif _should_stop_on_master_done(run):
+        logger.info(
+            "%s: run is done because master job is done",
+            fmt(run_model),
+        )
```
jvstme: (nit) The extra log message seems inconsistent with other status changes. The status change is already logged on line 370

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -313,6 +313,13 @@ async def _process_active_run(session: AsyncSession, run_model: RunModel):
...
+        new_status = RunStatus.TERMINATING
+        termination_reason = RunTerminationReason.ALL_JOBS_DONE
```
jvstme, line 319: (nit) `RunTerminationReason.MASTER_DONE`? `ALL_JOBS_DONE` can be misleading here

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -434,3 +441,12 @@ def _can_retry_single_job(run_spec: RunSpec) -> bool:
     # We could make partial retry in some multi-node cases.
     # E.g. restarting a worker node, independent jobs.
     return False
+
+
+def _should_stop_on_master_done(run: Run) -> bool:
+    if run.run_spec.merged_profile.stop_criteria != StopCriteria.MASTER_DONE:
+        return False
+    for job in run.jobs:
+        if job.job_spec.job_num == 0 and job.job_submissions[-1].status == JobStatus.DONE:
```
jvstme, line 447: (nit) Can also check for `termination_reason == JobTerminationReason.DONE_BY_RUNNER` to terminate the run faster, without waiting for the `terminating` -> `done` master job transition. See line 241

src/dstack/_internal/core/models/profiles.py
```diff
@@ -32,6 +33,17 @@ class TerminationPolicy(str, Enum):
     DESTROY_AFTER_IDLE = "destroy-after-idle"
 
 
+class StartupOrder(str, Enum):
+    ANY = "any"
+    MASTER_FIRST = "master-first"
+    WORKERS_FIRST = "workers-first"
```
jvstme, line 39: (nit) I'm not sure about calling non-master nodes "workers", because the master node is also a "worker" - it performs the same work other nodes do.

I can suggest to use "secondary" (`secondary-first`) or avoid any names (`master-last`). Although we might still need a name to use in the code

jvstme (approved):

src/dstack/_internal/server/background/tasks/process_runs.py
r4victor: Good point but I don't think we need to introduce separate statuses besides DONE since stop_criteria determines the reason. It'll be sufficient to show ALL_JOBS_DONE appropriately, and I believe #2716 will change it.

src/dstack/_internal/core/models/profiles.py
r4victor: master/worker is standard terminalogy used for pytorch, mpi, etc. Let's not reinvent.

src/dstack/_internal/server/background/tasks/process_runs.py
r4victor: I'm not sure if we want to terminate the run before the master is really done.

PR #2715, opened 2025-05-30
Update TRL Single Node example to uv

Bihan:


examples/single-node-training/trl/README.md
```diff
@@ -21,19 +21,19 @@ env:
...
+  - uv pip install torch==2.6.0
+  - uv pip install transformers bitsandbytes peft wandb
+  - uv pip install flash_attn --no-build-isolation
   - git clone https://github.com/huggingface/trl
   - cd trl
-  - pip install .
-  - | 
+  - uv pip install .
+  - |
     accelerate launch \
```
peterschmidt85, line 33: is just `accelerate` OK? is it required to have `uv run accelerate`?

peterschmidt85 (approved):

examples/single-node-training/trl/README.md
Bihan: @peterschmidt85 Just accelerate is ok because command is executed inside `venv`. I have tested it.

PR #2737, opened 2025-06-03
Support file upload to gcs bucket

colinjc:
Add support for uploading file blobs to GCS.


src/dstack/_internal/server/services/storage/gcs.py
```diff
@@ -0,0 +1,45 @@
+from typing import Optional
+
+from google.cloud.exceptions import NotFound
```
r4victor: Should also be inside `try`/`except`.

src/dstack/_internal/server/services/storage/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+    elif settings.SERVER_GCS_BUCKET:
+        if not GCS_AVAILABLE:
+            raise ValueError("GCS dependencies are not installed")
+        _default_storage = GCSStorage(
+            bucket=settings.SERVER_GCS_BUCKET,
+        )
+
+
+def get_default_storage() -> Optional[S3Storage]:
```
r4victor: Should return `Optional[BaseStorage]`

src/dstack/_internal/server/services/storage/base.py
```diff
@@ -0,0 +1,25 @@
...
+    ):
+        raise NotImplementedError
+
+    def get_code(
+        self,
+        project_id: str,
+        repo_id: str,
+        code_hash: str,
+    ) -> Optional[bytes]:
+        raise NotImplementedError
```
r4victor: Required method should be decorated with `@abstractmethod` and have empty body (`pass` or `...`) instead of `raise NotImplementedError`

r4victor (changes requested):
Looks good overall

r4victor (approved):

r4victor:
@colinjc, thanks!

peterschmidt85:
@colinjc Quick question, have you tested the PR? It works? We haven't tested it on our side.

colinjc:
I tested it against GCS, but not S3

Issue #2742, opened 2025-06-04
[Feature]: Self-Serve Join/Leave and Better Membership Management


haydnli-shopify:
This is a extension on the ticket we have created earlier https://github.com/dstackai/dstack/issues/2670

### Problem

Right now, Dstack projects are private by default. Only project admins or owners can see them, and new users have to be manually added before they can do anything. This creates a bottleneck—especially in larger orgs (like Shopify), where we want folks to be able to discover and join most projects on their own without needing someone to manually invite them.

This feature is something we're also tracking internally at Shopify as part of our efforts to get Dstack ready for GA.

### Solution

Here’s what we’d like to see:

- Add Project property to indicate private (default) or public:
Project owners or admins should be able to mark a project as "public." Public projects would be visible to everyone in the organization once they log in.
 
- Let users join/leave public projects:
If a project is public, any user can join it with a "Join Project" button (either from the list or the project detail view). Similarly, they should be able to leave with a "Leave Project" button.

A few rules:
     Owners and admins can’t leave their own projects.
     Join/Leave buttons should be context-aware and disabled when the user is not allowed to perform the action (e.g., project creators or admins trying to leave, or users trying to join a project they’re already in).

- Visibility rules:
Public projects appear in the Projects list for everyone.
Private projects stay visible only to members, owners, and admins.

- Restrict access to project setup details
 Even for public projects, CLI setup should only be shown once a user joins.

### Workaround

There really isn’t one. Managing project members has to be done manually by a project admin right now.

### Would you like to help us implement this feature by sending a PR?

Yes

haydnli-shopify:
Open Question
- We’d love feedback from the Dstack team on this:
Should non-members be able to see any project metadata (e.g. description, config, recent activity) if a project is public? 
This could help users decide whether they want to join, but we’re unsure how much should be shown upfront. Would it make sense to adjust the UI to surface some of this information before joining? Open to guidance on what level of visibility makes sense here.

- What should the UX be for showing users projects they are already a member of vs projects they can join? Some ideas:
Keep showing a single Projects table/list; users join from the Project details page
Show 2 tables/lists on the Projects page: one for My Projects and one below for Other Projects
Show a single Projects table/list and enable other public projects to be joined using the search bar

peterschmidt85:
Hi @haydnli-shopify, certainly makes sense and matches exactly how [I see it](https://github.com/dstackai/dstack/issues/2670#issuecomment-2900749474).

[...]

PR #2759, opened 2025-06-05
Add public projects

### LINKED ISSUE #2670: [Feature]: Project owners should be able to make project as public
### Problem

When users first login they're met with an empty list of projects, and can only access an existing project if another user manually adds them. This creates a painful onboarding experience where users can't just self service join their target project within their organization and forces them to rely on other project members being available. 

### Solution

If I'm running an isolated instance for an organization generally I want users to be able to see most if not all projects and join them by clicking a button. To do this project admins should have the ability mark a project as public, which would then make it visible to all users and make a join button available for non-members.

### Workaround

_No response_

Yes

haydnli-shopify:
## Summary
Implements support for public projects to improve user onboarding experience. Users can now create public projects that are visible to all users, while maintaining backward compatibility with private projects as the default.

This PR is part of the larger self-serve membership management effort described in [#2742](https://github.com/dstackai/dstack/issues/2742).

Fixes #2670

## Changes Made
- Added `is_public` field to Project model (database, core model, API schemas)
- Updated `/api/projects/create` endpoint to accept `is_public` parameter  
- Implemented access control: non-members see only public projects, members see all accessible projects
- Optimized project listing with single efficient SQL query
- Added comprehensive test coverage (5 new tests covering creation and listing)
- Maintained backward compatibility (projects default to private)

## API Usage
```bash
# Create public project
curl -X POST /api/projects/create -H "Content-Type: application/json" -d '{"project_name": "my-project", "is_public": true}'

# Create private project (default)
curl -X POST /api/projects/create -H "Content-Type: application/json" -d '{"project_name": "my-project"}'
```

## Testing
All tests pass including new comprehensive coverage:
- `test_creates_public_project`
- `test_creates_private_project_by_default` 
- `test_creates_private_project_explicitly`
- `test_returns_public_projects_to_non_members`
- `test_member_sees_both_public_and_private_projects`


src/dstack/_internal/server/services/projects.py
```diff
@@ -56,7 +56,30 @@ async def list_user_projects(
     if user.global_role == GlobalRole.ADMIN:
         projects = await list_project_models(session=session)
     else:
-        projects = await list_user_project_models(session=session, user=user)
+        # Single query to get both:
```
dakota-shop: Is there anything that depends on `list_user_projects()` that makes the assumption that the user is a member of all returned projects?

src/dstack/_internal/server/services/projects.py
```diff
@@ -56,7 +56,30 @@ async def list_user_projects(
...
+                ) | (
+                    # OR project is public and user is not a member
+                    (ProjectModel.is_public == True) & 
+                    ProjectModel.id.notin_(
+                        select(MemberModel.project_id).where(MemberModel.user_id == user.id)
+                    )
+                )
+            )
+        )
+        projects = list(res.scalars().all())
```
colinjc: Put this logic in a helper function

haydnli-shopify: Fixed, its now in a helper function now! Lemme know if any other changes needed!

src/dstack/_internal/server/services/projects.py
```diff
@@ -56,7 +56,30 @@ async def list_user_projects(
...
+            .where(
+                ProjectModel.deleted == False,
+                # Either user is a member, OR project is public and user is not a member
+                (
+                    # User is a member (regardless of public/private)
+                    ProjectModel.id.in_(
+                        select(MemberModel.project_id).where(MemberModel.user_id == user.id)
+                    )
+                ) | (
+                    # OR project is public and user is not a member
```
dakota-shop: Seems like this query logic duplicates list_user_project_models() but with an extra clause; I wonder if it would be cleaner to add a parameter to `list_user_project_models()` to control whether this `OR` gets added.

src/dstack/_internal/server/services/projects.py
haydnli-shopify: I just fixed this by preserving backward compatibility, the function list_user_projects() now only returns projects where the user is a member (restoring original behavior). 
Some unit tests are also added!

src/dstack/_internal/server/services/projects.py
haydnli-shopify: I make some changes here based on your suggestions, I added a parameter to the function and the code is cleaner rn!

src/dstack/_internal/server/services/projects.py
```diff
@@ -53,10 +53,38 @@ async def list_user_projects(
...
         projects = await list_user_project_models(session=session, user=user)
+    
+    projects = sorted(projects, key=lambda p: p.created_at)
+    return [
+        project_model_to_project(p, include_backends=False, include_members=False)
+        for p in projects
+    ]
+
+
+async def list_user_accessible_projects(
```
dakota-shop, line 71: It looks to me like your new method is mostly a copy/paste of list_user_projects. I wonder if it makes sense to have a parameter there instead of having the separate method. We can ask the Dstack folks what they prefer.

src/dstack/_internal/server/services/projects.py
```diff
@@ -232,19 +262,58 @@ async def list_user_project_models(
...
+            .where(
+                ProjectModel.deleted == False,
+                # Either user is a member, OR project is public and user is not a member
+                (
+                    # User is a member (regardless of public/private)
+                    ProjectModel.id.in_(
+                        select(MemberModel.project_id).where(MemberModel.user_id == user.id)
+                    )
+                ) | (
+                    # OR project is public and user is not a member
```
dakota-shop: I wonder if it's possible to change the query semantics to "Either user is a member OR project is public" and remove duplicates before returning - that could reduce the query complexity.

src/dstack/_internal/server/services/projects.py
haydnli-shopify: Sure, I would just keep this like this for now

src/dstack/_internal/server/services/projects.py
haydnli-shopify: I just cleaned up my code earlier, but yes the way u mentioned is quite useful!

src/dstack/_internal/server/models.py
```diff
@@ -202,6 +202,7 @@ class ProjectModel(BaseModel):
     name: Mapped[str] = mapped_column(String(50), unique=True)
     created_at: Mapped[datetime] = mapped_column(NaiveDateTime, default=get_current_datetime)
     deleted: Mapped[bool] = mapped_column(Boolean, default=False)
+    is_public: Mapped[bool] = mapped_column(Boolean, default=False)
```
r4victor, line 205: Adding a new db column requires a db migration.

See https://github.com/dstackai/dstack/blob/master/contributing/MIGRATIONS.md

r4victor (changes requested):
The PR currently does not work since it's missing a db migration. Please ensure you've tested the changes.

Also note that we release from master, so PRs should not introduce partially working features or should hide them behind feature flags. This PR only makes public projects visible in the list but what happens if I open the project? I assume it will error. I'm ok with adding join/leave functionality in a separate PR, but at least it should not error when opening.

r4victor:
I suppose we should differentiate joined and not-joined projects in the UI, so maybe the API will return some field indicating that?

src/dstack/_internal/server/models.py
r4victor: Also note that the migration should work for existing databases. That is, in the migration, first create a nullable field, then set is_public for existing projects to False, then make it non-nullable.

haydnli-shopify: Thanks @r4victor! Added the migration 35f732ee4cf5_add_projectmodel_is_public.py following the the suggestion u gave!

src/dstack/_internal/server/services/projects.py
haydnli-shopify: @r4victor @dakota-shop raised a good point about the API design - list_user_accessible_projects is quite similar to list_user_projects with the main difference being whether to include public projects or not.

Would you prefer to lkeep separate functions (current approach we have) or add a parameter like include_public_projects=False to list_user_projects

What do you think works better for the dstack codebase patterns?

r4victor: Yes, I think having one function with include_public_projects parameter makes sense.

r4victor: > Returns projects where the user is a member.

the description is not entirely accurate since for global admins it's all projects.

r4victor:
@haydnli-shopify, now there is a problem that clicking on a public project (/api/projects/{project_name}/get) would result in 403 since the /get endpoint allows only project members. Is your plan to make public projects "non-clickable" in the UI before joining? Just want to confirm

Another option would be to show public project details before joining – it doesn't break any confidentiality since users can join and see anyway.

haydnli-shopify:
> @haydnli-shopify, now there is a problem that clicking on a public project (/api/projects/{project_name}/get) would result in 403 since the /get endpoint allows only project members. Is your plan to make public projects "non-clickable" in the UI before joining? Just want to confirm
> 
> Another option would be to show public project details before joining – it doesn't break any confidentiality since users can join and see anyway.

I went with your second option - the /api/projects/{project_name}/get endpoint now allows any authenticated user to view public project details before joining. Added a new ProjectMemberOrPublicAccess permission class that maintains existing member access while enabling public project discovery.
The 403 issue should be resolved. It should give a access denied here.


By the way, You could also check the PR for the second part of this issue here #2779, which handles works after we finish merging this current PR

r4victor:
I recommend installing pre-commits so that linting is done automatically: https://github.com/dstackai/dstack/blob/master/contributing/DEVELOPMENT.md

src/dstack/_internal/server/migrations/versions/35f732ee4cf5_add_projectmodel_is_public.py
```diff
@@ -0,0 +1,39 @@
...
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Add is_public column as nullable first
+    with op.batch_alter_table("projects", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("is_public", sa.Boolean(), nullable=True))
+
+    # Set is_public to False for existing projects
```
r4victor, line 25: Please avoid WHAT comments that duplicate the code. If LLM generated, please remove.

r4victor:
@haydnli-shopify, also please avoid force pushing PRs once under review – makes it unnecessary harder for others to review and contribute. If you need to pull changes from master, just do a regular merge into a PR branch.

r4victor:
@haydnli-shopify Have you tested the UI? I found the public project page won't open if a user is not a member because the UI requests /api/project/{project_name}/gateways/list which is only for members. I'm strongly against making project endpoints available for non-members even for public projects since it would complicate permission system and would be harder to keep track of. So I suggest the UI changes:
* Do not show project details to non-members at all.
* Or show a different view that does not require calls to project-specific endpoints.

I think this PR can be merged to keep things going, so please address the issue in #2795.

r4victor (approved):

Issue #2736, opened 2025-06-03
[Feature]: Health metrics


Nadine-H:
### Problem

We need to have metrics to understand the health of the dstack system, including the dstack server, database and the infrastructure the jobs run on. The current dstack metrics are focused more on usage rather than system health.

### Solution

As a starting point, I'm proposing we add at least the basic http server metrics, such as request/response latency and error rate broken by dstack operation.

Some additional metrics to consider: the amount of time a run is pending, latency from an apply request is accepted to job start, DB health metrics like connections error rate and latencies.

### Workaround

We're able to capture some of the http metrics from our load balancer but they're not granular enough. It's useful to see which dstack operations are failing for example or taking a long time to execute. The other metrics need to come from dstack so there is no workaround.

### Would you like to help us implement this feature by sending a PR?

Yes

r4victor:
@Nadine-H I suggest we close this issue and open new issues / PRs for specific metrics.

PR #2760, opened 2025-06-05
Add health metrics (Part 1)

Nadine-H:
Part of #2736 -  adding http metrics.

Chose to implement these metrics as custom rather than relying on standard fastapi metrics via instrumentation libraries. Reason being, we need to label `project_name` in order to get metrics at endpoint and project level. Standard libraries like `prometheus-fastapi-instrumentator` and `opentelemetry-instrumentation-fastapi` don't allow adding custom dynamic labels to their default metrics. 

**Note:**

- When adding these metrics, they were overwriting the existing custom metrics since they're both emitted on the same `/metrics` endpoint.
- I had to fetch the http metrics in the `metrics` router endpoint, and combine them with the custom metrics.
- The current approach is emitting the metrics as plain text on the `/metrics` endpoint, however the more standard Prometheus way is to register all the metrics using `prometheus_client` which uses a global registry to collect the metrics from all the app and takes care of emitting them on the `/metrics` endpoint.
- I did this implementation as a quick solution, however I will look into using `prometheus_client` in part 2 of this ticket for the additional custom metrics, if it makes adding new metrics easier, in which case it might require some refactoring of the existing prometheus metrics code.

### DIFF
--- a/src/dstack/_internal/server/app.py
+++ b/src/dstack/_internal/server/app.py
@@ -11,7 +11,7 @@ from fastapi import FastAPI, Request, Response, status
 from fastapi.datastructures import URL
 from fastapi.responses import HTMLResponse, JSONResponse, RedirectResponse
 from fastapi.staticfiles import StaticFiles
-from prometheus_fastapi_instrumentator import Instrumentator
+from prometheus_client import Counter, Histogram
 
 from dstack._internal.cli.utils.common import console
 from dstack._internal.core.errors import ForbiddenError, ServerClientError
@@ -64,6 +64,18 @@ from dstack._internal.utils.ssh import check_required_ssh_version
 
 logger = get_logger(__name__)
 
+# Server HTTP metrics
+REQUESTS_TOTAL = Counter(
+    "dstack_server_requests_total",
+    "Total number of HTTP requests",
+    ["method", "endpoint", "http_status", "project_name"],
+)
+REQUEST_DURATION = Histogram(
+    "dstack_server_request_duration_seconds",
+    "HTTP request duration in seconds",
+    ["method", "endpoint", "http_status", "project_name"],
+)
+
 
 def create_app() -> FastAPI:
     if settings.SENTRY_DSN is not None:
@@ -78,7 +90,6 @@ def create_app() -> FastAPI:
 
     app = FastAPI(docs_url="/api/docs", lifespan=lifespan)
     app.state.proxy_dependency_injector = ServerProxyDependencyInjector()
-    Instrumentator().instrument(app, metric_namespace="dstack", metric_subsystem="server")
     return app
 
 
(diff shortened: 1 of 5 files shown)


Nadine-H:
Realized that the `handler` label in the metrics doesn't resolve the `project_name` in the endpoint path:
`dstack_server_http_requests_total{handler="/api/projects/{project_name}/get",method="POST",status="2xx"} 1.0`

I think it's because the endpoint path is defined as a pattern, trying to see how I can add a custom label for `project_name` using `prometheus-fastapi-instrumentator`

Nadine-H:
Converting the PR to draft while I address the issue above, so no need for reviews in the meantime.

src/dstack/_internal/server/app.py
```diff
@@ -225,6 +240,36 @@ async def log_request(request: Request, call_next):
         )
         return response
 
+    # this middleware must be defined after the log_request middleware
+    @app.middleware("http")
+    async def log_http_metrics(request: Request, call_next):
+        def _extract_project_name(request: Request):
+            project_name = None
+            prefix = "/api/project/"
+            if request.url.path.startswith(prefix):
```
dakota-shop, line 249: Instead of manually parsing this out, I wonder if the parsing might already be done for us... anything interesting in the request object, e.g. in `path_params`?

Nadine-H: No `path_params` isn't available at the time the middleware is called. [Middlewares are executed before FastAPI routes the request to an endpoint](https://fastapi.tiangolo.com/tutorial/middleware/), so we only have access to the raw path.

> A "middleware" is a function that works with every request before it is processed by any specific path operation. And also with every response before returning it.

Nadine-H: This is the request object when it gets to the middleware:
`
request.__dict__
{'scope': {'type': 'http', 'asgi': {'version': '3.0', 'spec_version': '2.3'}, 'http_version': '1.1', 'server': ('127.0.0.1', 3000), 'client': ('127.0.0.1', 55612), 'scheme': 'http', 'method': 'POST', 'root_path': '', 'path': '/api/project/main/repos/get', 'raw_path': b'/api/project/main/repos/get', 'query_string': b'', 'headers': [(b'host', b'127.0.0.1:3000'), (b'user-agent', b'python-requests/2.32.3'), (b'accept-encoding', b'gzip, deflate, br'), (b'accept', b'*/*'), (b'connection', b'keep-alive'), (b'authorization', b'Bearer REDACTED'), (b'x-api-version', b'0.0.0'), (b'content-type', b'application/json'), (b'content-length', b'60')], 'state': {}, 'app': <fastapi.applications.FastAPI object at 0x1182e7680>}, '_receive': <function BaseHTTPMiddleware.__call__.<locals>.call_next.<locals>.receive_or_disconnect at 0x12b1494e0>, '_send': <function empty_send at 0x1186fbe20>, '_stream_consumed': False, '_is_disconnected': False, '_form': None, '_wrapped_rcv_disconnected': False, '_wrapped_rcv_consumed': False, '_wrapped_rc_stream': <async_generator object Request.stream at 0x12b151380>}
`

un-def (approved):
LGTM

PR #2767, opened 2025-06-06
Exclude backward incompatible fields from rest plugin calls

colinjc:
Minor version mismatches between the plugin server and dstack server were causing 422 errors


r4victor:
@colinjc I did some refactoring to avoid importing from client-related code and handle gateways and volumes. Should be good now.

r4victor (approved):

PR #2778, opened 2025-06-09
2674 prop filter

olgenn:


peterschmidt85 (approved):
Please update `Find runs` -> `Filter runs by properties`, otherwise good to merge!

peterschmidt85:
@olgenn Found only after merging, there is a critical regression. If no filters are set, no runs are shown.

I DM-ed you a few examples I reproduced on `https://sky.dstack.ai`

It's very important to fix before the release. Worst case scenario, if we cannot fix before the release tomorrow, lets rollback the change.

PR #2779, opened 2025-06-09
Add batch of members endpoints to fix race conditions (#2760)

### LINKED ISSUE #2684: [Feature]: Endpoint for adding single member to project
### Problem

Currently the only way to add members to a project is through the set members endpoint. The issue is that this endpoint requires passing an authoritative list of all members when called, which introduces a race condition when it's called in parallel. We're hitting this when trying to automatically add users to specific projects using some automations that we can't control the parallelism of. 

### Solution

Add an endpoint to append a new member to a project, instead of a full replacement. 

### Workaround

_No response_

Yes

haydnli-shopify:
## Summary

Implements atomic single and batch member management endpoints to eliminate race conditions when multiple automations add/remove project members simultaneously. The existing `set_members` endpoint required passing an authoritative list of **all** members, creating race conditions in concurrent scenarios.

This PR enhances the member management system to support both single-user and batch operations through the same endpoints, providing flexibility for different use cases while maintaining backwards compatibility.

**Fixes #2684 (parent ticket #2742)**

## Changes Made

- Enhanced `AddProjectMemberRequest` to accept single member or array of members  
- Enhanced `RemoveProjectMemberRequest` to accept single username or array of usernames  
- Implemented `add_project_members()` and `remove_project_members()` service functions supporting both single and batch operations  
- Enhanced `POST /api/projects/{project_name}/add_member` endpoint for single/batch operations  
- Enhanced `POST /api/projects/{project_name}/remove_member` endpoint for single/batch operations  
- Extended `ProjectsAPIClient` with backwards-compatible methods and new batch methods  
- Comprehensive permission controls (managers/admins only, global admin privileges) for all operations  
- Support for username or email lookup in both single and batch modes  
- Protection against removing all project admins  
- Atomic database operations eliminating read-modify-write patterns  
- Comprehensive test suite covering single, batch, and edge case scenarios

## API Usage

**Add a single member**  
`POST /api/projects/my-project/add_member`  
Body:
```json
{
  "members": {
    "username": "john@example.com",
    "project_role": "user"
  }
}
```

**Add multiple members (batch)**  
`POST /api/projects/my-project/add_member`  
Body:
```json
{
  "members": [
    {"username": "john@example.com", "project_role": "user"},
    {"username": "jane_doe", "project_role": "manager"},
    {"username": "alice", "project_role": "user"}
  ]
}
```

**Remove a single member**  
`POST /api/projects/my-project/remove_member`  
Body:
```json
{
  "usernames": "john_doe"
}
```

**Remove multiple members (batch)**  
`POST /api/projects/my-project/remove_member`  
Body:
```json
{
  "usernames": ["john_doe", "jane@example.com", "alice"]
}
```

**Update existing member role (via add_member)**  
`POST /api/projects/my-project/add_member`  
Body:
```json
{
  "members": {
    "username": "jane",
    "project_role": "admin"
  }
}
```

## Race Condition Fix

Previously, concurrent automations had to read and rewrite the entire project member list using the `set_members` endpoint, which led to race conditions and unintended overwrites. This PR introduces atomic `add_member` and `remove_member` operations that support both single and batch operations, allowing changes to apply independently and safely—even when triggered simultaneously.


peterschmidt85:
Why not allow adding/removing a batch of users?

src/dstack/_internal/server/routers/projects.py
```diff
@@ -91,3 +93,41 @@ async def set_project_members(
     )
     await session.refresh(project)
     return projects.project_model_to_project(project)
+
+
+@router.post(
+    "/{project_name}/add_member",
+)
+async def add_project_member(
```
colinjc: Why not `add_project_members` and always accept a list of members, instead of having to handle an edge case for single member?

src/dstack/_internal/server/schemas/projects.py
```diff
@@ -24,3 +24,13 @@ class MemberSetting(CoreModel):
 
 class SetProjectMembersRequest(CoreModel):
     members: List[MemberSetting]
+
+
+class AddProjectMemberRequest(CoreModel):
+    # Support both single member and multiple members
+    members: Union[MemberSetting, List[MemberSetting]]
```
colinjc: Much simpler to always take a list

src/dstack/_internal/server/services/projects.py
```diff
@@ -201,6 +201,90 @@ async def set_project_members(
...
+    
+    # Create lookup maps for both username and email
+    username_to_user = {user.name: user for user in users_found}
+    email_to_user = {user.email: user for user in users_found if user.email}
+    
+    # Get current max member_num
+    max_member_num = 0
+    for member in project.members:
+        if member.member_num is not None and member.member_num > max_member_num:
+            max_member_num = member.member_num
```
colinjc: I don't really get this approach to sorting the member list. Why not just add a "member_added" timestamp to the table and sort on (member_added, username)? @peterschmidt85

src/dstack/_internal/server/services/projects.py
```diff
@@ -201,6 +201,90 @@ async def set_project_members(
...
+
+    # Process each member to add
+    for member_setting in members_list:
+        user_to_add = username_to_user.get(member_setting.username) or email_to_user.get(member_setting.username)
+        if user_to_add is None:
+            # Skip users that don't exist (could log warning)
```
colinjc: I would error on adding a non-existing user

src/dstack/_internal/server/services/projects.py
```diff
@@ -201,6 +201,90 @@ async def set_project_members(
...
+            continue
+
+        # Check if user is already a member
+        existing_member = None
+        for member in project.members:
+            if member.user_id == user_to_add.id:
+                existing_member = member
+                break
```
colinjc: There could be a lot of members, could stick the existing ids in a set for quick checks instead of iterating the whole array every new user. 

```python
# earlier in function
member_ids = {m.user_id for m in project.members}
...
 if user_to_add.id in member_ids:
...
```

src/dstack/_internal/server/services/projects.py
```diff
@@ -449,3 +533,80 @@ def _is_project_admin(
...
     return False
+
+
+async def remove_project_members(
+    session: AsyncSession,
+    user: UserModel,
+    project: ProjectModel,
+    usernames: Union[str, List[str]],
+):
+    """Remove single or multiple members from a project."""
```
colinjc: Same as above, always accepting a list seems cleaner.

peterschmidt85:
One more thought:
In order to support `Join` or `Leave` functionality, can we allow `add_members` and `remove_members` when a user adds/removes themselves to a public project – without reqing project/global admin rights.

One this is done, we can also add UI via a separate PR.

src/dstack/_internal/server/security/permissions.py
```diff
@@ -99,6 +99,78 @@ async def __call__(
...
+        if project is None:
+            raise error_not_found()
+
+        # Global admin can always manage projects
+        if user.global_role == GlobalRole.ADMIN:
+            return user, project
+
+        # Project managers can remove members
+        project_role = get_user_project_role(user=user, project=project)
+        if project_role in [ProjectRole.ADMIN, ProjectRole.MANAGER]:
```
dakota-shop: This check seems redundant if we are going to `return user, project` below anyway.

src/dstack/_internal/server/services/projects.py
dakota-shop: I agree; also I wonder about a race condition where `member_num` could be the same for 2 members added simultaneously.

src/dstack/_internal/server/services/projects.py
```diff
@@ -201,6 +231,101 @@ async def set_project_members(
...
+        if member.member_num is not None and member.member_num > max_member_num:
+            max_member_num = member.member_num
+
+    # Process each member to add
+    for member_setting in members:
+        user_to_add = username_to_user.get(member_setting.username) or email_to_user.get(
+            member_setting.username
+        )
+        if user_to_add is None:
+            # Error on adding non-existing users instead of silently skipping
```
dakota-shop: I think you can remove this comment since the code is obvious

src/dstack/_internal/server/services/projects.py
```diff
@@ -201,6 +231,101 @@ async def set_project_members(
...
+            await add_project_member(
+                session=session,
+                project=project,
+                user=user_to_add,
+                project_role=member_setting.project_role,
+                member_num=max_member_num,
+                commit=False,
+            )
+            # Keep track of newly added members for subsequent iterations
+            member_ids.add(user_to_add.id)
```
dakota-shop: If you add a member and the user ID `X` gets added to `member_ids`, on a subsequent iteration you will check `member_ids` and find `X` on line 307; then on line 309, will you find `X` in `member_by_user_id`?

My guess this situation would only occur if the same user identifier appears multiple times in `members`? Is that a valid request?

src/dstack/_internal/server/services/projects.py
```diff
@@ -449,3 +606,96 @@ def _is_project_admin(
...
+    member_by_user_id = {m.user_id: m for m in project.members}
+
+    # Find members to remove and validate permissions
+    members_to_remove = []
+    admin_removals = 0
+
+    for username in usernames:
+        user_to_remove = username_to_user.get(username) or email_to_user.get(username)
+        if user_to_remove is None:
+            # Error on removing non-existing users instead of silently skipping
```
dakota-shop: Useless comment can be removed :)

src/dstack/_internal/server/services/projects.py
```diff
@@ -449,3 +606,96 @@ def _is_project_admin(
...
+        select(UserModel).where((UserModel.name.in_(usernames)) | (UserModel.email.in_(usernames)))
+    )
+    users_found = res.scalars().all()
+
+    # Create lookup maps
+    username_to_user = {user.name: user for user in users_found}
+    email_to_user = {user.email: user for user in users_found if user.email}
+
+    # Build a set of member IDs for faster membership checks
+    member_user_ids = {m.user_id for m in project.members}
```
dakota-shop: If you are building a map below with `user_id` as the key, why do you need this set? You could just look up the key in that map and I think it would be just as efficient.

src/dstack/_internal/server/services/projects.py
haydnli-shopify: Yes, in the new logic I eliminated the redundant member_ids set and now only use the member_by_user_id map for all tracking.

src/dstack/_internal/server/services/projects.py
haydnli-shopify: Removed the redundant member_user_ids set. Now only using the member_by_user_id map for both membership checks and member lookups. Same as above

haydnli-shopify:
@r4victor Both ticket are ready for review
PR for backend APIs: https://github.com/dstackai/dstack/pull/2779
PR for UI: https://github.com/dstackai/dstack/pull/2795

src/dstack/_internal/server/schemas/projects.py
```diff
@@ -8,7 +8,11 @@
 
 class CreateProjectRequest(CoreModel):
     project_name: str
-    is_public: bool = False
+    is_public: Optional[bool] = False
```
r4victor: Should not be optional – unnecessary and leads to wrong type annotations downstream.

src/dstack/_internal/server/schemas/projects.py
```diff
@@ -25,3 +29,13 @@ class MemberSetting(CoreModel):
 
 class SetProjectMembersRequest(CoreModel):
     members: List[MemberSetting]
+
+
+class AddProjectMemberRequest(CoreModel):
+    # Always accept a list of members for cleaner API design
```
r4victor: Please clean up unnecessary comments

src/dstack/_internal/server/security/permissions.py
```diff
@@ -122,16 +130,88 @@ async def __call__(
...
         project_role = get_user_project_role(user=user, project=project)
         if project_role is not None:
             return user, project
 
+        # If not a member, check if project is public
         if project.is_public:
             return user, project
 
+        # Neither member nor public project
+        raise error_forbidden()
```
r4victor, line 142: You're overriding my changes in #2759.

src/dstack/_internal/server/routers/projects.py
```diff
@@ -92,3 +97,60 @@ async def set_project_members(
...
+    return projects.project_model_to_project(project)
+
+
+@router.post(
+    "/{project_name}/update_visibility",
+)
+async def update_project_visibility(
+    body: UpdateProjectVisibilityRequest,
+    session: AsyncSession = Depends(get_session),
+    user_project: Tuple[UserModel, ProjectModel] = Depends(ProjectManager()),
```
r4victor: Only ProjectAdmin must be able to change such settings. Project managers can only manage members.

src/dstack/_internal/server/security/permissions.py
```diff
@@ -122,16 +130,88 @@ async def __call__(
...
+
+
+class ProjectManagerOrPublicJoin:
+    """
+    Allows:
+    1. Project managers to add any members
+    2. Any authenticated user to join public projects themselves
+    """
```
r4victor: Having a different permission class for every endpoint and then also having permissions check at the service layer complicates things. Have you considered making two different APIs: one for admins/managers to add/remove multiple users and another for users to join/leave?

r4victor: At the very least, rename to ProjectManagerOrPublicProject (that's what it really is) and also fix the docstring (it does not check if users join themselves – that's done in the service layer).

src/dstack/_internal/server/security/permissions.py
haydnli-shopify: Fixed! I've updated the code to exactly match the structure from #2759 if thats what you are mentioning!

src/dstack/_internal/server/security/permissions.py
haydnli-shopify: @r4victor I've renamed it to `ProjectManagerOrPublicProject` and fixed the docstring to accurately reflect what the permission class does in the current code here. 
Also, I can refactor this in a follow-up PR to have dedicated endpoints for admin/manager operations vs user self-service actions if you think that would be cleaner. 
Let me know if you'd prefer that approach and I can work on it!

peterschmidt85:
Is this PR a part of #2795? If so, should it be closed?

haydnli-shopify:
> Is this PR a part of #2795? If so, should it be closed?

Yes it is part of #2795, in order to make the code review easier, we split the PR into two different one, once this is merged, the changes in this PR cannot be seen in the UI one

peterschmidt85:
@r4victor please confirm if this is good to merge

src/dstack/_internal/server/services/projects.py
```diff
@@ -497,8 +596,102 @@ def _is_project_admin(
     user: UserModel,
     project: ProjectModel,
 ) -> bool:
+    # Check if user is the project owner
```
r4victor: @haydnli-shopify, once again, please remove all redundant comments that duplicate the code. We won't merge PRs with comments clutter.

If a function becomes hard to follow, better refactor the code into several appropriately named functions.

https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/#h2-b7593c698a940

haydnli-shopify: Sure, I just removed unnecessary comments!

peterschmidt85:
@haydnli-shopify Since we merged #2795, this PR can be closed, right?

haydnli-shopify:
> @haydnli-shopify Since we merged #2795, this PR can be closed, right?

Yes, i will close it!

PR #2796, opened 2025-06-12
Health metrics (Part 2)

Nadine-H:
Part of #2736 

Adding two custom metrics: 
- `dstack_submit_to_provision_duration_seconds`: Time from when a run has been submitted and first job provisioning
- `dstack_pending_runs_total`: Total number of pending runs

We can add metrics later too, but I think for now these two are helpful to see if there are runs stuck in SUBMITTED or PENDING states, which could be due to an issue with dstack or the underlying infrastructure.


src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -369,6 +370,26 @@ async def _process_active_run(session: AsyncSession, run_model: RunModel):
             run_model.status.name,
             new_status.name,
         )
+        if run_model.status == RunStatus.SUBMITTED and new_status == RunStatus.PROVISIONING:
+            current_time = datetime.datetime.now(datetime.timezone.utc)
```
un-def: Can be replaced with `dstack._internal.utils.common.get_current_datetime`

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -369,6 +370,26 @@ async def _process_active_run(session: AsyncSession, run_model: RunModel):
...
+            submit_to_provision_duration = (
+                current_time - run_model.submitted_at.replace(tzinfo=datetime.timezone.utc)
+            ).total_seconds()
+            logger.info(
+                "%s: run took %.2f seconds from submision to provisioning.",
+                fmt(run_model),
+                submit_to_provision_duration,
+            )
+            project_name = run_model.project.name
+            run_type = RunSpec.__response__.parse_raw(run_model.run_spec).configuration.type
```
un-def: `RunSpec` is already parsed, see line 218

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -369,6 +370,26 @@ async def _process_active_run(session: AsyncSession, run_model: RunModel):
...
+            run_metrics.log_submit_to_provision_duration(
+                submit_to_provision_duration, project_name, run_type
+            )
+
+        if new_status == RunStatus.PENDING:
+            run_spec = RunSpec.__response__.parse_raw(run_model.run_spec)
```
un-def: Same here, you can reuse `run_spec` from line 218

src/tests/_internal/server/background/tasks/test_process_runs.py
```diff
@@ -75,7 +75,26 @@ class TestProcessRuns:
...
+        with patch(
+            "dstack._internal.server.background.tasks.process_runs.run_metrics"
+        ) as mock_run_metrics:
+            await process_runs.process_runs()
+
+            mock_run_metrics.log_submit_to_provision_duration.assert_called_once()
+            args = mock_run_metrics.log_submit_to_provision_duration.call_args[0]
+            assert args[1] == run.project.name
+            assert args[2] == "service"
+            # Assert the duration is close to our expected duration (within 0.05 second tolerance)
```
un-def, line 104: This makes the test flaky, I suggest mocking `dstack._internal.utils.common.get_current_datetime` instead to make the assert reliable

src/tests/_internal/server/background/tasks/test_process_runs.py
```diff
@@ -201,8 +237,27 @@ async def test_submitted_to_provisioning_if_any(self, test_db, session: AsyncSes
...
+            "dstack._internal.server.background.tasks.process_runs.run_metrics"
+        ) as mock_run_metrics:
+            await process_runs.process_runs()
+
+            mock_run_metrics.log_submit_to_provision_duration.assert_called_once()
+            args = mock_run_metrics.log_submit_to_provision_duration.call_args[0]
+            assert args[1] == run.project.name
+            assert args[2] == "service"
+            assert isinstance(args[0], float)
+            # Assert the duration is close to our expected duration (within 0.05 second tolerance)
```
un-def: Same here, I think mocking `common.get_current_datetime` to return a predetermined value would make the assert easier and the test more reliable

src/dstack/_internal/server/background/metrics.py
```diff
@@ -0,0 +1,52 @@
+from prometheus_client import Counter, Histogram
```
un-def: I don't think the `background` package is the right place for this module. Not sure what is the best place, maybe `server.services.prometheus`? We can convert it to a package with submodules to separate the existing functions from the new `RunMetrics` class/instance.

cc @r4victor

Nadine-H: Makes sense, I'm not sure of the best naming that would distinguish the existing `prometheus.py` module and the new `metrics.py`. The major difference between them is how the metrics are collected; the earlier is more of a pull approach where the metrics are fetched when the `/metrics` endpoint is invoked, and the latter is more a push approach where the metrics are written to the prometheus client right when the runs are being processed. So maybe something like `pull_metrics` and `push_metrics`?

un-def: > the earlier is more of a pull approach where the metrics are fetched when the /metrics endpoint is invoked, and the latter is more a push approach where the metrics are written to the prometheus client right when the runs are being processed

But still, both are pull metrics (in terms of Prometheus, that is, clients pull metrics from the `/metrics` endpoint; there is also a [push approach](https://prometheus.io/docs/instrumenting/pushing/), which we currently don't use).

As the new metrics are about server health, I'd call the module `health.py`, `server.py` or something similar. Not sure about a new name for the existing metrics (run/job counters). I'll ask the team for suggestions, and if we come to the agreement, I'll push a commit renaming the modules (if, not, we can merge the PR as is and rename the modules later).

un-def (approved):

src/dstack/_internal/server/background/metrics.py
un-def: Renamed:

* `push_metrics.py` => `client_metrics.py` (for `prometheus_client`-based metrics)
* `pull_metrics.py` => `custom_metrics.py` (for metrics with custom implementations)

PR #2799, opened 2025-06-14
Rework default Docker images

### LINKED ISSUE #2793: [UX] Pre-build a EFA version of the default Docker image

peterschmidt85:
This PR does:

- [x] Adds `mpirun` and `/opt/nvcc-tests/build` to the `base` image
- [x] Drops separate images per Python version
- [x] Adds `devel-efa` image that comes with `/opt/nvcc-tests/build` and `uv` (fixes #2793)
- [x] Updates the Docker image version to 0.10
- [x] Updates the format of the base Docker image name to {version}-{'base|devel|devel-efa'}
- [x] Drops pre-pooling Docker images with VMs
- [x] Allows to use any `python` (not only minor versions)
- [x] Automatically choose `devel-efa` on EFA-enabled AWS instances
- [x] Add `OMPI_MCA_pml`, `OMPI_MCA_btl`, `OMPI_MCA_btl_tcp_if_exclude`, `NCCL_SOCKET_IFNAME`. See an [example](https://github.com/aws-samples/awsome-distributed-training/blob/84f05c607a021714753566afb8e75391a279b75f/2.ami_and_containers/containers/pytorch/0.nvcr-pytorch-aws.dockerfile#L202-L208).
- [x] Updated to Ubuntu 22.04

To be done separately:
- [ ] Build Docker & cloud images
- [ ] Updates documentation and examples (incl. add AWS EFA example)

Staged Docker images: https://hub.docker.com/repository/docker/dstackai/base-stgn/tags?name=ubuntu22.04&page=1


peterschmidt85:
https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html says `NCCL_SOCKET_IFNAME=^docker,lo` is unesessary. 

> By default, the loopback interface (lo) and docker interfaces (docker*) would not be selected unless there are no other interfaces available.

peterschmidt85:
One minor issue is that if users start using a different `python` versions, older clients will become incompatible:

<details><summary>Traceback</summary>
```
dstack ps
  File "/Users/cheptsov/.local/bin/dstack", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/cheptsov/.local/share/uv/tools/dstack/lib/python3.13/site-packages/dstack/_internal/cli/main.py", line 89, in main
    args.func(args)
    ~~~~~~~~~^^^^^^
  File "/Users/cheptsov/.local/share/uv/tools/dstack/lib/python3.13/site-packages/dstack/_internal/cli/commands/ps.py", line 49, in _command
    runs = self.api.runs.list(all=args.all, limit=args.last)
  File "/Users/cheptsov/.local/share/uv/tools/dstack/lib/python3.13/site-packages/dstack/api/_public/runs.py", line 715, in list
    runs = self._api_client.runs.list(
        project_name=self._project,
        repo_id=None,
        limit=1,
    )
  File "/Users/cheptsov/.local/share/uv/tools/dstack/lib/python3.13/site-packages/dstack/api/server/_runs.py", line 48, in list
    return parse_obj_as(List[Run.__response__], resp.json())
  File "pydantic/tools.py", line 38, in pydantic.tools.parse_obj_as
  File "pydantic/main.py", line 347, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 2 validation errors for ParsingModel[List[dstack._internal.core.models.runs.RunResponse]]
__root__ -> 0 -> run_spec -> configuration -> DevEnvironmentConfigurationResponse -> python
  '3.13.2' is not a valid PythonVersion (type=value_error)
__root__ -> 0 -> run_spec -> __root__
  Missing configuration (type=value_error)
```

</details>

peterschmidt85:
> One minor issue is that if users start using a different `python` versions, older clients will become incompatible:

@r4victor Should we truncate the Python version to `<major>.<minor>` to keep backward compatibility?

src/dstack/_internal/core/models/configurations.py
```diff
@@ -193,8 +186,8 @@ class BaseRunConfiguration(CoreModel):
         Optional[RegistryAuth], Field(description="Credentials for pulling a private Docker image")
     ] = None
     python: Annotated[
-        Optional[PythonVersion],
```
r4victor: Why removing PythonVersion enum? I don't see much benefit allowing to specify a patch version.

src/dstack/version.py
```diff
@@ -1,3 +1,3 @@
 __version__ = "0.0.0"
 __is_release__ = False
-base_image = "0.9"
```
r4victor, line 3: You should change the image version in the master after the images are built and published, so it's better to change it in a separate PR.

docker/base/Dockerfile
```diff
@@ -1,28 +1,78 @@
...
+    && apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/3bf863cc.pub \
+    && apt-get update --fix-missing \
+    && apt-get upgrade -y \
+    && ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime \
+    && apt-get install -y tzdata \
+    && dpkg-reconfigure --frontend noninteractive tzdata \
+    && cuda_version=$(echo ${CUDA_VERSION} | awk -F . '{ print $1"-"$2 }') \
+    && apt-get install -y --no-install-recommends \
+        cuda-libraries-dev-${cuda_version} \
+        cuda-nvcc-${cuda_version} \
```
r4victor, line 23: Installing nvcc in the base image, not only devel?

peterschmidt85: It is a builder stage, it's not included into the final one

src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -441,6 +443,40 @@ def _should_wait_for_other_nodes(run: Run, job: Job, job_model: JobModel) -> boo
     return False
 
 
+def _patch_base_image_for_aws_efa(
```
r4victor: If this is going to be in this file, at least move it to the end.

src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -517,14 +553,14 @@ def _process_provisioning_with_shim(
         cpu = None
         memory = None
         network_mode = NetworkMode.HOST
-
+    image_name = _patch_base_image_for_aws_efa(job_spec, job_provisioning_data)
```
r4victor, line 522: This is very hacky since you'll still have a different image specified in job spec, which can be hard to reason about and to figure out what's going on. It should be documented in JobSpec/get_default_image that image_name can be different from the specified there and why.

docker/base/efa/Dockerfile
```diff
@@ -1,15 +1,15 @@
-ARG BASE_IMAGE=dstackai/base:py3.12-0.7-cuda-12.1
+# syntax = edrevo/dockerfile-plus
```
r4victor, line 1: Can we live without it? An unfamiliar dependency that is no longer maintained.

src/tests/_internal/server/background/tasks/test_process_running_jobs.py
```diff
@@ -56,6 +63,9 @@
 
 pytestmark = pytest.mark.usefixtures("image_config_mock")
 
+if TYPE_CHECKING:
+    from dstack._internal.core.models.runs import JobProvisioningData
```
r4victor: Why

src/dstack/_internal/core/models/configurations.py
peterschmidt85: Rolled-back

src/dstack/version.py
peterschmidt85: Because we also change the format of the Docker image name, it's not possible to change it in a separate PR.
Moreover, we can build the new VM images from the PR branch before the merge, so it's OK.

docker/base/efa/Dockerfile
peterschmidt85: Without this dependency, we would need to duplicate the code

src/tests/_internal/server/background/tasks/test_process_running_jobs.py
peterschmidt85: Fixed

src/dstack/_internal/server/background/tasks/process_running_jobs.py
peterschmidt85: Fixed

src/dstack/_internal/server/background/tasks/process_running_jobs.py
peterschmidt85: Added documentation to `JobSpec/get_default_image`

src/dstack/version.py
r4victor: Ok, let's build the images from the PR then and merge when published.

r4victor (approved):

un-def (approved):

PR #2816, opened 2025-06-17
[UI] Allow to hide the Tour panel

olgenn:
Close https://github.com/dstackai/dstack-cloud/issues/286

### DIFF
--- a/frontend/src/App/constants.ts
+++ b/frontend/src/App/constants.ts
@@ -1,2 +1,3 @@
 export const AUTH_DATA_STORAGE_KEY = 'authData';
 export const MODE_STORAGE_KEY = 'mode';
+export const TUTORIAL_SHOW_STARTUP_STORAGE_KEY = 'tutorial-show-startup';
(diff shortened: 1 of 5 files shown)


peterschmidt85:
All works fine except one use-case:

- I log in as a user that already finished Tour steps.
- The Tour sidebar is not shown
- But if I click `Tour`, i see `Show at startup` is on

Screenshot 2025-06-18 at 20 32 14

- Should it off

olgenn:
@peterschmidt85  I think the toggle should be enabled by default, since it disables the Tour panel regardless of whether the user has completed the tour steps or not.

peterschmidt85:
I would auto-disable it.
Otherwise it’s misleading IMO. It says to show, but it doesn’t show.

olgenn:
@peterschmidt85 if user already finished Tour steps and enabled this switcher, Tour panel will be opened after refresh page. Right?

peterschmidt85:
Yeah, let's try this behavior

olgenn:
@peterschmidt85 I implemented this behavior. Please review it

peterschmidt85 (approved):

PR #2818, opened 2025-06-18
Add Open-R1 Example

Bihan:


examples/distributed-training/open-r1/.dstack.yml
```diff
@@ -0,0 +1,83 @@
...
+  - MODEL=Qwen/Qwen2.5-Coder-7B-Instruct
+  # Qwen2.5-Coder-7B-Instruct has 28 attention heads and should be divisible by TP and DP
+  - TP=4
+  - DP=2
+
+# Commands of the task
+commands:
+  - uv pip install vllm==0.8.5.post1
+  - uv pip install setuptools
+  - uv pip install https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.4.post1/flash_attn-2.7.4.post1+cu12torch2.6cxx11abiFALSE-cp312-cp312-linux_x86_64.whl
```
peterschmidt85, line 27: (Minor) I guess a simply pip install will try to build the wheel? Isn't there a way to prevent that without hardcoding the wheel url?

examples/distributed-training/open-r1/.dstack.yml
```diff
@@ -0,0 +1,83 @@
...
+    echo "VLLM host IP (last node): $VLLM_HOST"
+    
+    if [ "$USE_VLLM" = "true" ]; then
+      if [ "$DSTACK_NODE_RANK" -eq $(($DSTACK_NODES_NUM - 1)) ]; then
+        # Last Node runs VLLM server
+        echo "Starting VLLM server on Last Node (IP: $VLLM_HOST)"
+        trl vllm-serve --model $MODEL  --tensor_parallel_size $TP --data_parallel_size $DP --host 0.0.0.0
+      else
+        # Training node - adjust world size and nodes count for training
+        GPUS_PER_NODE=$(($DSTACK_GPUS_NUM / $DSTACK_NODES_NUM))
```
peterschmidt85: We already have the built-in `DSTACK_GPUS_PER_NODE` - which is calculated the same way.

examples/distributed-training/open-r1/.dstack.yml
```diff
@@ -0,0 +1,83 @@
...
+commands:
+  - uv pip install vllm==0.8.5.post1
+  - uv pip install setuptools
+  - uv pip install https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.4.post1/flash_attn-2.7.4.post1+cu12torch2.6cxx11abiFALSE-cp312-cp312-linux_x86_64.whl 
+  - git clone https://github.com/huggingface/open-r1.git
+  - cd open-r1
+  - uv pip install .
+  - |
+    # Get the last IP from DSTACK_NODES_IPS for vLLM node
+    VLLM_HOST=$(echo $DSTACK_NODES_IPS | tr ' ' '\n' | tail -n 1)
```
peterschmidt85: Shouldn't we move this under `if [ "$USE_VLLM" = "true" ]; then`?

examples/distributed-training/open-r1/.dstack.yml
```diff
@@ -0,0 +1,83 @@
...
+        # Last Node runs VLLM server
+        echo "Starting VLLM server on Last Node (IP: $VLLM_HOST)"
+        trl vllm-serve --model $MODEL  --tensor_parallel_size $TP --data_parallel_size $DP --host 0.0.0.0
+      else
+        # Training node - adjust world size and nodes count for training
+        GPUS_PER_NODE=$(($DSTACK_GPUS_NUM / $DSTACK_NODES_NUM))
+        ADJUSTED_NODES_NUM=$(($DSTACK_NODES_NUM - 1))
+        ADJUSTED_GPUS_TOTAL=$(($GPUS_PER_NODE * $ADJUSTED_NODES_NUM))
+        # Other nodes run training
+        echo "Starting training with VLLM on $VLLM_HOST"
```
peterschmidt85: Do we need this `echo`? Just thinking of simplifying the configuration. Same for the `echo` above...

Bihan: Echo is not necessary. We can remove it to simplify configuration

examples/distributed-training/open-r1/.dstack.yml
```diff
@@ -0,0 +1,83 @@
...
+            --hub_model_id sjbbihan/Qwen2.5-Coder-7B-Instruct \
+            --use_vllm false
+    fi
+
+resources:
+  gpu: 80GB:8
+  shm_size: 128GB
+
+volumes:
+   - /checkpoints:/checkpoints
```
peterschmidt85, line 78: (Minor) Just curious, given the VLLM may run on a random node, would checkpoints recover just work?

examples/distributed-training/open-r1/.dstack.yml
Bihan: @peterschmidt85 
The `flash_attn` url has specific ABI flag (`cxx11abiFALSE`). ABI flag can be TRUE or FALSE and the `torch `package installed by `vllm==0.8.5.post1` needs FALSE. 

To check whether we need TRUE or FALSE we need to do
`python -c "import torch; print(torch._C._GLIBCXX_USE_CXX11_ABI)"` which will return either TRUE or FASE.


I far as I remember, when I did `uv pip install flash_attn==2.7.4` I got undefined symbol error:
`ImportError: /root/.venv/lib/python3.12/site-packages/flash_attn_2_cuda.cpython-312-x86_64-linux-gnu.so: undefined symbol: _ZN3c105ErrorC2ENS_14SourceLocationENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
`

examples/distributed-training/open-r1/.dstack.yml
Bihan: @peterschmidt85 This is a very interesting question. I think theoretically it should not work if nodes are shuffled. This is because the node in which vLLM is running is not recognized by `accelerate` . The vLLM node is not within accelerate's `world size`.

Issue #2180, opened 2025-01-09
[Feature]: Zero-downtime in-place update for services


peterschmidt85:
### Problem

Currently, when re-deploying a service (e.g., a newer version), `dstack` stops the existing run before submitting the new one. This causes a disruption, which is unacceptable for any production-grade service.

### Solution

1. Introduce a revision number for service replicas. 
2. During an in-place update, ensure that older replicas are only stopped once the newer replicas are successfully running.

### Would you like to help us implement this feature by sending a PR?

Yes

jvstme:
There is a hacky workaround to achieve zero-downtime deployment. It works if there are only changes in the repo and not in the service configuration.

1. Add at least one extra replica in the `replicas` configuration property and reapply the configuration. This will upload the new repo revision to `dstack` and start a new replica/replicas to handle traffic while the other replicas are being redeployed.
1. Manually stop the old replicas by making your application exit cleanly. For example, attach to a replica with `dstack attach --replica <replica num> <run name>`, SSH into the container, and send a signal to your application. This will force `dstack` to redeploy the stopped replicas using the new repo revision.
1. Remove the redundant replica(s) from the `replicas` configuration property and reapply the configuration.

This workaround is very limited and inconvenient, so we are interested in supporting proper zero-downtime deployments orchestrated by `dstack`.

movchan74:
Hi @jvstme 

I've tried the workaround, but I ran into some issues with scaling (adding new replicas).

I've tried to scale from 1 replica to 2. When my app was starting, the API became unstable, and ~50% of requests were failing because they were directed to the replica where the app was not ready yet. I got 404 because my routes were not added yet. 

It takes a few minutes to start my app because it has to load all the models on the GPU, etc. So I cannot use scaling because it degrades the service. 

Maybe we can add a health check in nginx to make sure that the requests are not sent to the broken replica?

This is the issue not only for this workaround but for the scaling in general.

jvstme:
Hi @movchan74

I think Nginx has built-in availability checks enabled by default, but they only consider the replica unavailable if Nginx can't communicate with it at all, e.g. if the app is not bound to the port. So if you could adjust your app to only bind to the port once it's ready, it should work.

But many apps can't be implemented that way, so I agree that we'll need to support health checks. We intend to support them natively in #2181. I think this will be a prerequisite for the current ticket

movchan74:
Hi @jvstme, thanks for the pointers so far!

[...]

PR #2821, opened 2025-06-19
Rolling deployment for services

jvstme:
Support rolling deployment when some service
configuration properties are changed. During the
deployment, `dstack` updates service replicas one
by one. It first starts a new version of the
replica, then waits until it is running, then
stops the old version of the replica.

As part of the implementation, introduce a new run
and job property - `deployment_num`. When a new
configuration is applied, the `deployment_num` of
the run is incremented. Then, `dstack` gradually
updates the jobs so that their `deployment_num`
matches that of the run. Some jobs are updated
in-place, if the new configuration does not affect
their spec, others are redeployed as described
above.

```shell
> dstack apply

Active run test-service already exists. Detected configuration changes that can be updated in-place: ['image', 'env', 'commands']
Update the run? [y/n]: y
⠋ Launching test-service...
 NAME                            BACKEND          RESOURCES                        PRICE    STATUS       SUBMITTED
 test-service deployment=1                                                                  running      11 mins ago
   replica=0 job=0 deployment=0  aws (us-west-2)  cpu=2 mem=1GB disk=100GB (spot)  $0.0026  terminating  11 mins ago
   replica=1 job=0 deployment=1  aws (us-west-2)  cpu=2 mem=1GB disk=100GB (spot)  $0.0026  running      1 min ago
```

Initial and main part of #2180

### DIFF
--- a/src/dstack/_internal/server/background/tasks/process_runs.py
+++ b/src/dstack/_internal/server/background/tasks/process_runs.py
@@ -385,7 +385,7 @@ async def _handle_run_replicas(
         )
         return
 
-    await _update_jobs_to_new_deployment_in_place(run_model)
+    await _update_jobs_to_new_deployment_in_place(run_model, run_spec)
     if _has_out_of_date_replicas(run_model):
         non_terminated_replica_count = len(
             {j.replica_num for j in run_model.jobs if not j.status.is_finished()}
@@ -425,7 +425,7 @@ async def _handle_run_replicas(
             )
 
 
-async def _update_jobs_to_new_deployment_in_place(run_model: RunModel) -> None:
+async def _update_jobs_to_new_deployment_in_place(run_model: RunModel, run_spec: RunSpec) -> None:
     """
     Bump deployment_num for jobs that do not require redeployment.
     """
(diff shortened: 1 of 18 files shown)


jvstme:
I'm yet to test on Postgres and possibly add some unit tests. The rest should be ready for the initial release

src/dstack/_internal/server/background/tasks/process_runs.py
```diff
@@ -378,6 +338,130 @@ async def _process_active_run(session: AsyncSession, run_model: RunModel):
...
+    Bump deployment_num for jobs that do not require redeployment.
+    """
+
+    for replica_num, job_models in group_jobs_by_replica_latest(run_model.jobs):
+        if all(j.status.is_finished() for j in job_models):
+            continue
+        if all(j.deployment_num == run_model.deployment_num for j in job_models):
+            continue
+        new_job_specs = await get_job_specs_from_run_spec(
+            run_spec=RunSpec.__response__.parse_raw(run_model.run_spec),
```
un-def: Maybe parse the run spec only once outside the loop?

un-def (approved):

src/dstack/_internal/server/services/services/__init__.py
```diff
@@ -258,3 +259,21 @@ def _get_gateway_https(configuration: GatewayConfiguration) -> bool:
...
+) -> None:
+    scaler = get_service_scaler(configuration)
+    stats = None
+    if run_model.gateway_id is not None:
+        conn = await get_or_add_gateway_connection(session, run_model.gateway_id)
+        stats = await conn.get_stats(run_model.project.name, run_model.run_name)
+    run_model.desired_replica_count = scaler.get_desired_count(
+        current_desired_count=run_model.desired_replica_count,
+        stats=stats,
+        last_scaled_at=last_scaled_at,
```
r4victor, line 278: get_desired_count's last_scaled_at is datetime but int is passed

jvstme: Fixed the annotation in #2840

PR #2828, opened 2025-06-21
[UX] Simplify the use of Docker inside containers #2468

### LINKED ISSUE #2468: [UX] Simplify the use of Docker inside containers
Currently, users have to specify our Docker image, enable `privileged`, and add `start-dockerd` manually. The ideal solution would be to allow them to configure this declaratively, e.g., by specifying `docker: true`, which would handle everything automatically.

While users could still use the old method if they prefer, those who don’t want to would benefit from this new property.

peterschmidt85:
Done: 
- [x] Introduce the property
- [x] Update tests

To be done:
- [ ] Update the docs (#2831)

### DIFF
--- a/src/tests/_internal/server/routers/test_runs.py
+++ b/src/tests/_internal/server/routers/test_runs.py
@@ -216,7 +216,9 @@ def get_dev_env_run_plan_dict(
                     "home_dir": "/root",
                     "image_name": image_name,
                     "user": None,
-                    "privileged": True if docker else privileged,
+                    "privileged": True
+                    if docker
+                    else (privileged if privileged is not None else False),
                     "job_name": f"{run_name}-0-0",
                     "replica_num": 0,
                     "job_num": 0,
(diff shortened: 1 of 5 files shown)


src/tests/_internal/server/routers/test_runs.py
```diff
@@ -1204,6 +1317,55 @@ async def test_submits_run(
...
+                json=body,
+            )
+        assert response.status_code == 200, response.json()
+        assert response.json() == run_dict
+        res = await session.execute(select(RunModel))
+        run = res.scalar()
+        assert run is not None
+        res = await session.execute(select(JobModel))
+        job = res.scalar()
+        assert job is not None
```
un-def, line 1367: Last three lines are duplicate

un-def (approved):

src/tests/_internal/server/routers/test_runs.py
peterschmidt85: They are not the same. The code checks that the job is also returned.

un-def: Right, I overlooked that one is `RunModel` but the other is `JobModel`

PR #2834, opened 2025-06-22
[Bug]: Run/job status in UI/CLI is shown as `provisioning` instead of `pulling`

### LINKED ISSUE #2830: [Bug]: Run/job status in UI/CLI is shown as `provisioning` instead of `pulling`
### Steps to reproduce

-

### Actual behaviour

-

### Expected behaviour

-

### dstack version

0.19.15

### Server logs

```shell

```

### Additional information

_No response_

peterschmidt85:
Fixes #2830

### DIFF
--- a/src/dstack/_internal/core/models/runs.py
+++ b/src/dstack/_internal/core/models/runs.py
@@ -548,11 +548,15 @@ class Run(CoreModel):
             retry_on_events = (
                 jobs[0].job_spec.retry.on_events if jobs and jobs[0].job_spec.retry else []
             )
+            job_status = (
+                jobs[0].job_submissions[-1].status if jobs and jobs[0].job_submissions else None
+            )
             termination_reason = Run.get_last_termination_reason(jobs[0]) if jobs else None
         except KeyError:
             return values
         values["status_message"] = Run._get_status_message(
             status=status,
+            job_status=job_status,
             retry_on_events=retry_on_events,
             termination_reason=termination_reason,
         )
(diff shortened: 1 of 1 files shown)


src/dstack/_internal/core/models/runs.py
```diff
@@ -548,11 +548,15 @@ def _status_message(cls, values) -> Dict:
             retry_on_events = (
                 jobs[0].job_spec.retry.on_events if jobs and jobs[0].job_spec.retry else []
             )
+            job_status = (
+                jobs[0].job_submissions[-1].status if jobs and jobs[0].job_submissions else None
+            )
```
jvstme, line 553: For multi-job runs, it can be misleading to rely solely on the status of the first job.

Consider a two-replica service. If the first replica is `pulling` and the second is `running`, the current implementation will set the run status message to `pulling`.

```
 NAME               BACKEND          RESOURCES                        PRICE    STATUS   SUBMITTED  
 test-service                                                                  pulling  3 mins ago 
   replica=0 job=0  aws (us-west-2)  cpu=2 mem=1GB disk=100GB (spot)  $0.0025  pulling  21 sec ago 
   replica=1 job=0  aws (us-west-2)  cpu=2 mem=1GB disk=100GB (spot)  $0.0025  running  3 mins ago
```

However, if the first replica is `running` and the second is `pulling`, the message will be `running`.

```
 NAME               BACKEND          RESOURCES                        PRICE    STATUS   SUBMITTED  
 test-service                                                                  running  8 mins ago 
   replica=0 job=0  aws (us-west-2)  cpu=2 mem=1GB disk=100GB (spot)  $0.0025  running  6 mins ago 
   replica=1 job=0  aws (us-west-2)  cpu=2 mem=1GB disk=100GB (spot)  $0.0025  pulling  16 sec ago
```

This inconsistency is misleading because the replicas are supposed to have equal importance.

I can suggest to only rely on the job status for single-job runs. In multi-job runs, job statuses will not affect the run status message.

```suggestion
            job_status = (
                jobs[0].job_submissions[-1].status
                if len(jobs) == 1 and jobs[0].job_submissions
                else None
            )
```

jvstme (approved):

src/dstack/_internal/core/models/runs.py
peterschmidt85: Sorry, merged without! Will fix separately!

Issue #2497, opened 2025-04-09
[Feature]: Delete unused volumes after idle duration


colinjc:
### Problem

Currently volumes once created live forever until they're deleted. If we have dozens of dynamically created volumes created for specific workloads this will lead to volumes being forgotten and sitting unused, costing a lot of money. dstack currently provides no mechanism to automatically clean up unused volumes.

### Solution

Users should be able to specify `idle_duration` on volumes like they can with jobs, where idleness is the time since the volume was last used by a job.

### Would you like to help us implement this feature by sending a PR?

Yes

haydnli-shopify:
@peterschmidt85  will be working on it!

PR #2842, opened 2025-06-24
feat: add volume idle duration cleanup feature (#2497)

haydnli-shopify:
# Summary

Adds automatic cleanup of unused volumes after a configurable idle duration. Users can now set `auto_cleanup_duration` for volumes; if a volume isn't used by any job for that period, it's deleted automatically.

Addresses [Feature]: Delete unused volumes after idle duration #2497.

# Changes

- Add `auto_cleanup_duration` to `VolumeConfiguration`
- Track last usage with `last_job_processed_at` in `VolumeModel`
- Background task (`process_idle_volumes`) runs every 60s to delete idle volumes
- Update job processing to refresh volume usage timestamps
- Database migration for new field
- Unit and integration tests for all scenarios
- Validation to prevent external volumes from using auto-cleanup (not supported)

# Usage

- Set `auto_cleanup_duration` in volume config (YAML or API)
- Volumes not used for the specified time are auto-deleted
- Volumes still attached or with `auto_cleanup_duration: off` are never deleted
- External volumes (with `volume_id`) cannot use auto-cleanup

# Testing

- All core logic and edge cases are covered by tests
- Validation tests ensure external volumes are properly handled


src/dstack/_internal/core/models/volumes.py
```diff
@@ -44,6 +45,16 @@ class VolumeConfiguration(CoreModel):
         Optional[str],
         Field(description="The volume ID. Must be specified when registering external volumes"),
     ] = None
+    idle_duration: Annotated[
+        Optional[Union[str, int, bool]],
```
dakota-shop: Why do we need `int`? If 'off' means disable, why do we need `bool`?

src/dstack/_internal/core/models/volumes.py
```diff
@@ -44,6 +45,16 @@ class VolumeConfiguration(CoreModel):
         Optional[str],
         Field(description="The volume ID. Must be specified when registering external volumes"),
     ] = None
+    idle_duration: Annotated[
```
dakota-shop: Should this be named `max_idle_duration`, or maybe `auto_cleanup_idle_duration`?

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
```diff
@@ -0,0 +1,116 @@
...
+        if project not in volumes_by_project:
+            volumes_by_project[project] = []
+        volumes_by_project[project].append(volume_model.name)
+
+    for project, volume_names in volumes_by_project.items():
+        logger.info("Deleting idle volumes for project %s: %s", project.name, volume_names)
+        try:
+            await delete_volumes(session, project, volume_names)
+        except Exception as e:
+            logger.error("Failed to delete idle volumes for project %s: %s", project.name, str(e))
```
dakota-shop: If you want to log the stacktrace here (looks like we do), use `logger.exception` and leave out `str(e)` (assuming the dstack logger is a Python-standard logging interface)

src/dstack/_internal/core/models/volumes.py
haydnli-shopify: Yes, the boolean is no longer needed here, we could use string 'off' to overlap this
However, I would suggest keeping integer here, 
     since the integer here could be use for specifying idle duration in seconds directly, and -1 as a special value for specifying disabled

dakota-shop: Why would it be useful to specify this value in seconds when we can force it to be supplied as a human-readable string?

src/dstack/_internal/core/models/volumes.py
haydnli-shopify: Yes, I will have all the name changed!

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
haydnli-shopify: Sure, I have replace all the logger function with the correct ones

haydnli-shopify:
For this PR,
I've tested it locally and it's working correctly. I created test volumes with different auto-cleanup settings (30s, disabled, integer values) and verified the configuration parsing works. I tested the full lifecycle - volumes get created, go idle, and are automatically deleted after the specified duration. The background task handles all the edge cases we discussed.

If needed, I can upload my test configurations to show the testing approach.

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
```diff
@@ -0,0 +1,104 @@
...
+                .order_by(VolumeModel.last_processed_at.asc())
+                .limit(10)
+                .with_for_update(skip_locked=True)
+            )
+            volumes = list(res.unique().scalars().all())
+            if not volumes:
+                return
+            for volume in volumes:
+                await session.refresh(volume, ["project", "attachments"])
+                lockset.add(volume.id)
```
r4victor: Should not issue multiple queries under `async with lock` since this lock is supposed to be released asap. Instead take lock, select ids, add them to lockset, release the lock, then load the volumes with related attributes in one query – see how other processing tasks do that.

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
```diff
@@ -0,0 +1,104 @@
...
+    by_project = {}
+    for volume in volumes:
+        project = volume.project
+        if project not in by_project:
+            by_project[project] = []
+        by_project[project].append(volume.name)
+
+    for project, names in by_project.items():
+        try:
+            await delete_volumes(session, project, names)
```
r4victor: delete_volumes does the locking and it's not reentrant so you cannot use it from a task that locks too. It also commits halfway through, so your select for update will end there. Better to write a new deleting function in process_idle_volumes.

src/tests/_internal/server/background/tasks/test_process_idle_volumes.py
```diff
@@ -0,0 +1,163 @@
...
+            volume_provisioning_data=get_volume_provisioning_data(),
+        )
+
+        volume.last_job_processed_at = (
+            datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(hours=2)
+        ).replace(tzinfo=None)
+        await session.commit()
+
+        with patch(
+            "dstack._internal.server.background.tasks.process_idle_volumes.delete_volumes"
```
r4victor: Why mocking?

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
haydnli-shopify: Thanks for the response! I've updated it to take the lock and only select volume IDs, add them to the lockset and release the lock immediately, then load the full volumes with related attributes in a single query outside the lock. 
In this way our lock is released as soon as possible, and we still get all the data we need efficiently.

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
haydnli-shopify: I've fixed this by creating a new _delete_idle_volumes function that handles the deletion directly
this function simply marks volumes as deleted in the database, i think this could avoids the reentrant lockin here..

src/tests/_internal/server/background/tasks/test_process_idle_volumes.py
haydnli-shopify: I've removed the mocking and updated the test
the previous commit was mocking delete_volumes and just checking that it was called, but I realized tht doesn't actually test if the volume gets deleted properly. 
the current test runs the background task and then verifies that the volume is actually marked as deleted in the database by checking volume.deleted and volume.deleted_at.

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
```diff
@@ -0,0 +1,120 @@
...
+            for volume in volumes:
+                if _should_delete_volume(volume):
+                    to_delete.append(volume)
+
+            if to_delete:
+                await _delete_idle_volumes(session, to_delete)
+
+        finally:
+            for volume_id in volume_ids:
+                lockset.discard(volume_id)
```
r4victor: Can use lockset.difference_update

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
```diff
@@ -0,0 +1,120 @@
...
+from dstack._internal.utils.common import get_current_datetime
+from dstack._internal.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+async def process_idle_volumes():
+    lock, lockset = get_locker(get_db().dialect_name).get_lockset(VolumeModel.__tablename__)
+    async with get_session_ctx() as session:
+        # Take lock, select IDs, add to lockset, release lock
```
r4victor:
```suggestion
```

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
```diff
@@ -0,0 +1,120 @@
...
+    last_used = volume.last_job_processed_at or volume.created_at
+    last_used_utc = last_used.replace(tzinfo=datetime.timezone.utc)
+    now = get_current_datetime()
+
+    idle_time = now - last_used_utc
+    return max(idle_time, datetime.timedelta(0))
+
+
+async def _delete_idle_volumes(session: AsyncSession, volumes: List[VolumeModel]):
+    """Delete idle volumes without using the delete_volumes function to avoid locking conflicts."""
```
r4victor: Not a useful, confusing docstring

~/.dstack/server/config.yml
r4victor, line 1: Accidentally committed?

r4victor (changes requested):
The PR is ok but it does not actually deletes idle volumes – just marks them as deleted in the db.

r4victor:
The issue and the PR description say to support `idle_duration` for volumes, but the PR actually supports `auto_cleanup_duration`. Which one have you intended to support?

src/dstack/_internal/server/background/tasks/process_idle_volumes.py
```diff
@@ -0,0 +1,163 @@
...
+
+    await session.commit()
+
+
+async def _delete_volume_from_cloud(session: AsyncSession, volume_model: VolumeModel):
+    """Delete volume from cloud provider. Based on volumes.py:_delete_volume"""
+    volume = volume_model_to_volume(volume_model)
+
+    if volume.external:
+        return
```
r4victor: Maybe disallow creating external volumes with auto_cleanup_duration if it's not supported?

r4victor (changes requested):
Created a volume:

```
type: volume
name: my-gcp-volume
backend: gcp
region: europe-west4
availability_zone: europe-west4-b
size: 100GB
auto_cleanup_duration: 1m
```

process_idle_volumes fails with exception `sqlalchemy.exc.MissingGreenlet: greenlet_spawn` – should be easy to reproduce if creating any volume.

src/dstack/_internal/server/migrations/versions/d268739bd76b_merge_volume_cleanup_and_secrets_.py
r4victor, line 1: Avoid merge db migrations, rebase migrations on master

haydnli-shopify:
@r4victor Thanks for your response, I have fixed the issue you mentioned, and the PR description has been updated! A force push happened in this fix because I removed the merge migration as requested, updated the migration chain to properly rebase on master, and added validation to prevent external volumes from using auto_cleanup_duration at creation time.

r4victor:
@haydnli-shopify, Just do a regular merge to pull changes from master to avoid force push. No need to rebase the branch on master since we'll merge with squash anyway.

haydnli-shopify:
> @haydnli-shopify, Just do a regular merge to pull changes from master to avoid force push. No need to rebase the branch on master since we'll merge with squash anyway.

Sure! Thanks for your response. Feel free to let me know if there's anything to be improved in this PR!

r4victor (approved):

r4victor:
Made some adjustments, merging

Issue #2844, opened 2025-06-25
[Bug]: GCP may return logs in random order


peterschmidt85:
### Steps to reproduce

1.
Ensure GCP lgo storage is configured

2.

```
type: task
name: print-numbers

commands:
  - |
    python -c "for i in range(1, 100001): print(i)"
```

3. dstack logs print-numbers

### Actual behaviour

1. Logs appear as shuffled

### Expected behaviour

1. Logs shouldn't be shaffled

### dstack version

0.19.15

### Server logs

```shell

```

### Additional information

This happens because most of the logs have the same timestamp (in cases where they’re output very quickly).

I believe the reason is here:
https://github.com/dstackai/dstack/blob/2db1752326111fa7b4f7e174c6847966dda2dee4/runner/internal/executor/timestamp.go#L42

> Why aren’t we combining logs in the runner to send them in larger batches? Am I right in thinking that this would fundamentally solve the problem?

If we switch to plain text logs (#2841), they will be sent in batches, so that might potentially solve the issue too — although the risk still remains.

Also, GCP logs are returned very slowly compared to AWS (or reading from a file).

So, the options:
	1.	Stop using GCP Logging (e.g. use file log storage with a shared volume)
	2.	Solve this after migrating to plain text logs (via #2841).
	3.	Start combining logs in the runner

So far, I like options 1 and 2 the most.

peterschmidt85:
Should be fixed now as we migrated to ANSI stripped logs. If we see the issue again, we'll reopen it.

Issue #2783, opened 2025-06-10
[Bug]:


peterschmidt85:
### Steps to reproduce

1. Run a job that generates a lot of logs very quickly

### Actual behaviour

1. The server becomes unresponsive

### Expected behaviour

1. The server stays responsive

### dstack version

0.19.12

### Server logs

```shell

```

peterschmidt85:
@colinjc @dakota-shop if possible please add more context, meanwhile we will try to reproduce as is

r4victor:
Failed to reproduce so far on master, Ubuntu 24.04, GCP logging. RSS stays stable when increasing log output. At both 100Kb/s and 1MB/s of logs, RSS is ~450MB. At 100MB/s, it's ~550MB.

PR #2845, opened 2025-06-25
[Bug]: Use a unique token for log pagination instead of a timestamp

### LINKED ISSUE #2832: [Bug]: Can't see the entire log of a run via UI
### Steps to reproduce

0. Ensure no AWS CloudWatch or GCP Logging is not used to store run logs
1.
```
type: task
name: print-numbers

commands:
  - |
    python -c "for i in range(1, 100001): print(i)"
```

### Actual behaviour

1. The UI only shows logs starting from `98983`.

### Expected behaviour

1. The UI should allow to see all logs starting from `1`

### dstack version

0.19.15

### Server logs

```shell

```

### Additional information

1. Note, `dstack logs` works OK

### LINKED ISSUE #2833: [Bug]: Use a unique token for log pagination instead of a timestamp
### Steps to reproduce

0.
- Adjust the source code to display log `log.timestamp` instead of `log.message`
- Add `print("next_start_time", next_start_time)` at the end of the `while loop` (`resp = self._api_client.logs.poll` in `runs.py`)

1.
Run the following with the file/AWS/GCP logging.

```yaml
type: task
name: print-numbers

commands:
  - |
    python -c "for i in range(1, 100001): print(i)"
```

2. 

```shell
$ dstack logs print-numbers | sort | uniq -c > print-numbers.txt
```

### Actual behaviour

The polling function across all storages (file/GCP/AWS) uses `start_time` as a way to paginate logs.
In case too many event share the same timestamp, it truncates the output.

### Expected behaviour `start_time` for pagination. 
It should always use a unique token:
- AWS/GCP - support their own
- File logging may use line number from the file.

The polling function should NOT use 

### dstack version

0.19.15

### Server logs

```shell

```

### Additional information

[print-numbers.txt](https://github.com/user-attachments/files/20886613/print-numbers.txt)
[print-numbers-aws.txt](https://github.com/user-attachments/files/20886614/print-numbers-aws.txt)

peterschmidt85:
Fixes #2832, #2833, and potentially #2783

Here's what this PR does:

[Major]
- [x] Introduces `next_token` property in `/logs/poll`
- [x] File, AWS, GCP log storages now return paginated logs using `next_token`
- [x] Both CLI and UI now use `next_token` to load paginated logs (this fixes #2832)

[Minor]
- [x] Fixed the potential OOM issue with File log storage
- [x] UI now is capable of showing up to 10000000 rows (previously was 1000)

Here's what this PR doesn't solve:

- [x] #2844


src/dstack/_internal/server/services/logs/aws.py
```diff
@@ -78,14 +78,22 @@ def poll_logs(self, project: ProjectModel, request: PollLogsRequest) -> JobSubmi
...
-                cw_events = []
+                # Check if the group exists to distinguish between group not found vs stream not found
+                try:
+                    self._check_group_exists(self._group)
+                    # Group exists, so the error must be due to missing stream
+                    logger.debug("Stream %s not found, returning dummy response", stream)
+                    cw_events = []
+                except LogStorageError:
+                    # Group doesn't exist, re-raise the LogStorageError
+                    raise
```
r4victor, line 96: Why catching and re-raising immediately?

peterschmidt85: Yeah, it looks a bit odd but the idea is to catch a abstract exception and re-raise only if indeed streams do not exist.
This allows to  heck stream existence only if error occurs.

src/dstack/_internal/server/services/logs/gcp.py
```diff
@@ -78,35 +79,35 @@ def poll_logs(self, project: ProjectModel, request: PollLogsRequest) -> JobSubmi
...
             logger.warning("GCP Logging exception: %s", repr(e))
             # GCP Logging has severely low quota of 60 reads/min for entries.list
             raise ServerClientError(
                 "GCP Logging read request limit exceeded."
                 " It's recommended to increase default entries.list request quota from 60 per minute."
             )
-        # We intentionally make reading logs slow to prevent hitting GCP quota.
-        # This doesn't help with many concurrent clients but
-        # should help with one client reading all logs sequentially.
-        time.sleep(1)
```
r4victor, line 108: Without it, you'd hit quota and wouldn't be able to read the logs at all.

r4victor (approved):

src/dstack/_internal/server/services/logs/gcp.py
peterschmidt85: Yeah. Do you think we can raise a specific exception and handle it on the client?

r4victor: We already return an error when the quota is hit, see above. At that point, the client can do very little – you may have to wait for a minute or so. So we prevent quota hit in the first place.

src/dstack/_internal/server/services/logs/aws.py
r4victor: You catch and re-raise LogStorageError, which is basically the same as not catching it in the first place.

Issue #2738, opened 2025-06-04
[Feature] Support `files` in run configurations


peterschmidt85:
1. **Introduce the `files` property**

```
files:
  - .
```
or

```
files:
  - .:/workflow
```
or

```
files:
  - examples/
    path: /workflow/examples
```

or

```
files:
  - ./examples
    path: ./examples
```

un-def:
Closed via https://github.com/dstackai/dstack/pull/2848

PR #2848, opened 2025-06-26
Add `files` property to run configurations

un-def:
Each item in the `files` property maps a local path (a file or a dir) to a container path. Each local path is packed into a tar archive and uploaded to the server similar to the code repo archive/diff.

On the server, each archive is stored in the DB as `FileArchiveModel` linked to the user. Archive blobs are optionally uploaded to the storage (again, similar to the code blob).

When the job is submitted to the runner, files (if any) are uploaded after `/api/submit` but before `/api/upload_code`.

The runner unpacks archives as follows:
* If the path already exists, it's removed
* If any parent dirs of the path are missing, they are created as owned by the run user. The owner of the existing dirs is not changed.
* The owner of the path (and all subpaths in the case of the directory) is set to the run user. The permissions from the archive (and thus from the user's machine) are preserved.

Part-of: https://github.com/dstackai/dstack/issues/2738


src/dstack/api/server/_files.py
```diff
@@ -0,0 +1,18 @@
...
+
+from pydantic import parse_obj_as
+
+from dstack._internal.core.models.files import FileArchive
+from dstack._internal.server.schemas.files import GetFileArchiveByHashRequest
+from dstack.api.server._group import APIClientGroup
+
+
+class FilesAPIClient(APIClientGroup):
+    def get_archive_by_hash(self, hash: str) -> FileArchive:
```
r4victor, line 11: Any plans to use this function?

r4victor (approved):

src/dstack/api/server/_files.py
un-def: `dstack` itself does not use this API method at the moment, but it can be used to avoid uploading a large archive if it's already uploaded (e.g., in `RunCollection.apply_plan`):

1. Create an archive, get a hash
2. Check if the archive is already uploaded, calling this method
3. If the archive is already uploaded, reuse `id` returned by the method, upload the archive otherwise

We could improve `RunCollection.apply_plan` by adding this additional check for large archives (for small ones the additional API call only increases execution time).

PR #2849, opened 2025-06-26
[Bug]: Docker In Docker does not work with AMD 

### LINKED ISSUE #2847: [Bug]: Docker In Docker does not work with AMD
### Steps to reproduce

1. Create Docker In Docker Config as following:

```
type: dev-environment
name: dnd-amd

docker: true

ide: vscode

resources:
  gpu: MI300X:8
```

2. Apply the config
`dstack apply -f dnd.dstack.yml`

### Actual behaviour

It outputs validation message saying:

image` is required if `resources.gpu.vendor` is `amd

### Expected behaviour

Docker In Docker should not require `image:` with AMD and dstack should use its own image.

### dstack version

0.19.16

### Server logs

```shell

```

### Additional information

_No response_

peterschmidt85:
Fixes #2847

### DIFF
--- a/src/dstack/_internal/cli/services/configurators/run.py
+++ b/src/dstack/_internal/cli/services/configurators/run.py
@@ -398,9 +398,10 @@ class BaseRunConfigurator(ApplyEnvVarsConfiguratorMixin, BaseApplyConfigurator):
         else:
             has_amd_gpu = vendor == gpuhunt.AcceleratorVendor.AMD
             has_tt_gpu = vendor == gpuhunt.AcceleratorVendor.TENSTORRENT
-        if has_amd_gpu and conf.image is None:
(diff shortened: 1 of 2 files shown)


peterschmidt85:
@Bihan You can test the fix by installing `dstack` CLI from this branch. No server changes are required.

peterschmidt85:
@un-def I assume Docker-In-Docker should also work on Tenstorrent, is this correct?

un-def (approved):

un-def:
@peterschmidt85 

> I assume Docker-In-Docker should also work on Tenstorrent, is this correct?

Should work with any hardware (other than NVIDIA) that don't require a specific Docker runtime

Bihan (approved):

PR #2865, opened 2025-07-01
[runner] Fix `~` expansion in `files`

### LINKED ISSUE #2864: [Bug]: `~` in `files` resolves to `/workflow` if `user` is not set in either image or conf
### Steps to reproduce

`task.dstack.yml`:

```yaml
#user: root

commands:
  - ls -lah ~
  - ls -lah ~/.ssh
  - ls -lah /workflow

files:
 - ~/.ssh/id_rsa
 - file.txt:~/file.txt
```

1. `dstack apply --no-repo -f task.dstack.yml`:
    ```
    total 44K
    drwx------ 1 root root 4.0K Jul  1 11:01 .
    drwxr-xr-x 1 root root 4.0K Jul  1 11:01 ..
    -rw-r--r-- 1 root root 3.1K Jul  1 11:01 .bashrc
    drwxr-xr-x 3 root root 4.0K Jul  1 11:01 .cache
    drwxr-xr-x 3 root root 4.0K Jun 18 02:54 .config
    drwxr-xr-x 3 root root 4.0K Jul  1 11:01 .dstack
    drwxr-xr-x 4 root root 4.0K Jun 18 02:54 .local
    -rw-r--r-- 1 root root  161 Jul  9  2019 .profile
    drwx------ 1 root root 4.0K Jun 18 02:54 .ssh
    total 20K
    drwx------ 1 root root 4.0K Jun 18 02:54 .
    drwx------ 1 root root 4.0K Jul  1 11:01 ..
    -rw------- 1 root root  781 Jul  1 11:01 authorized_keys
    total 20K
    drwxr-xr-x 4 root root 4.0K Jul  1 11:01 .
    drwxr-xr-x 1 root root 4.0K Jul  1 11:01 ..
    drwxr-xr-x 2 root root 4.0K Jul  1 11:01 .ssh
    drwxr-xr-x 4 root root 4.0K Jul  1 11:01 .venv
    -rw-r--r-- 1 root root    5 Jul  1 11:01 file.txt
    ```
2. Uncomment `user: root` and apply the conf again:
    ```
    total 48K
    drwx------ 1 root root 4.0K Jul  1 10:59 .
    drwxr-xr-x 1 root root 4.0K Jul  1 10:59 ..
    -rw-r--r-- 1 root root 3.1K Jul  1 10:59 .bashrc
    drwxr-xr-x 3 root root 4.0K Jul  1 10:59 .cache
    drwxr-xr-x 3 root root 4.0K Jun 18 02:54 .config
    drwxr-xr-x 3 root root 4.0K Jul  1 10:59 .dstack
    drwxr-xr-x 4 root root 4.0K Jun 18 02:54 .local
    -rw-r--r-- 1 root root  161 Jul  9  2019 .profile
    drwx------ 1 root root 4.0K Jul  1 10:59 .ssh
    -rw-r--r-- 1 root root    5 Jul  1 10:59 file.txt
    total 24K
    drwx------ 1 root root 4.0K Jul  1 10:59 .
    drwx------ 1 root root 4.0K Jul  1 10:59 ..
    -rw------- 1 root root  781 Jul  1 10:59 authorized_keys
    -rw------- 1 root root 2.6K Jul  1 10:59 id_rsa
    total 12K
    drwxr-xr-x 3 root root 4.0K Jul  1 10:59 .
    drwxr-xr-x 1 root root 4.0K Jul  1 10:59 ..
    drwxr-xr-x 4 root root 4.0K Jul  1 10:59 .venv
    ```

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### dstack version

https://github.com/dstackai/dstack/commit/b2a28f552d5c998ea6bce949ab800ae8fd31b696

### Server logs

```shell

```

### Additional information

_No response_

un-def:
Use the runner process user (`root` expected) as the job user if the `user` directive is not specified in either the image or the dstack configuration.

Previously, the repo dir (`/workflow`) was used as a fallback for the home dir if the user is not specified.

Fixes: https://github.com/dstackai/dstack/issues/2864

### DIFF
--- a/runner/internal/executor/executor.go
+++ b/runner/internal/executor/executor.go
@@ -334,6 +334,9 @@ func (ex *RunExecutor) execJob(ctx context.Context, jobLogFile io.Writer) error 
 
 	// User must be already set
 	user := ex.jobSpec.User
+	// Strictly speaking, we need CAP_SETUID and CAP_GUID (for Cmd.Start()->
+	// Cmd.SysProcAttr.Credential) and CAP_CHOWN (for startCommand()->os.Chown()),
+	// but for the sake of simplicity we instead check if we are root or not
 	if ex.currentUid == 0 {
 		log.Trace(
 			ctx, "Using credentials",
@@ -341,23 +344,9 @@ func (ex *RunExecutor) execJob(ctx context.Context, jobLogFile io.Writer) error 
 			"username", user.GetUsername(), "groupname", user.GetGroupname(),
 			"home", user.HomeDir,
 		)
-
-		// 1. Ideally, We should check uid, gid, and supplementary groups mismatches,
-		// but, for the sake of simplicity, we only check uid. Unprivileged runner
-		// should not receive job requests where user credentials do not match the
-		// current user's ones in the first place (it should be handled by the server)
-		// 2. Strictly speaking, we need CAP_SETUID and CAP_GUID (for Cmd.Start()->
-		// Cmd.SysProcAttr.Credential) and CAP_CHOWN (for startCommand()->os.Chown()),
-		// but for the sake of simplicity we instead check if we are root or not
-		if *user.Uid != ex.currentUid && ex.currentUid != 0 {
-			return gerrors.Newf("cannot start job as %d, current uid is %d", *user.Uid, ex.currentUid)
-		}
-
 		if cmd.SysProcAttr == nil {
 			cmd.SysProcAttr = &syscall.SysProcAttr{}
 		}
-		// It's safe to setuid(2)/setgid(2)/setgroups(2) as unprivileged user if we use
-		// user's own credentials (basically, it's noop)
 		cmd.SysProcAttr.Credential = &syscall.Credential{
 			Uid:    *user.Uid,
 			Gid:    *user.Gid,
--- a/runner/internal/executor/executor.go
+++ b/runner/internal/executor/executor.go
@@ -133,16 +133,17 @@ func (ex *RunExecutor) Run(ctx context.Context) (err error) {
 	ctx = log.WithLogger(ctx, log.NewEntry(logger, int(log.DefaultEntry.Logger.Level))) // todo loglevel
 	log.Info(ctx, "Run job", "log_level", log.GetLogger(ctx).Logger.Level.String())
 
-	if ex.jobSpec.User != nil {
-		if err := fillUser(ex.jobSpec.User); err != nil {
-			ex.SetJobStateWithTerminationReason(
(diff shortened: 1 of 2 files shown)


runner/internal/executor/executor.go
```diff
@@ -331,15 +332,15 @@ func (ex *RunExecutor) execJob(ctx context.Context, jobLogFile io.Writer) error
 		cmd.Dir = workingDir
 	}
 
+	// User must be already set
 	user := ex.jobSpec.User
-	if user != nil {
+	if ex.currentUid == 0 {
```
jvstme, line 340: Some lines and comments in the `if` branch below are now irrelevant.

```go
// 1. Ideally, We should check uid, gid, and supplementary groups mismatches,
// but, for the sake of simplicity, we only check uid. Unprivileged runner
// should not receive job requests where user credentials do not match the
// current user's ones in the first place (it should be handled by the server)
// 2. Strictly speaking, we need CAP_SETUID and CAP_GUID (for Cmd.Start()->
// Cmd.SysProcAttr.Credential) and CAP_CHOWN (for startCommand()->os.Chown()),
// but for the sake of simplicity we instead check if we are root or not
if *user.Uid != ex.currentUid && ex.currentUid != 0 {
	return gerrors.Newf("cannot start job as %d, current uid is %d", *user.Uid, ex.currentUid)
}
```

```go
// It's safe to setuid(2)/setgid(2)/setgroups(2) as unprivileged user if we use
// user's own credentials (basically, it's noop)
```

runner/internal/executor/files.go
```diff
@@ -39,67 +39,69 @@ func (ex *RunExecutor) setupFiles(ctx context.Context) error {
...
-		if err := mkdirAll(ctx, dir, uid, gid); err != nil {
+		archivePath := path.Join(ex.archiveDir, fa.Id)
+		if err := extractFileArchive(ctx, archivePath, fa.Path, ex.workingDir, uid, gid, homeDir); err != nil {
 			return gerrors.Wrap(err)
 		}
+	}
 
-		if err := os.RemoveAll(p); err != nil {
-			log.Warning(ctx, "Failed to remove", "path", p, "err", err)
-		}
```
jvstme, line 73: This part seems to be missing from the new `extractFileArchive` function. Isn't it necessary? I think it was necessary to delete existing directories so that new directories could be extracted in their place

jvstme (approved):

runner/internal/executor/files.go
un-def: Good catch

PR #2876, opened 2025-07-04
[Feature] Strip ANSI codes from run logs and store them as plain text instead of bytes 

### LINKED ISSUE #2841: [Feature] Browsable, queryable, and searchable logs
### Problem

Because, currently, logs are stored as base64-encoded bytes, it's extremely hard to:

1. Show logs along with their timestamps.
2. [MAJOR] Implement descending pagination
3. Allow to search for logs in a given time range
4. [MAJOR] Query logs, esp. when they are stored in an external system such as AWS CloudWatch or GCP Logging
5. Support other external log storage systems (via a OpenTelemetry reporting interface)

### Underlying problem

1. Currently, containers run in the TTY mode. This allows to stream interactive logs via a websocket. Interactive logs are important to show interactive output from `pip`, `uv` and other similar tools.
2. At the sametime, to store also these logs on the server. Because TTY logs are in bytes (this is the only way they can be shown in the terminal once received via a websocket), we encode them via base64.
3. As the result, a) logs are encoded in base64 and not searchable; b) logs can be displayed in UI /CLI only if appended from the very beginning (due to ANSI codes). This means we cannot load these via pagination.

## Solution

In order to solve this, we need to automatically strip ANSI codes from the logs and convert bytes to plain text. Then, this text can be stored in any storage and then a) it can be searchable; b) it can be loaded via pagination starting from the most recent.

Note, stripping ANSI codes is a non-trivial task as it requires emulating support for all ANSI codes and buffering the logs.

Here's the initial implementation of this algorithm that needs to be tested: http://github.com/dstackai/ansistrip

**TODO:**

- [ ] Update the runner to use `ansistrip` and write the logs without encoding.
   - [x] Initial algorythm http://github.com/dstackai/ansistrip
   - [ ] Tune timeout settings for optimal experience and performance
   - [ ] Test extensively on various workloads
- [ ] Update the File/GCP/CloudWatch logging services to properly support pagination and descending
   - [x] Initial support with File/GCP/CloudWatch logging services
   - [ ] Support descending
- [ ] Update the `dstack logs` CLI and UI to support pagination and descending
   - [x] Initial update of CLI/UI
   - [ ] CLI: Support displaying timestamps, plus `--start`
   - [ ] UI: Support displaying timestamps, and descending

> [!WARNING]
> This is going to be a breaking change as we are dropping encoding/decoding.

peterschmidt85:
Fixes #2841

Out of the scope:
- [ ] Showing timestamps, descending support with UI, `--start` in UI, etc

^ This will be done in a separate PR

### DIFF
--- a/runner/internal/executor/executor.go
+++ b/runner/internal/executor/executor.go
@@ -473,12 +473,12 @@ func (ex *RunExecutor) execJob(ctx context.Context, jobLogFile io.Writer) error 
 	defer func() { _ = cmd.Wait() }() // release resources if copy fails
 
 	stripper := ansistrip.NewWriter(ex.jobLogs, AnsiStripFlushInterval, AnsiStripMaxDelay, MaxBufferSize)
+	defer stripper.Close()
 	logger := io.MultiWriter(jobLogFile, ex.jobWsLogs, stripper)
 	_, err = io.Copy(logger, ptm)
 	if err != nil && !isPtyError(err) {
 		return gerrors.Wrap(err)
 	}
-	stripper.Close()
 	return gerrors.Wrap(cmd.Wait())
 }
 
--- a/runner/internal/executor/executor.go
+++ b/runner/internal/executor/executor.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/creack/pty"
+	"github.com/dstackai/ansistrip"
 	"github.com/dstackai/dstack/runner/consts"
 	"github.com/dstackai/dstack/runner/internal/connections"
 	"github.com/dstackai/dstack/runner/internal/gerrors"
@@ -47,6 +48,7 @@ type RunExecutor struct {
 	state           string
 	jobStateHistory []schemas.JobStateEvent
 	jobLogs         *appendWriter
+	jobWsLogs       *appendWriter
 	runnerLogs      *appendWriter
 	timestamp       *MonotonicTimestamp
 
@@ -86,6 +88,7 @@ func NewRunExecutor(tempDir string, homeDir string, workingDir string, sshPort i
 		state:           WaitSubmit,
 		jobStateHistory: make([]schemas.JobStateEvent, 0),
 		jobLogs:         newAppendWriter(mu, timestamp),
+		jobWsLogs:       newAppendWriter(mu, timestamp),
 		runnerLogs:      newAppendWriter(mu, timestamp),
 		timestamp:       timestamp,
 
@@ -129,7 +132,8 @@ func (ex *RunExecutor) Run(ctx context.Context) (err error) {
 		}
 	}()
 
-	logger := io.MultiWriter(runnerLogFile, os.Stdout, ex.runnerLogs)
+	stripper := ansistrip.NewWriter(ex.runnerLogs, 500*time.Millisecond, 3*time.Second)
+	logger := io.MultiWriter(runnerLogFile, os.Stdout, stripper)
 	ctx = log.WithLogger(ctx, log.NewEntry(logger, int(log.DefaultEntry.Logger.Level))) // todo loglevel
 	log.Info(ctx, "Run job", "log_level", log.GetLogger(ctx).Logger.Level.String())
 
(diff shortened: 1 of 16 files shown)


runner/internal/executor/executor.go
```diff
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/creack/pty"
+	"github.com/dstackai/ansistrip"
```
r4victor, line 22: Any benefit having this in a separate repo vs adding to dstack?

runner/internal/executor/executor.go
```diff
@@ -188,6 +201,7 @@ func (ex *RunExecutor) Run(ctx context.Context) (err error) {
 		select {
 		case <-ctx.Done():
 			log.Error(ctx, "Job canceled")
+			stripper.Close()
```
r4victor: Not seeing any benefit calling stripper.Close() from multiple places. Just close it with defer like other cleanups?

runner/internal/executor/executor.go
```diff
@@ -129,7 +141,8 @@ func (ex *RunExecutor) Run(ctx context.Context) (err error) {
 		}
 	}()
 
-	logger := io.MultiWriter(runnerLogFile, os.Stdout, ex.runnerLogs)
+	stripper := ansistrip.NewWriter(ex.runnerLogs, AnsiStripFlushInterval, AnsiStripMaxDelay)
```
r4victor: BTW, do we actually need stripper for runner logs?

src/dstack/_internal/server/services/logs/aws.py
```diff
@@ -271,7 +269,7 @@ def _runner_log_event_to_cloudwatch_event(
     ) -> _CloudWatchLogEvent:
         return {
             "timestamp": runner_log_event.timestamp,
-            "message": b64encode_raw_message(runner_log_event.message),
+            "message": runner_log_event.message.decode(),
```
r4victor: .decode() will raise an exception if bytes are invalid utf-8. If it's handled by the runner, please leave a comment since it's very non-obvious when the runner itself returns base64 encoded bytes instead of strings.

runner/internal/executor/executor.go
r4victor: ansistrip docs should mention what it does with non-ascii, non-utf-8 sequences.

src/dstack/api/_public/runs.py
```diff
@@ -224,7 +223,7 @@ def logs(
                     ),
                 )
                 for log in resp.logs:
-                    yield base64.b64decode(log.message)
+                    yield log.message.encode()
```
r4victor, line 231: No need to return bytes anymore – the HTTP API change is backward incompatible anyway since it no longer returns base64-encoded bytes, so we may break the Python API just as well.

r4victor (approved):

src/dstack/api/_public/runs.py
peterschmidt85: I'd suggest that we change this in a separate PR where I want to refactor the method and have two separate methods for attached and detached logs.

runner/internal/executor/executor.go
peterschmidt85: If we don't use utf-8, we would have to use base64. That's why it's much easier to reuse the same code for both job and runner logs.

peterschmidt85:
@r4victor Just to confirm, this PR breaks backward compatibility. After upgrading, old logs won't be accessible.

runner/internal/executor/executor.go
```diff
@@ -431,11 +448,13 @@ func (ex *RunExecutor) execJob(ctx context.Context, jobLogFile io.Writer) error
...
 	defer func() { _ = cmd.Wait() }() // release resources if copy fails
 
-	logger := io.MultiWriter(jobLogFile, ex.jobLogs)
+	stripper := ansistrip.NewWriter(ex.jobLogs, AnsiStripFlushInterval, AnsiStripMaxDelay, MaxBufferSize)
+	logger := io.MultiWriter(jobLogFile, ex.jobWsLogs, stripper)
 	_, err = io.Copy(logger, ptm)
 	if err != nil && !isPtyError(err) {
 		return gerrors.Wrap(err)
 	}
+	stripper.Close()
```
un-def: Is there a reason not to use `defer stripper.Close()` right after the stripper is created?

un-def (approved):

r4victor:
> @r4victor Just to confirm, this PR breaks backward compatibility. After upgrading, old logs won't be accessible.

What happens to the old logs exactly? I don't see the code filtering them out, so users will see old logs in base64 in the UI / CLI?

olgenn (approved):

PR #2884, opened 2025-07-08
Add/remove SSH instances via in-place update

### LINKED ISSUE #2645: [Feature]: Fleet `ssh_config` in-place update
### Problem

Users cannot add or remove nodes from an SSH fleet without stopping all runs that are using the fleet.

### Solution

_No response_

### Workaround

_No response_

Yes

un-def:
It's now possible to add/remove instances to/from SSH fleets applying the fleet configuration with adjusted `ssh_config.hosts`.

Currently, the old and the new configurations must be equal except for the `identity_file` fields (ignored) and the order of hosts in the `ssh.config` list (but each host must remain the same or be removed).

Closes: https://github.com/dstackai/dstack/issues/2645

### DIFF
--- a/src/dstack/_internal/server/services/fleets.py
+++ b/src/dstack/_internal/server/services/fleets.py
@@ -413,6 +413,7 @@ async def apply_plan(
             current=current_fleet_sensitive.spec.configuration.ssh_config.hosts,
             new=spec.configuration.ssh_config.hosts,
         )
+        # `_check_can_update_fleet_spec` ensures hosts are not changed
         assert not changed_hosts, changed_hosts
         active_instance_nums: set[int] = set()
         removed_instance_nums: list[int] = []
--- a/src/dstack/_internal/server/services/fleets.py
+++ b/src/dstack/_internal/server/services/fleets.py
@@ -1,8 +1,10 @@
 import uuid
+from collections.abc import Callable
 from datetime import datetime, timezone
-from typing import List, Literal, Optional, Tuple, Union, cast
+from functools import wraps
+from typing import List, Literal, Optional, Tuple, TypeVar, Union, cast
 
-from sqlalchemy import and_, func, or_, select
+from sqlalchemy import and_, func, or_, select, update
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload, selectinload
 
@@ -13,10 +15,12 @@ from dstack._internal.core.errors import (
     ResourceExistsError,
     ServerClientError,
 )
+from dstack._internal.core.models.common import ApplyAction, CoreModel
 from dstack._internal.core.models.envs import Env
 from dstack._internal.core.models.fleets import (
     ApplyFleetPlanInput,
     Fleet,
+    FleetConfiguration,
     FleetPlan,
     FleetSpec,
     FleetStatus,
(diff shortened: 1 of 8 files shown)


src/dstack/_internal/server/services/fleets.py
```diff
@@ -327,87 +349,138 @@ async def apply_plan(
...
+                "Failed to apply plan. Resource has been changed. Try again or use force apply."
+            )
+
+    _check_can_update_fleet_spec(current_fleet_sensitive.spec, spec)
+
+    spec_json = spec.json()
+    current_fleet_model.spec = spec_json
+    await session.execute(
+        update(FleetModel).where(FleetModel.id == current_fleet_model.id).values(spec=spec_json)
     )
```
r4victor: Can we avoid doing update here? You already set current_fleet_model.spec, so it'll be updated on commit. Issuing update() here will upgrade sqlite transaction to write transaction, which can lead to "database is locked" if not committed immediately after.

src/dstack/_internal/server/services/fleets.py
```diff
@@ -327,87 +349,138 @@ async def apply_plan(
...
 
+    if (
+        current_fleet_sensitive.spec.configuration.ssh_config is not None
+        and spec.configuration.ssh_config is not None
+    ):
+        added_hosts, removed_hosts, changed_hosts = _calculate_ssh_hosts_changes(
+            current=current_fleet_sensitive.spec.configuration.ssh_config.hosts,
+            new=spec.configuration.ssh_config.hosts,
+        )
+        assert not changed_hosts, changed_hosts
```
r4victor: Worth commenting that _check_can_update_fleet_spec ensures hosts are not changed?

src/dstack/_internal/server/services/fleets.py
```diff
@@ -327,87 +349,138 @@ async def apply_plan(
     plan: ApplyFleetPlanInput,
     force: bool,
 ) -> Fleet:
-    return await create_fleet(
```
r4victor, line 330: I don't understand how apply_plan is supposed to handle concurrent execution. Won't it be possible to create multiple instances with same instance_num or do something worse?

un-def: https://github.com/dstackai/dstack/pull/2884/commits/bcc212ff5cd9b171c675c4dbc8af7bdefbb8d1df

src/dstack/_internal/server/services/fleets.py
```diff
@@ -370,94 +370,57 @@ async def apply_plan(
...
+    await session.commit()
+    async with (
+        get_locker(get_db().dialect_name).lock_ctx(FleetModel.__tablename__, [fleet_model.id]),
+        get_locker(get_db().dialect_name).lock_ctx(InstanceModel.__tablename__, instances_ids),
     ):
-        added_hosts, removed_hosts, changed_hosts = _calculate_ssh_hosts_changes(
-            current=current_fleet_sensitive.spec.configuration.ssh_config.hosts,
-            new=spec.configuration.ssh_config.hosts,
+        # Refetch after lock
+        # TODO: Lock instances with FOR UPDATE?
```
r4victor, line 393: If you don't lock the instances, a concurrent transaction that changes instances (e.g. process_instances->_check_instance) can override the changes (deletion) so that the instance won't be deleted. Doesn't seem critical for ssh instances but still can cause pretty nasty bugs.

r4victor (approved):

src/dstack/_internal/server/services/fleets.py
un-def: The locking code, including this TODO, was copied from `delete_fleets()`, and we decided to fix locking everywhere later.

PR #2914, opened 2025-07-18
Support scheduled runs

### LINKED ISSUE #2836: [Feature]: Scheduled tasks
### Problem

Sometimes, you may want to run tasks at a regular schedule (cron). Currently, `dstack` does't allow this so users have to use external systems such as AirFlow, CI/CD, etc.

It can be a lot more convenient if `dstack` allows to specify a `schedule` property. 

### Solution

This can be implemented via the syntax similar to [GitHub Actions](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule):

```
type: task
name: some-name

commands:
  - ./some_script.py
 
schedule:
  cron: "15 4,5 * * *"
```

Now, if you invoke `dstack apply`, the task status will become `scheduled`. 

The underlying mechanism is going to be the same as with the `retry` policy (when task stay in the `retrying status). Once finished, the task, will again becomes `scheduled`.

> [!NOTE]
> 1. The `cron` can be an array (like in GitHub).
> 2. The `cron` may als osupport aliases (like daily, etc [like in Kubernetes](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#schedule-syntax))

### Workaround

_No response_

Yes

r4victor:
Closes #2836

This PR introduces a `schedule` property for runs that allows starting runs periodically by specifying a cron expression:

```yaml
type: task
nodes: 1
schedule:
  cron: "*/15 * * * *"
commands:
  - ...
```

Notes:
* `schedule` is supported for all runs except services with scaling to zero (would require distinguishing pending scaled-to-zero services and pending scheduled services, which would be complicated, not worth it).
* `schedule` works with `retry` – a failed scheduled task will be retried according to `retry`. If retry duration is exceeded – it'll be run again according to `schedule`.
* If a scheduled run is stopped, it won't be scheduled again (since there has to be a way to stop schedule runs from scheduling). To schedule a run again, users may re-submit the run by applying the run configuration.

Implementation details:
* When a scheduled task is submitted, it’s created with status “pending” without jobs. RunModel gets a new next_triggered_at field that specifies the next time the task needs to be started. process_runs then processes pending runs that are ready to be processed according to next_triggered_at and creates job submissions for it. When a scheduled task is terminated, it goes to pending again with next_triggered_at updated, and the process repeats.
* process_runs does not select scheduled runs that are not ready to be scheduled (optimization).

TBD:
* Docs


src/dstack/_internal/core/models/profiles.py
```diff
@@ -167,6 +168,24 @@ def validate_time_window(cls, v: Union[int, str]) -> int:
         return v
 
 
+class Schedule(CoreModel):
+    cron: Annotated[
```
peterschmidt85, line 172: I thought we want to support multiple cron expression.

r4victor: yeah, I can do it next week since it'll be some more work.

peterschmidt85: But will i break backward compatibility if we add it later?

r4victor: Supported multiple crons

PR #2915, opened 2025-07-18
[UI] Query logs using descending

### LINKED ISSUE #2892: [UI] Query logs using `descending`
This issue has two sub-issues:

1. Implement infinite scrolling of logs using `descending` and loading the logs from end and them if the user scrolls up, load more logs.

> [!NOTE]
> Note, this implies that we ditch [xterm](https://xtermjs.org/) and use any standard component for showing logs.

2. Every log item should be shown along with the timestamp: first timestamp column, and then the log entry.

> [!NOTE]
> In the future, we'll add additional controls, e.g. to show/hide timestamps, and to specify start/end range.

peterschmidt85:
Fixes #2892

This includes:

- [x] Allowing `descending` in pull log requests
- [x] Supporting `descending` in `FileLogStorage` (this implements reading of lines from a file in a reversed order)
- [x] Changing the pull log limit from 1000 to 100 (since now our chunks are larger than they are used to be)
- [x] Re-implemented to make `FileLogStorage._read_lines_reversed` fast regardless of `start_offset`

Plus, in addition to #2892, this PR updates `CloudWatchLogStorage` to implement retrying when it fetches empty log pages. The maximum number of retries is set to `10`. In case `ascending` is used, after hitting the limit `CloudWatchLogStorage` sets `next_token` to `None` (meaning no more logs). This is not ideal but given that ascending mode is not going to be used anymore, we can live with that for now. If `descending` is used, `CloudWatchLogStorage` returns `next_token` properly. (allowing the UI or CLI implement the logic of trying again).

Out of the scope:

- [ ] Updating `dstack log` to use `descending` and support `--start`

### DIFF
--- a/src/dstack/_internal/server/services/logs/aws.py
+++ b/src/dstack/_internal/server/services/logs/aws.py
@@ -117,10 +117,7 @@ class CloudWatchLogStorage(LogStorage):
             if cw_events:
                 return cw_events, next_token
 
-            if not next_token:
-                return [], None
-
-            if next_token == previous_next_token:
+            if not next_token or next_token == previous_next_token:
                 return [], None
 
             previous_next_token = next_token
--- a/src/dstack/_internal/server/services/logs/aws.py
+++ b/src/dstack/_internal/server/services/logs/aws.py
@@ -55,6 +55,8 @@ class CloudWatchLogStorage(LogStorage):
     PAST_EVENT_MAX_DELTA = int((timedelta(days=14)).total_seconds()) * 1000 - CLOCK_DRIFT
     # "None of the log events in the batch can be more than 2 hours in the future."
     FUTURE_EVENT_MAX_DELTA = int((timedelta(hours=2)).total_seconds()) * 1000 - CLOCK_DRIFT
+    # Maximum number of retries when polling for log events to skip empty pages.
+    MAX_RETRIES = 10
 
     def __init__(self, *, group: str, region: Optional[str] = None) -> None:
         with self._wrap_boto_errors():
@@ -80,7 +82,7 @@ class CloudWatchLogStorage(LogStorage):
         next_token: Optional[str] = None
         with self._wrap_boto_errors():
             try:
-                cw_events, next_token = self._get_log_events(stream, request)
+                cw_events, next_token = self._get_log_events_with_retry(stream, request)
             except botocore.exceptions.ClientError as e:
                 if not self._is_resource_not_found_exception(e):
                     raise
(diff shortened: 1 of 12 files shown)


peterschmidt85:
Pending changes from @olgenn:

1. Fix the height of the `Logs` UI component.
2. Fix the background color of the `Logs` UI component.
3. Add `Legacy mode` toggler (off by default) that checks if log messages are base64-encoded strings and in this case decodes them.

olgenn (approved):

src/dstack/_internal/server/services/logs/aws.py
```diff
@@ -101,7 +103,45 @@ def poll_logs(self, project: ProjectModel, request: PollLogsRequest) -> JobSubmi
...
+                diagnose=request.diagnose,
+            )
+
+        if not request.descending:
+            logger.debug(
+                "Stream %s: exhausted %d retries without finding logs, returning empty response",
+                stream,
+                self.MAX_RETRIES,
+            )
+        return [], next_token if request.descending else None
```
r4victor, line 146: Needs a comment explaining why returning next_token in case of descending and None otherwise

src/dstack/_internal/server/services/logs/aws.py
```diff
@@ -101,7 +103,45 @@ def poll_logs(self, project: ProjectModel, request: PollLogsRequest) -> JobSubmi
...
+            cw_events, next_token = self._get_log_events(stream, current_request)
+
+            if cw_events:
+                return cw_events, next_token
+
+            if not next_token:
+                return [], None
+
+            if next_token == previous_next_token:
+                return [], None
```
r4victor: seems natural to combine into one if

src/dstack/_internal/server/services/logs/filelog.py
```diff
@@ -94,6 +103,99 @@ def poll_logs(self, project: ProjectModel, request: PollLogsRequest) -> JobSubmi
...
+        if request.next_token:
+            try:
+                start_offset = int(request.next_token)
+                if start_offset < 0:
+                    raise ValueError("Offset must be non-negative")
+            except (ValueError, TypeError):
+                raise ServerClientError(
+                    f"Invalid next_token for descending read: {request.next_token}. "
+                    f"Must be a non-negative integer offset."
+                )
```
r4victor: Do we need different next_token parsing/validation for descending and ascending? Seems like it can be moved to a separate function.

src/dstack/_internal/server/services/logs/filelog.py
```diff
@@ -94,6 +103,99 @@ def poll_logs(self, project: ProjectModel, request: PollLogsRequest) -> JobSubmi
...
+        offset of the start of each line. This is memory-efficient for large files.
+        """
+        with open(filepath, "rb") as f:
+            if start_offset is None:
+                f.seek(0, os.SEEK_END)
+                cursor = f.tell()
+            else:
+                f.seek(0, os.SEEK_END)
+                file_size = f.tell()
+                cursor = file_size
```
r4victor: The two if branches are functionally the same.

r4victor:
> This is not ideal but given that ascending mode is not going to be used anymore, we can live with that for now.

You mean not used by the UI? Still used by the CLI and the API default t.

PR #2927, opened 2025-07-24
Service probes

### LINKED ISSUE #2181: [Feature]: Allow to configure a service healthcheck
### Problem

When running a service, I currently have to manually check the endpoint repeatedly to determine if it’s ready to handle requests (e.g. the model is loaded) or if there are any health issues. This is particularly critical for production-grade services.

### Solution

1. Add an option in the service configuration to define a URL that dstack can use to:
a) Verify the service is up and healthy.
b) Warm up the service.
2. If the service does not pass the health check within the configured timeout, mark it as unhealthy.

### Workaround

_No response_

Yes

jvstme:
Allow configuring HTTP probes for services and use probe results to avoid downtime during
deployments.

```yaml
type: service
port: 8000
replicas: 2
commands:
- mkdir /www
- cd /www
- touch health-0
- touch health-1
- python -m http.server
probes:
- type: http
  url: /health-0
  interval: 5s
  ready_after: 2
- type: http
  url: /health-1
  interval: 5s
  timeout: 1
```

https://github.com/user-attachments/assets/6b59ee9c-a986-4a11-94a6-5169978f69a8

#2181


src/dstack/_internal/server/models.py
```diff
@@ -427,6 +427,9 @@ class JobModel(BaseModel):
     replica_num: Mapped[int] = mapped_column(Integer)
     deployment_num: Mapped[int] = mapped_column(Integer)
     job_runtime_data: Mapped[Optional[str]] = mapped_column(Text)
+    probes: Mapped[list["ProbeModel"]] = relationship(
+        back_populates="job", lazy="selectin", order_by="ProbeModel.probe_num"
```
r4victor: Pls avoid autoload of related attributes since it leads to inefficient queries – always load explicitly when needed.

r4victor:
I recently made an analysis of dstack server db queries that showed a substantial load on the db and slow api endpoints caused by inefficient/redundant queries. Improved this in #2928 and #2929. For now on, I recommend loading only the required attributes and relationship when makes sense. Please adjust this for probes.

src/dstack/_internal/server/background/tasks/process_probes.py
```diff
@@ -0,0 +1,169 @@
...
+                    job_spec: JobSpec = JobSpec.__response__.parse_raw(probe.job.job_spec_data)
+                    probe_config = job_spec.probes[probe.probe_num]
+                    # Schedule the next probe execution in case this execution is interrupted
+                    probe.due = get_current_datetime() + _get_probe_async_processing_timeout(
+                        probe_config
+                    )
+                    # Execute the probe asynchronously outside of the DB session
+                    PROBES_SCHEDULER.add_job(partial(_process_probe_async, probe, probe_config))
+        finally:
+            probe_lockset.difference_update(probe_ids)
```
r4victor, line 84: You should commit changes before releasing the in-memory locks. See https://github.com/dstackai/dstack/blob/e7c45258657180141f3cb3d92b519df0bad0919b/contributing/LOCKING.md?plain=1#L59

src/dstack/_internal/server/background/tasks/process_probes.py
```diff
@@ -0,0 +1,169 @@
...
+                "Expected probe interval to be parsed by validator"
+            )
+            await session.execute(
+                update(ProbeModel)
+                .where(ProbeModel.id == probe.id)
+                .values(
+                    success_streak=0 if not success else ProbeModel.success_streak + 1,
+                    due=get_current_datetime() + timedelta(seconds=probe_config.interval),
+                )
+            )
```
r4victor: I think this is problematic – you have batch_size=100, so there can be that many _process_probe_async. If every function produces a write transactions – there can be too many of them and you don't control the write rate anymore, which can lead to "database is locked" on SQLite and overall is not good. I think we should batch probe execution and updates as well. Unbounded `probe_config.timeout` is a problem here. Perhaps we should put some restrictions on it to make batch updates work.

src/dstack/_internal/core/models/configurations.py
```diff
@@ -162,6 +164,58 @@ class RateLimit(CoreModel):
     ] = 0
 
 
+class ProbeConfig(CoreModel):
+    type: Literal["http"]  # expect other probe types in the future, namely `exec`
+    url: Annotated[str, Field(description="The URL to request")] = "/"
+    timeout: Annotated[
+        Union[int, str],
+        Field(description=("Maximum amount of time the HTTP request is allowed to take")),
+    ] = "10s"
```
r4victor: Please avoid putting defaults directly into models. We did it in the past and it caused lots of issues. Primarily being unable to update defaults on the server-side and detecting if the user set the field or it was set by model defaults. Fields should be Optional[] with defaults set on the server.

src/dstack/_internal/server/background/tasks/process_probes.py
```diff
@@ -0,0 +1,169 @@
...
+from dstack._internal.server.services.logging import fmt
+from dstack._internal.server.services.ssh import container_ssh_tunnel
+from dstack._internal.utils.common import get_current_datetime, get_or_error
+from dstack._internal.utils.logging import get_logger
+
+logger = get_logger(__name__)
+BATCH_SIZE = 100
+SSH_CONNECT_TIMEOUT = timedelta(seconds=10)
+PROCESSING_OVERHEAD_TIMEOUT = timedelta(minutes=1)
+PROBES_SCHEDULER = AsyncIOScheduler()
```
r4victor, line 34: Why introducing a new scheduler and adding _process_probe_async there? I think process_probes should take probes for processing, commit, and then process probes (asyncio.gather if need process multiple concurrently), waiting for processing to complete. There can be multiple instances of process_probes controlled by max_instances= of the background scheduler. The current approach puts no upper bound on active processing transactions.

r4victor: We'll also migrate the rest of background tasks to a similar approach, and I don't think we want to introduce a separate scheduler for every one.

src/dstack/_internal/server/background/tasks/process_probes.py
r4victor: At the same time, I think batch updates also complicate things since processing different resources becomes dependent (at least time-wise). Per-resource updates may be ok and performant enough for Postgres. We should check how much the problem it would be for sqlite and what else can we do there differently.

r4victor: A good solution for sqlite (while doing an update for every probe processed) may be an app-level lock that syncs all probe updates. SQLite syncs concurrent writes anyway and its file-based locking is much less efficient than our app-level lock. There is a per-resource locking now but it doesn't serve that purpose.

src/dstack/_internal/server/background/tasks/process_probes.py
jvstme: > Why introducing a new scheduler

The scheduler is just a way to run fire-and-forget tasks. Similar behavior can be achieved with `asyncio.create_task`, but it requires [boilerplate code](https://docs.python.org/3/library/asyncio-task.html#creating-tasks) to store references to tasks, so I decided to use a scheduler instead.

> We'll also migrate the rest of background tasks to a similar approach, and I don't think we want to introduce a separate scheduler for every one.

We can introduce a global scheduler for fire-and-forget tasks when needed.

> I think process_probes should take probes for processing, commit, and then process probes (asyncio.gather if need process multiple concurrently), waiting for processing to complete

That would result in `process_probes` being blocked until all probes are executed. If there are probes that take a long time to execute, such as LLM prompts, the `max_instances` of `process_probes` tasks will be quickly exhausted and shorter probes will not be executed at the user-configured interval.

> The current approach puts no upper bound on active processing transactions.

There will typically be at most one instance of the processing task (`process_probes`), since this task exits quickly after starting `_process_probe_async` tasks. As for `_process_probe_async` — yes, there can be as many of these tasks as there are active probes. This may be acceptable, because these tasks happen outside of a db transaction and spend most of the time waiting for I/O from the service or from SSH. Each task can trigger a db transaction, but it should be short since it only consists of one `UPDATE` query.

Setting bounds on the number of `_process_probe_async` tasks will result in probes not being executed at the user-configured interval.

If the number of db write transactions is a problem, even through these transactions are short, we can try to limit them by having a separate worker write the results of probe executions in batches instead of writing them for each probe in `_process_probe_async`.

If the overall number of concurrent probe execution tasks is a problem, we can try to solve it by moving probe execution to `dstack-runner`, but this requires further research.

r4victor: > If the number of db write transactions is a problem

it is a problem for sqlite – but I think we can mitigate it with a app-level lock.

r4victor: Agree overall but

> there can be as many of these tasks as there are active probes. This may be acceptable, because these tasks happen outside of a db transaction and spend most of the time waiting for I/O from the service or from SSH.

there should be limit on the probe timeout and number of probes, otherwise the server can be overloaded with many never-ending probes.

src/dstack/_internal/server/background/tasks/process_probes.py
jvstme: > A good solution for sqlite (while doing an update for every probe processed) may be an app-level lock

We can consider this separately, since it may require more testing. The current solution works on SQLite without errors with 50 parallel probe executions. With 100 executions, database-is-locked errors do happen from time to time, but I don't think 100 parallel probes is realistic for SQLite setups

r4victor (approved):

Issue #2669, opened 2025-05-20
[Bug]: Lambda instance becomes ureachable after `dstack server` restart


jvstme:
### Steps to reproduce

1. Create an instance using the `lambda` backend, wait until it becomes `idle` or `busy`.
2. Restart `dstack server`

### Actual behaviour

The instance becomes unreachable and never recovers. If it was running a job, the job is terminated. `dstack-shim` no longer runs on the instance.

The first shim health check attempt fails with this error:
```
           DEBUG    dstack._internal.server.background.tasks.process_instances:747 Check instance cloud-0 status. shim health: Can't request shim:          
                    ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```
All other health checks fail with this error:
```
           DEBUG    dstack._internal.server.background.tasks.process_instances:747 Check instance cloud-0 status. shim health: Can't request shim:          
                    ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```

### Expected behaviour

The instance remains `idle` or `busy`.

### dstack version

master

### Server logs

```shell

```

Bihan:
The reason for this issue is currently `SSH session does not return and waits for the shim process to complete`. To test this hypothesis test is conducted as below:

**Testing Steps:**
Modified `_run_ssh_command()` to print testing logs as below:

```
def _run_ssh_command(hostname: str, ssh_private_key: str, command: str):
    with tempfile.NamedTemporaryFile("w+", 0o600) as f:
        f.write(ssh_private_key)
        f.flush()
        subprocess.run(
            [
                "ssh",
                "-F",
                "none",
                "-o",
                "StrictHostKeyChecking=no",
                "-i",
                f.name,
                f"hotaisle@{hostname}",
                command,
            ],
            stdout=subprocess.DEVNULL,
            stderr=subprocess.DEVNULL,
        )
        print(f"SSH command completed successfully to {hostname}: {command[:50]}...")
```

**Result: Summary Logs**

```
INFO     The dstack server (no version) is running at http://127.0.0.1:3000/                                                                                          
...
...                                                       
[12:19:06] DEBUG    dstack._internal.server.app:247 Processed request POST http://127.0.0.1:3000/api/project/bihan/runs/get in 0.013133s. Status: 200                           
[12:19:08] DEBUG    dstack._internal.server.background.tasks.process_runs:153 run(b275ff)amd-test: processing run                                                               
[...]

PR #2935, opened 2025-07-30
add hotaisle backend

Bihan:
Notes:
1. Only two instance types are available in Hotaisle with CPUs: 8-core and 13-core. Other fields are not configurable.
2. Custom HotAisleProvider is added in HotasileCompute because Hotaisle requires `api_key `and `team_handle`
3. **Issue:** Like [Lambda's Issue](https://github.com/dstackai/dstack/issues/2669), instance becomes ureachable after dstack server restart

### DIFF
--- a/docs/docs/concepts/backends.md
+++ b/docs/docs/concepts/backends.md
@@ -579,9 +579,9 @@ gcloud projects list --format="json(projectId)"
     Using private subnets assumes that both the `dstack` server and users can access the configured VPC's private subnets.
     Additionally, [Cloud NAT](https://cloud.google.com/nat/docs/overview) must be configured to provide access to external resources for provisioned instances.
 
-## Hotaisle
+## Hot Aisle
 
-Log in to the SSH TUI as described in the [Hotaisle Quick Start :material-arrow-top-right-thin:{ .external }](https://hotaisle.xyz/quick-start/).
+Log in to the SSH TUI as described in the [Hot Aisle Quick Start :material-arrow-top-right-thin:{ .external }](https://hotaisle.xyz/quick-start/).
 Create a new team and generate an API key for the member in the team.
 
 Then, go ahead and configure the backend:
@@ -600,6 +600,12 @@ projects:
 ```
 
 </div>
+
+??? info "Required permissions"
+    The API key must have the following roles assigned:
+
+    * **Owner role for the user** - Required for creating and managing SSH keys
+    * **Operator role for the team** - Required for managing virtual machines within the team
 
 ## Lambda
 
--- a/docs/docs/concepts/backends.md
+++ b/docs/docs/concepts/backends.md
@@ -579,6 +579,28 @@ gcloud projects list --format="json(projectId)"
     Using private subnets assumes that both the `dstack` server and users can access the configured VPC's private subnets.
     Additionally, [Cloud NAT](https://cloud.google.com/nat/docs/overview) must be configured to provide access to external resources for provisioned instances.
 
+## Hotaisle
+
+Log in to the SSH TUI as described in the [Hotaisle Quick Start :material-arrow-top-right-thin:{ .external }](https://hotaisle.xyz/quick-start/).
+Create a new team and generate an API key for the member in the team.
+
+Then, go ahead and configure the backend:
+
+<div editor-title="~/.dstack/server/config.yml">
+
+```yaml
+projects:
+- name: main
+  backends:
+    - type: hotaisle
+      team_handle: hotaisle-team-handle
+      creds:
+        type: api_key
+        api_key: 9c27a4bb7a8e472fae12ab34.3f2e3c1db75b9a0187fd2196c6b3e56d2b912e1c439ba08d89e7b6fcd4ef1d3f
+```
+
+</div>
+
 ## Lambda
 
 Log into your [Lambda Cloud :material-arrow-top-right-thin:{ .external }](https://lambdalabs.com/service/gpu-cloud) account, click API keys in the sidebar, and then click the `Generate API key`
--- a/src/dstack/_internal/core/backends/hotaisle/api_client.py
+++ b/src/dstack/_internal/core/backends/hotaisle/api_client.py
@@ -2,6 +2,7 @@ from typing import Any, Dict, Optional
 
 import requests
 
+from dstack._internal.core.backends.base.configurator import raise_invalid_credentials_error
 from dstack._internal.utils.logging import get_logger
 
 API_URL = "https://admin.hotaisle.app/api"
@@ -9,7 +10,7 @@ API_URL = "https://admin.hotaisle.app/api"
 logger = get_logger(__name__)
 
 
-class HotaisleAPIClient:
+class HotAisleAPIClient:
     def __init__(self, api_key: str, team_handle: str):
         self.api_key = api_key
         self.team_handle = team_handle
@@ -19,28 +20,42 @@ class HotaisleAPIClient:
             self._validate_user_and_team()
             return True
         except requests.HTTPError as e:
-            if e.response.status_code in [401, 403]:
-                return False
+            if e.response.status_code == 401:
+                raise_invalid_credentials_error(
+                    fields=[["creds", "api_key"]], details="Invalid API key"
+                )
+            elif e.response.status_code == 403:
+                raise_invalid_credentials_error(
+                    fields=[["creds", "api_key"]],
+                    details="Authenticated user does note have required permissions",
+                )
             raise e
-        except ValueError:
-            return False
+        except ValueError as e:
+            error_message = str(e)
+            if "No Hot Aisle teams found" in error_message:
+                raise_invalid_credentials_error(
+                    fields=[["creds", "api_key"]],
+                    details="Valid API key but no teams found for this user",
+                )
+            elif "not found" in error_message:
+                raise_invalid_credentials_error(
+                    fields=[["team_handle"]], details=f"Team handle '{self.team_handle}' not found"
+                )
+            raise e
 
     def _validate_user_and_team(self) -> None:
         url = f"{API_URL}/user/"
         response = self._make_request("GET", url)
-
-        if response.ok:
-            user_data = response.json()
-        else:
-            response.raise_for_status()
+        response.raise_for_status()
+        user_data = response.json()
 
         teams = user_data.get("teams", [])
         if not teams:
-            raise ValueError("No Hotaisle teams found for this user")
+            raise ValueError("No Hot Aisle teams found for this user")
 
         available_teams = [team["handle"] for team in teams]
         if self.team_handle not in available_teams:
-            raise ValueError(f"Hotaisle Team '{self.team_handle}' not found.")
+            raise ValueError(f"Hot Aisle team '{self.team_handle}' not found.")
 
     def upload_ssh_key(self, public_key: str) -> bool:
         url = f"{API_URL}/user/ssh_keys/"
@@ -50,8 +65,7 @@ class HotaisleAPIClient:
 
         if response.status_code == 409:
             return True  # Key already exists - success
-        if not response.ok:
-            response.raise_for_status()
+        response.raise_for_status()
         return True
 
     def create_virtual_machine(
@@ -59,38 +73,28 @@ class HotaisleAPIClient:
     ) -> Dict[str, Any]:
         url = f"{API_URL}/teams/{self.team_handle}/virtual_machines/"
         response = self._make_request("POST", url, json=vm_payload)
-
-        if not response.ok:
-            response.raise_for_status()
-
+        response.raise_for_status()
         vm_data = response.json()
         return vm_data
 
     def get_vm_state(self, vm_name: str) -> str:
         url = f"{API_URL}/teams/{self.team_handle}/virtual_machines/{vm_name}/state/"
         response = self._make_request("GET", url)
-
-        if not response.ok:
-            response.raise_for_status()
-
+        response.raise_for_status()
         state_data = response.json()
         return state_data["state"]
 
-    def terminate_virtual_machine(self, vm_name: str) -> bool:
+    def terminate_virtual_machine(self, vm_name: str) -> None:
         url = f"{API_URL}/teams/{self.team_handle}/virtual_machines/{vm_name}/"
         response = self._make_request("DELETE", url)
-
-        if response.status_code == 204:
-            return True
-        else:
-            response.raise_for_status()
+        response.raise_for_status()
 
     def _make_request(
         self, method: str, url: str, json: Optional[Dict[str, Any]] = None, timeout: int = 30
     ) -> requests.Response:
         headers = {
             "accept": "application/json",
-            "Authorization": self.api_key,
+            "Authorization": f"Token {self.api_key}",
         }
         if json is not None:
             headers["Content-Type"] = "application/json"
--- a/src/dstack/_internal/core/backends/hotaisle/api_client.py
+++ b/src/dstack/_internal/core/backends/hotaisle/api_client.py
@@ -1,0 +1,104 @@
+from typing import Any, Dict, Optional
+
+import requests
+
+from dstack._internal.utils.logging import get_logger
+
+API_URL = "https://admin.hotaisle.app/api"
+
+logger = get_logger(__name__)
+
+
+class HotaisleAPIClient:
+    def __init__(self, api_key: str, team_handle: str):
+        self.api_key = api_key
+        self.team_handle = team_handle
+
+    def validate_api_key(self) -> bool:
+        try:
+            self._validate_user_and_team()
+            return True
+        except requests.HTTPError as e:
+            if e.response.status_code in [401, 403]:
+                return False
+            raise e
+        except ValueError:
+            return False
+
+    def _validate_user_and_team(self) -> None:
+        url = f"{API_URL}/user/"
+        response = self._make_request("GET", url)
+
+        if response.ok:
+            user_data = response.json()
+        else:
+            response.raise_for_status()
+
+        teams = user_data.get("teams", [])
+        if not teams:
+            raise ValueError("No Hotaisle teams found for this user")
+
+        available_teams = [team["handle"] for team in teams]
+        if self.team_handle not in available_teams:
+            raise ValueError(f"Hotaisle Team '{self.team_handle}' not found.")
+
+    def upload_ssh_key(self, public_key: str) -> bool:
+        url = f"{API_URL}/user/ssh_keys/"
+        payload = {"authorized_key": public_key}
+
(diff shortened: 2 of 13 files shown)


Bihan:
**How to test**

**1. Add hotaisle creds**
```
type: hotaisle
    team_handle: <TEAM_HANDLE>
    creds:
      type: api_key
      api_key: <API_KEY>
```

**2. GPUHunt** 
Use https://github.com/Bihan/gpuhunt/tree/add_hotaisle_vm

**3. Serve Qwen**
```
type: service
name: qwen-vllm-amd

image: rocm/vllm:latest
env:
  - MODEL_ID=Qwen/Qwen2.5-7B-Instruct
  - MAX_MODEL_LEN=4096

commands:
  - vllm serve $MODEL_ID --max-model-len $MAX_MODEL_LEN

port: 8000
model: Qwen/Qwen2.5-7B-Instruct

resources:
  gpu: MI300X:1
```

src/dstack/_internal/core/backends/hotaisle/models.py
```diff
@@ -0,0 +1,45 @@
+from typing import Annotated, List, Literal, Optional, Union
+
+from pydantic import Field
+
+from dstack._internal.core.models.common import CoreModel
+
+
+class HotaisleAPIKeyCreds(CoreModel):
+    type: Annotated[Literal["api_key"], Field(description="The type of credentials")] = "api_key"
+    api_key: Annotated[str, Field(description="The Hotaisle API key")]
```
jvstme: I think the platform name is stylized as `Hot Aisle`: https://hotaisle.xyz/

I'd suggest to fix the name everywhere for consistency with their brand.

- Fix `Hotaisle` -> `Hot Aisle` in all strings, comments, and docs.

Similarly, fix the case in class names for consistency within our codebase. Currently, there is `HotAisleProvider` in `gpuhunt`, but `HotaisleBackend` in `dstack`.

- Fix `Hotaisle` -> `HotAisle` in all class names.

src/dstack/_internal/core/backends/hotaisle/configurator.py
```diff
@@ -0,0 +1,62 @@
...
+    def _get_config(self, record: BackendRecord) -> HotaisleConfig:
+        return HotaisleConfig.__response__(
+            **json.loads(record.config),
+            creds=HotaisleCreds.parse_raw(record.auth),
+        )
+
+    def _validate_creds(self, creds: AnyHotaisleCreds, team_handle: str):
+        api_client = HotaisleAPIClient(creds.api_key, team_handle)
+        if not api_client.validate_api_key():
+            raise_invalid_credentials_error(fields=[["creds", "api_key"]])
```
jvstme: `api_client.validate_api_key()` also validates the `team_handle`, not just the API key. So ideally:

- `fields` should be set to `[["team_handle"]]` if the handle doesn't pass the validation, e.g. if the specified team does not exist.
- We could pass a string to `raise_invalid_credentials_error(details=...)` to describe the error, e.g. invalid key, missing team, missing roles, etc.

This would help users troubleshoot backend configuration errors.

I think the easiest way to implement this is to call `raise_invalid_credentials_error` directly in `api_client.validate_api_key`

src/dstack/_internal/core/backends/hotaisle/compute.py
```diff
@@ -0,0 +1,213 @@
...
+    def get_payload_from_offer(self, instance_type) -> dict:
+        # Only two instance types are available in Hotaisle with CPUs: 8-core and 13-core. Other fields are
+        # not configurable.
+        cpu_cores = instance_type.resources.cpus
+        if cpu_cores == 8:
+            cpu_model = "Xeon Platinum 8462Y+"
+            frequency = 2800000000
+        else:  # cpu_cores == 13
+            cpu_model = "Xeon Platinum 8470"
+            frequency = 2000000000
```
jvstme: Hot Aisle can add more instance types in the future, which will break this code. For example, if they add a 30-CPU instance, this code will assume the CPU model is `Xeon Platinum 8470`, which is not necessarily the case.

Ideally, we should store these specs in the `gpuhunt` catalog item instead of hardcoding them in `dstack`. But that requires introducing backend-specific fields to `gpuhunt` catalog items, which is not that trivial, so I'd suggest not to do it in this PR.

As a simpler temporary solution, I can suggest to create a mapping from the instance type to any specs that need to be hardcoded.
  ```python
  INSTANCE_TYPE_SPECS = {
      "Xeon Platinum 8470-13c-224gb-1-mi300x": {
          "cpu_model": "Xeon Platinum 8470",
          "cpu_frequency": 2000000000,
          "cpu_manufacturer": "Intel"
      },
      ...
  }
  ```

If the instance type is not present in the mapping, skip the offer.

src/dstack/_internal/core/backends/hotaisle/compute.py
```diff
@@ -0,0 +1,213 @@
...
+                "cores": cpu_cores,
+                "frequency": frequency,
+            },
+            "disk_capacity": 13194139533312,
+            "ram_capacity": 240518168576,
+            "gpus": [
+                {
+                    "count": len(instance_type.resources.gpus),
+                    "manufacturer": "AMD",
+                    "model": "MI300X",
```
jvstme: Would it be possible to take `disk_capacity`, `ram_capacity`, `manufacturer`, and `model` from the offer instead of hardcoding them? I think Hot Aisle can change these parameters in the future, especially if adding more instance types

src/dstack/_internal/core/backends/hotaisle/compute.py
```diff
@@ -0,0 +1,213 @@
...
+            instance_id=vm_data["name"],
+            hostname=None,
+            internal_ip=None,
+            region=instance_offer.region,
+            price=instance_offer.price,
+            username="hotaisle",
+            ssh_port=22,
+            dockerized=True,
+            ssh_proxy=None,
+            backend_data=vm_data["ip_address"],
```
jvstme: We may need to store more properties in `backend_data` in the future, so consider storing a JSON object instead of just one IP address string. [Example](https://github.com/dstackai/dstack/blob/91659e7b68d7ded29b7c57658fce424fa383156d/src/dstack/_internal/core/backends/nebius/compute.py#L217)

src/dstack/_internal/core/backends/hotaisle/compute.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def _setup_instance(
+    hostname: str,
+    ssh_private_key: str,
+):
+    setup_commands = ("sudo apt-get update",)
+    _run_ssh_command(
+        hostname=hostname, ssh_private_key=ssh_private_key, command=" && ".join(setup_commands)
+    )
```
jvstme: Running `apt-get update` looks redundant, since we don't install any apt packages

src/dstack/_internal/core/backends/hotaisle/compute.py
```diff
@@ -0,0 +1,213 @@
...
+                },
+                daemon=True,
+            )
+            thread.start()
+
+    def terminate_instance(
+        self, instance_id: str, region: str, backend_data: Optional[str] = None
+    ):
+        vm_name = instance_id
+        self.api_client.terminate_virtual_machine(vm_name)
```
jvstme, line 169: Either this method or `api_client.terminate_virtual_machine` [should](https://github.com/dstackai/dstack/blob/b9ca0bedb4c7c62b2742fa385f17c3f596f6e3df/src/dstack/_internal/core/backends/base/compute.py#L99) ignore errors if the VM is already deleted. Currently, I'm getting repeated errors if I delete the VM manually via the TUI.

```
[12:47:03] WARNING  dstack._internal.server.background.tasks.process_instances:946 Failed to terminate instance horrible-squid-1-0. Will retry. Error: HTTPError('404 Client Error: Not Found  
                    for url: https://admin.hotaisle.app/api/teams/test/virtual_machines/enc1-gpuvm022/')                                                                                   
                      File "/dstack/src/dstack/_internal/server/background/tasks/process_instances.py", line 930, in _terminate                                         
                        await run_async(                                                                                                                                                       
                      File "/dstack/src/dstack/_internal/utils/common.py", line 21, in run_async                                                                        
                      File "/usr/lib64/python3.10/concurrent/futures/thread.py", line 58, in run                                                                                               
                      File "/dstack/src/dstack/_internal/core/backends/hotaisle/compute.py", line 153, in terminate_instance                                            
                        self.api_client.terminate_virtual_machine(vm_name)                                                                                                                     
                      File "/dstack/src/dstack/_internal/core/backends/hotaisle/api_client.py", line 86, in terminate_virtual_machine                                   
                        response.raise_for_status()                                                                                                                                            
                      File "/dstack/venv/lib64/python3.10/site-packages/requests/models.py", line 1024, in raise_for_status                                             
                        raise HTTPError(http_error_msg, response=self)                                                                                                                         
                    requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://admin.hotaisle.app/api/teams/test/virtual_machines/enc1-gpuvm022/
```

src/dstack/_internal/core/backends/hotaisle/api_client.py
```diff
@@ -0,0 +1,104 @@
...
+        self, vm_payload: Dict[str, Any], instance_name: str
+    ) -> Dict[str, Any]:
+        url = f"{API_URL}/teams/{self.team_handle}/virtual_machines/"
+        response = self._make_request("POST", url, json=vm_payload)
+
+        if not response.ok:
+            response.raise_for_status()
+
+        vm_data = response.json()
+        return vm_data
```
jvstme: The `instance_name` parameter is unused. If Hot Aisle doesn't allow setting the instance name, let's remove the parameter and anything related - instance name generation in `HotaisleCompute.create_instance` and the `MAX_INSTANCE_NAME_LEN` constant

src/dstack/_internal/core/backends/hotaisle/api_client.py
```diff
@@ -0,0 +1,104 @@
...
+            return False
+
+    def _validate_user_and_team(self) -> None:
+        url = f"{API_URL}/user/"
+        response = self._make_request("GET", url)
+
+        if response.ok:
+            user_data = response.json()
+        else:
+            response.raise_for_status()
```
jvstme: (nit) Here and in other methods below - the `if response.ok` check looks redundant. `raise_for_status()` does the same check internally and only raises an exception if the response status is 400+.

```suggestion
        response.raise_for_status()
        user_data = response.json()
```

src/dstack/_internal/core/backends/hotaisle/api_client.py
```diff
@@ -0,0 +1,104 @@
...
+        return state_data["state"]
+
+    def terminate_virtual_machine(self, vm_name: str) -> bool:
+        url = f"{API_URL}/teams/{self.team_handle}/virtual_machines/{vm_name}/"
+        response = self._make_request("DELETE", url)
+
+        if response.status_code == 204:
+            return True
+        else:
+            response.raise_for_status()
```
jvstme: The implementation doesn't match the `-> bool` return annotation - the method can return `None` if the response status is 2xx, except 204.

Considering the return value is not checked by the caller, I just wouldn't return anything.

```suggestion
    def terminate_virtual_machine(self, vm_name: str) -> None:
        url = f"{API_URL}/teams/{self.team_handle}/virtual_machines/{vm_name}/"
        response = self._make_request("DELETE", url)
        response.raise_for_status()
```

docs/docs/concepts/backends.md
```diff
@@ -579,6 +579,28 @@ gcloud projects list --format="json(projectId)"
     Using private subnets assumes that both the `dstack` server and users can access the configured VPC's private subnets.
     Additionally, [Cloud NAT](https://cloud.google.com/nat/docs/overview) must be configured to provide access to external resources for provisioned instances.
 
+## Hotaisle
+
+Log in to the SSH TUI as described in the [Hotaisle Quick Start :material-arrow-top-right-thin:{ .external }](https://hotaisle.xyz/quick-start/).
+Create a new team and generate an API key for the member in the team.
```
jvstme, line 585: Since Hot Aisle has a permission system, consider listing the roles that must be assigned to the API key.

From my testing, they are:
- The `owner` role for the user (for creating SSH keys)
- The `operator` role for the team (for managing VMs)

image

src/dstack/_internal/core/backends/hotaisle/api_client.py
```diff
@@ -0,0 +1,104 @@
...
+
+        teams = user_data.get("teams", [])
+        if not teams:
+            raise ValueError("No Hotaisle teams found for this user")
+
+        available_teams = [team["handle"] for team in teams]
+        if self.team_handle not in available_teams:
+            raise ValueError(f"Hotaisle Team '{self.team_handle}' not found.")
```
jvstme: (optional) This validation is already better than in most our backends, but we can further improve it by validating the roles assigned to the key, so that users can see permission-related errors earlier - when configuring the backend rather than when creating instances.

It should be possible to validate everything (the key, the user role, and the team roles) by calling only [`GET /user/api_keys/{prefix}/`](https://admin.hotaisle.app/api/docs/#/user/getUserAPIKeyDetailReq)

src/dstack/_internal/core/backends/hotaisle/api_client.py
```diff
@@ -0,0 +1,104 @@
...
+            return True
+        else:
+            response.raise_for_status()
+
+    def _make_request(
+        self, method: str, url: str, json: Optional[Dict[str, Any]] = None, timeout: int = 30
+    ) -> requests.Response:
+        headers = {
+            "accept": "application/json",
+            "Authorization": self.api_key,
```
jvstme: Same as in `gpuhunt`.
```suggestion
            "Authorization": f"Token {self.api_key}",
```

jvstme (commented):
@Bihan, sorry for the delayed review. The PR looks good overall, but please see my suggestions about some details

src/dstack/_internal/core/backends/hotaisle/models.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/configurator.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/compute.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/compute.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/compute.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/api_client.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/api_client.py
Bihan: Done

docs/docs/concepts/backends.md
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/api_client.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/api_client.py
Bihan: Thank you. Will plan to update it in the next iteration.

src/dstack/_internal/core/backends/hotaisle/compute.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/compute.py
```diff
@@ -211,3 +231,13 @@ def _run_ssh_command(hostname: str, ssh_private_key: str, command: str):
             stdout=subprocess.DEVNULL,
             stderr=subprocess.DEVNULL,
         )
+
+
+class HotAisleInstanceBackendData(CoreModel):
+    ip_address: str
+    vm_id: Optional[str] = None
```
jvstme, line 220: (nit) This field is unused, so I wouldn't include it in the model

src/dstack/_internal/core/backends/hotaisle/compute.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/api_client.py
```diff
@@ -19,28 +20,42 @@ def validate_api_key(self) -> bool:
...
+            error_message = str(e)
+            if "No Hot Aisle teams found" in error_message:
+                raise_invalid_credentials_error(
+                    fields=[["creds", "api_key"]],
+                    details="Valid API key but no teams found for this user",
+                )
+            elif "not found" in error_message:
+                raise_invalid_credentials_error(
+                    fields=[["team_handle"]], details=f"Team handle '{self.team_handle}' not found"
+                )
```
jvstme, line 43: (nit) Looking for patterns in our own error messages and then raising with another error message looks quite redundant. It's also error-prone, because we can change the error message in `_validate_user_and_team` and forget to change it here.

Some alternatives I can suggest:
- Raise with the same error message - `raise_invalid_credentials_error(details=str(e), ...)`
- Call `raise_invalid_credentials_error` directly in `_validate_user_and_team`
- (my favorite) Merge `validate_api_key` and `_validate_user_and_team` into one method and call `raise_invalid_credentials_error` directly

src/dstack/_internal/core/backends/hotaisle/api_client.py
Bihan: Done

src/dstack/_internal/core/backends/hotaisle/compute.py
```diff
@@ -0,0 +1,225 @@
...
+    InstanceConfiguration,
+    InstanceOfferWithAvailability,
+)
+from dstack._internal.core.models.placement import PlacementGroup
+from dstack._internal.core.models.runs import JobProvisioningData, Requirements
+from dstack._internal.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+MAX_INSTANCE_NAME_LEN = 60
```
jvstme, line 31: (nit) Unused

jvstme (approved):
@Bihan, thank you. I've left some nit-picking comments about the latest changes, but feel free to address them in a separate PR or ignore them, they are not that important. I think the PR is good to merge now

PR #2942, opened 2025-08-04
Fix/shim termination on dstack server restart

Bihan:
This PR fixes issue [#2669](https://github.com/dstackai/dstack/issues/2669), where the shim  launched via SSH gets terminated when the dstack server restarts.

**Issue Cause:**
dstack's local daemon thread creates an SSH connection to the VM. The shim installation command runs on the VM via this SSH connection, unlike cloud-init setup.Even though the shim runs on the VM, it's still a child of the SSH session.When dstack server restarts → daemon thread dies → SSH connection closes.When SSH session closes, the remote shell session ends, and any processes started by that session (including the shim) get terminated.

**Fix**
The shim launch_command is daemonized as `daemonized_command = f"{launch_command.rstrip('&')} >/tmp/dstack-shim.log 2>&1 & disown"`

### Note
- This branch is branch-out of the `hotaisle-backend` branch (PR #2935).

### DIFF
--- a/src/dstack/_internal/core/backends/configurators.py
+++ b/src/dstack/_internal/core/backends/configurators.py
@@ -55,6 +55,15 @@ except ImportError:
     pass
 
 try:
+    from dstack._internal.core.backends.hotaisle.configurator import (
+        HotaisleConfigurator,
+    )
+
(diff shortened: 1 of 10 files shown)


Bihan:
Will send a different PR removing Hotaisle Backend

PR #2952, opened 2025-08-06
Add NVIDIA GPU passive health checks

### LINKED ISSUE #2930: [Feature]: NVIDIA GPU passive health checks
### Problem

> Because of the gang scheduling semantics of ML jobs, failures have a large effect on the reliability of an entire job—a single failure of a system component can cause thousands of GPUs to sit idle. [...][T]he time between component failures may be small enough to be disruptive.

https://arxiv.org/html/2410.21680v1

### Solution

Leverage NVIDIA DCGM [Background Health Checks](https://docs.nvidia.com/datacenter/dcgm/latest/user-guide/feature-overview.html#background-health-checks).

### Workaround

_No response_

Yes

un-def:
* shim: Integrate libdcgm, add a new endpoint returning overall GPU health with list of incidents.
* Periodically pull instance health from shim, store the raw response in a new DB table. Infer overall instance health and store it in a new column of the "instances" table.
* Don't consider failed instances for submitted jobs. Note: instances with warnings are still considered for jobs.
* API: add a new method returning a list of instance health checks with unified structure.
* CLI: display "warning" and "failure" health statuses in the same way as "unreachable", below the instance status.

Closes: https://github.com/dstackai/dstack/issues/2930


src/dstack/_internal/core/backends/remote/provisioning.py
```diff
@@ -20,6 +20,7 @@
     Resources,
     SSHConnectionParams,
 )
+from dstack._internal.server.schemas.runner import HealthcheckResponse
```
r4victor: Importing from server to core should be avoided as a rule to keep import dependencies as simple as possible. remote/provisioning.py has no good reason to be in core, so if you really need to import from server there, I'd consider moving remote/provisioning.py to server services.

.github/workflows/build.yml
```diff
@@ -126,7 +126,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
```
r4victor, line 129: shim can no longer be compiled on macOS?

If so, maybe make a DCGMWrapper interface and provide two implementations: real implementation when go-dcgm works and dummy implementation when it doesn't and pick it at compile time via https://pkg.go.dev/cmd/go#hdr-Build_constraints.

r4victor (approved):

src/dstack/_internal/core/backends/remote/provisioning.py
un-def: Moved `provisioning.py` to `server/utils` and removed its dependency on `server/schemas`

.github/workflows/build.yml
un-def: Thanks for the idea, I'll bring back macOS support in a separate PR.

Issue #2851, opened 2025-06-27
[Feature] Support `repos` in run configurations (and make `dstack init` optional)


peterschmidt85:
**1. Introduce the `repos` property (similar to `files`)**  

```yaml
repos:
  - .:/workflow
```

or 

```yaml
repos:
  - .
```

or

```yaml
repos:
  - repo_path:  .
     path: .
     identity_key: ~/.ssh/id_rsa
```

or

```yaml
repos:
  - repo_path:  .
    identity_key: ~/.ssh/id_rsa
```

or

```yaml
repos:
  - repo_path:  .
    token: {env. GH_TOKEN}
```

> Note, if the repo is remote, `dstack` will try to use the default Git credentials (as it does now) using `~/.config/gh/hosts.yml` or `~/.ssh/id_rsa`.

A repo path can be also a GitHub URL (as it's supported now):

```
repos:
  - repo_path:  https://github.com/huggingface/open-r1
```

or 

```
repos:
  - https://github.com/huggingface/open-r1:/workflow
```

**2. Introduce `ssh_config.identity_file`**  
**3. Deprecate `dstack init`**

> [!WARNING]
> For the first version, let's allow up to one repo. Should help limit changes while addressing the main UX issue.

PR #2984, opened 2025-08-13
Deprecate local repos

### LINKED ISSUE #2974: Deprecate local repos
* Deprecate `--local` in `dstack init` and `dstack apply`, show a warning with a migration guide (*)
* Do not fall back to a local repo if no remote found
* For already initialized local repos show a warning with a migration guide (*)

(*) Something like "local repos are deprecated, use files instead: {configuration example}"

un-def:
* `--local` is deprecated
* `--ssh-identity` in `dstack init` is deprecated and ignored
* Repo is no longer required by default, but a warning is emitted to inform users about this change
* A new transitional `dstack init --remove` subcommand removes a local repo from `~/.dstack/config.yml`

[asciicast](https://asciinema.org/a/n8MyptYghecUZYmcx0oIynG85)

Closes: https://github.com/dstackai/dstack/issues/2974

### DIFF
--- a/docs/docs/concepts/repos.md
+++ b/docs/docs/concepts/repos.md
@@ -7,13 +7,13 @@ This allows accessing the directory files from within the run.
 ## Initialize a repo
 
 To use a directory with `dstack apply`, it must first be initialized as a repo by running [`dstack init`](../reference/cli/dstack/init.md).
-The directory can be either a regular local directory or a cloned Git repo.
+The directory must be a cloned Git repo.
 
 [`dstack init`](../reference/cli/dstack/init.md) is not required if you pass `-P` (or `--repo`) to [`dstack apply`](../reference/cli/dstack/apply.md) (see below).
 
 ### Git credentials
 
-If the directory is a cloned Git repo, [`dstack init`](../reference/cli/dstack/init.md) grants the `dstack` server access by uploading the current user's default
+[`dstack init`](../reference/cli/dstack/init.md) grants the `dstack` server access by uploading the current user's default
 Git credentials, ensuring that dstack can clone the Git repo when running the container.
 
 To use custom credentials, pass them directly with `--token` (GitHub token) or `--git-identity` (path to a private SSH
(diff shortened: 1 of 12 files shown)


peterschmidt85:
Sorry, just to confirm, will it be possible to specify `ssh-identity` via `dstack apply`?

un-def:
> Sorry, just to confirm, will it be possible to specify `ssh-identity` via `dstack apply`?

Sure

docs/docs/concepts/repos.md
r4victor, line 1: I think Repos docs should be rewritten to clarify that Repos are always git repos and contrast them with files. Currently, it's like the docs keep local repos in mind. Can be a separate PR.

docs/docs/reference/cli/dstack/apply.md
```diff
@@ -17,4 +17,9 @@ $ dstack apply --help
 
 </div>
 
+## User SSH key
+
+By default, `dstack` uses its own SSH key to access runs (`~/.dstack/ssh/id_rsa`).
```
r4victor:
```suggestion
By default, `dstack` uses its own SSH key to attach to runs (`~/.dstack/ssh/id_rsa`).
```

docs/docs/reference/cli/dstack/attach.md
```diff
@@ -13,4 +13,9 @@ $ dstack attach --help
 
 </div>
 
+## User SSH key
+
+By default, `dstack` uses its own SSH key to access runs (`~/.dstack/ssh/id_rsa`).
```
r4victor:
```suggestion
By default, `dstack` uses its own SSH key to attach to runs (`~/.dstack/ssh/id_rsa`).
```

r4victor (approved):

docs/docs/concepts/repos.md
un-def: The repo docs should definitely be revamped, I only removed/updated (now) incorrect information, docs update was an afterthought. I think we can improve the docs later, when https://github.com/dstackai/dstack/issues/2851 is implemented.

PR #3023, opened 2025-08-25
Add declarative repo configuration

un-def:
* `repos[].local_path,url` are equivalent to `--repo` option, the CLI option overrides the YAML configuration
* Legacy flow with implicitly loaded repos still supported but deprecated
* Repo info is no longer written to `config.yml`, but existing records are used for legacy flow
* `dstack init` is extended with `--repo`/`-P`, as in `dstack apply`

Updated docs and `repos[].path` support will be added in separate PRs.

Part-of: https://github.com/dstackai/dstack/issues/2851


r4victor (commented):
I suppose nothing happens if I specify configuration.repos in Python/HTTP API? That's fine, just pointing that configuration.repos can be stored on the server when submitted via HTTP API.

un-def:
> I suppose nothing happens if I specify configuration.repos in Python/HTTP API?

That's correct, Python API methods still require the `repo` argument, `repos` in the configuration is currently ignored.

> configuration.repos can be stored on the server when submitted via HTTP API.

If someone calls HTTP API directly, then yes. Our low-level API client excludes `repos` unconditionally for compatibility with old servers.

r4victor (approved):

Issue #2962, opened 2025-08-08
[Bug]: nebius.aio.service_error.RequestError: Request error DEADLINE_EXCEEDED: Deadline Exceeded


peterschmidt85:
### Steps to reproduce

Can happen on any Nebius API call, even terminating

### Actual behaviour

```
[13:43:07] DEBUG    dstack._internal.server.background.tasks.process_instances:930 Terminating runner instance
                    89.169.110.39
[13:43:18] WARNING  dstack._internal.server.background.tasks.process_instances:948 Failed to terminate instance cursor-0.
                    Will retry. Error: RequestError('Request error DEADLINE_EXCEEDED: Deadline Exceeded')
                    Traceback (most recent call last):
                      File "/Users/cheptsov/dstack/src/dstack/_internal/server/background/tasks/process_instances.py",
                    line 932, in _terminate
                        await run_async(
                        ...<4 lines>...
                        )
                      File "/Users/cheptsov/dstack/src/dstack/_internal/utils/common.py", line 21, in run_async
                        return await asyncio.get_running_loop().run_in_executor(None, func_with_args)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                      File
                    "/Users/cheptsov/.local/share/uv/python/cpython-3.13.3-macos-aarch64-none/lib/python3.13/concurrent/fu
                    tures/thread.py", line 59, in run
                        result = self.fn(*self.args, **self.kwargs)
                      File "/Users/cheptsov/dstack/src/dstack/_internal/core/backends/nebius/compute.py", line 237, in
                    terminate_instance
                        instance = resources.get_instance(self._sdk, instance_id)
                      File "/Users/cheptsov/dstack/src/dstack/_internal/core/backends/nebius/resources.py", line 325, in
                    get_instance
                        return LOOP.await_(
                               ~~~~~~~~~~~^
                            InstanceServiceClient(sdk).get(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                GetInstanceRequest(id=instance_id), timeout=REQUEST_TIMEOUT, metadata=REQUEST_MD
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                            )
                            ^
                        )
[...]

PR #3028, opened 2025-08-26
[Bug]: nebius.aio.service_error.RequestError: Request error DEADLINE_EXCEEDED: Deadline Exceeded #2962

peterschmidt85:
#2962

### DIFF
--- a/src/dstack/_internal/core/backends/nebius/resources.py
+++ b/src/dstack/_internal/core/backends/nebius/resources.py
@@ -110,11 +110,11 @@ def make_sdk(creds: NebiusServiceAccountCreds) -> SDK:
 
 def wait_for_operation(
     op: SDKOperation[Operation],
-    per_retry_timeout: float,
+    timeout: float,
     interval: float = 1,
 ) -> None:
     # Re-implementation of SDKOperation.wait() to avoid https://github.com/nebius/pysdk/issues/74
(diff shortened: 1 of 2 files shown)


src/dstack/_internal/core/backends/nebius/resources.py
```diff
@@ -110,16 +110,16 @@ def make_sdk(creds: NebiusServiceAccountCreds) -> SDK:
 
 def wait_for_operation(
     op: SDKOperation[Operation],
-    timeout: float,
+    per_retry_timeout: float,
     interval: float = 1,
 ) -> None:
     # Re-implementation of SDKOperation.wait() to avoid https://github.com/nebius/pysdk/issues/74
-    deadline = time.monotonic() + timeout
+    deadline = time.monotonic() + per_retry_timeout
```
jvstme: The name `per_retry_timeout` is off here. This parameter *is* a global timeout, not a per-retry timeout. It is used to calculate the global deadline for the operation, line 117.

The old name `timeout` was correct.

jvstme (approved):
This PR suggests to use the same constant timeouts, but set them as per-retry timeouts instead of per-method timeouts. The SDK defaults to 3 retries per method, which means that each method can now block for 3 times longer than before.

I don't think API calls are supposed to take that long, but I don't mind merging this PR if it helps.

PR #3030, opened 2025-08-26
Add digital ocean and amd dev backend

Bihan:
Steps
1. Login to both clouds https://cloud.digitalocean.com/login and https://amd.digitalocean.com/login

2. To configure backends add

```
  - type: amddevcloud
    project_name: TestProject
    creds:
      type: api_key
      api_key: <api_key_amd_dev_cloud>
  - type: digitalocean
    project_name: TestProjectNvidia
    creds:
      type: api_key
      api_key: <api_key_digitalocean>
```

Note: Will be updating the PR with **project_name** configuration. Creation of Droplet (VM) was not documented in api reference, but recently found that "project_id" field works.


src/dstack/_internal/core/backends/models.py
```diff
@@ -122,6 +129,7 @@
     CloudRiftBackendConfigWithCreds,
     CudoBackendConfigWithCreds,
     DataCrunchBackendConfigWithCreds,
+    BaseDigitalOceanBackendFileConfigWithCreds,
```
jvstme: (nit) You can drop this duplicate class and use `BaseDigitalOceanBackendConfigWithCreds` here instead.

The point of having both `*ConfigWithCreds` and `*FileConfigWithCreds` classes is to allow different configuration options depending on whether the backend is configured via `~/.dstack/server/config.yml` or via the API. Most backends don't need that and only define `*ConfigWithCreds`.

src/dstack/_internal/core/backends/digitalocean_base/models.py
```diff
@@ -0,0 +1,47 @@
...
+
+class BaseDigitalOceanBackendConfig(CoreModel):
+    type: Annotated[
+        Literal["amddevcloud", "digitalocean"],
+        Field(description="The type of backend"),
+    ] = "digitalocean"
+    project_name: Annotated[str, Field(description="The name of the DigitalOcean project")] = None
+    regions: Annotated[
+        Optional[List[str]],
+        Field(description="The list of DigitalOcean regions. Omit to use all regions"),
```
jvstme: (nit) I wouldn't mention DigitalOcean in the field descriptions so as not to confuse AMD Developer Cloud users

src/dstack/_internal/core/backends/digitalocean_base/compute.py
```diff
@@ -0,0 +1,176 @@
...
+        if not instance_offer.instance.resources.gpus:
+            # No GPUs, use CPU image
+            return "ubuntu-24-04-x64"
+
+        gpu_count = len(instance_offer.instance.resources.gpus)
+        gpu_name = instance_offer.instance.resources.gpus[0].name
+
+        if gpu_name == "MI300X":
+            # AMD GPU
+            return "digitaloceanai-rocmjupyter"
```
jvstme: `MI300X` isn't the only AMD GPU. It's safer to check the GPU vendor instead - `instance_offer.instance.resources.gpus[0].vendor`

src/dstack/_internal/core/backends/digitalocean_base/compute.py
```diff
@@ -0,0 +1,176 @@
...
+    ) -> JobProvisioningData:
+        instance_name = generate_unique_instance_name(
+            instance_config, max_length=MAX_INSTANCE_NAME_LEN
+        )
+
+        project_ssh_key = instance_config.ssh_keys[0]
+        ssh_key_id = self.api_client.get_or_create_ssh_key(
+            name=f"dstack-{instance_config.project_name}",
+            public_key=project_ssh_key.public,
+        )
```
jvstme, line 91: Is it necessary to upload the key via the API? The key is already embedded in the cloud-init user data

src/dstack/_internal/core/backends/digitalocean_base/api_client.py
```diff
@@ -0,0 +1,97 @@
...
+            logger.debug("DigitalOcean droplet %s not found", droplet_id)
+            return
+        response.raise_for_status()
+
+    def _make_request(
+        self, method: str, endpoint: str, json: Optional[Dict[str, Any]] = None, timeout: int = 30
+    ) -> requests.Response:
+        url = f"{self.base_url}{endpoint}"
+        headers = {
+            "Content-Type": "application/json",
```
jvstme: (nit) This header is set automatically when passing `json=` to `requests.request`

src/dstack/_internal/core/backends/digitalocean_base/compute.py
```diff
@@ -0,0 +1,176 @@
...
+            "ssh_keys": [ssh_key_id],
+            "backups": False,
+            "ipv6": False,
+            "monitoring": False,
+            "tags": [],
+            "user_data": get_user_data(
+                authorized_keys=instance_config.get_public_keys(),
+                backend_specific_commands=backend_specific_commands,
+            ),
+            **({"project_id": project_id} if project_id is not None else {}),
```
jvstme, line 118: To my understanding, `project_id=None` means the user misconfigured the backend and we couldn't find the project by name. Such configuration errors should be reported to the user and not ignored silently.

Ideally, this should also be validated in the backend configurator, so the user sees the error when creating the backend, not when attempting to create instances

jvstme (commented):
@Bihan, thanks for the PR. The division into two backends looks good to me 👍 Please also see my comments about some implementation details

src/dstack/_internal/core/backends/digitalocean_base/compute.py
Bihan: @jvstme 
> To my understanding, project_id=None means the user misconfigured the backend and we couldn't find the project by name.

Not necessarily. When `project_id=None`, DigitalOcean simply creates the instance in the `default project`. The `project_name` configuration is optional, so if it's not provided or doesn't match an existing project, the default project is used. This is expected behavior and not necessarily a misconfiguration.

src/dstack/_internal/core/backends/digitalocean_base/models.py
```diff
@@ -0,0 +1,47 @@
...
+AnyBaseDigitalOceanCreds = BaseDigitalOceanAPIKeyCreds
+BaseDigitalOceanCreds = AnyBaseDigitalOceanCreds
+
+
+class BaseDigitalOceanBackendConfig(CoreModel):
+    type: Annotated[
+        Literal["amddevcloud", "digitalocean"],
+        Field(description="The type of backend"),
+    ] = "digitalocean"
+    project_name: Annotated[str, Field(description="The name of the DigitalOcean project")] = None
```
jvstme: If `project_name` is optional, the annotation should be `Optional[str]`.

Also consider describing what a missing `project_name` means in the description

src/dstack/_internal/core/backends/digitalocean_base/compute.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+        size_slug = instance_offer.instance.name
+
+        if not instance_offer.instance.resources.gpus:
+            backend_specific_commands = SETUP_COMMANDS + DOCKER_INSTALL_COMMANDS
+        else:
+            backend_specific_commands = SETUP_COMMANDS
+
+        project_id = self.api_client.get_project_id(self.config.project_name)
```
jvstme: Avoid a redundant API call by not calling `get_project_id` if `project_name` is `None`

src/dstack/_internal/core/backends/digitalocean_base/compute.py
jvstme: > The `project_name` configuration is optional, so if it's not provided [...], the default project is used

Oh, I missed that it's optional. Then yes, it makes sense to use the default project if `project_name` is not set.

> or doesn't match an existing project

This doesn't make sense to me though. I can't see why a user would want to specify a project that doesn't exist. This clearly indicates a misconfiguration and should be reported to the user.

For example, if the user made a typo in the project name, they'd probably prefer an error message rather than  VMs being provisioned in a completely different project.

src/dstack/_internal/core/backends/models.py
Bihan: Done. Yes that was my typo.

src/dstack/_internal/core/backends/digitalocean_base/models.py
Bihan: Done

src/dstack/_internal/core/backends/digitalocean_base/compute.py
Bihan: > For example, if the user made a typo in the project name, they'd probably prefer an error message rather than VMs being provisioned in a completely different project.

Yes. I like this. Will log error message.

src/dstack/_internal/core/backends/digitalocean_base/compute.py
Bihan: Done

src/dstack/_internal/core/backends/digitalocean_base/compute.py
Bihan: @jvstme It's not strictly necessary to upload the SSH key via the API, but the key must exist in the user's account—either uploaded manually through the console or programmatically via the API. If the SSH key isn't registered with DigitalOcean, the Droplet creation API will fail with an error like:

`b8:fc:77:f2:87:51:0a:5d:fc:39:1b:6e:45:b7:f3:15 are invalid key identifiers for Droplet creation.`

src/dstack/_internal/core/backends/digitalocean_base/api_client.py
Bihan: Done

src/dstack/_internal/core/backends/digitalocean_base/compute.py
jvstme: I assume this error only occurs if you reference the key in [`ssh_keys`](https://github.com/Bihan/dstack/blob/1393761cbe33791af7b9642949619e4ca5862418/src/dstack/_internal/core/backends/digitalocean_base/compute.py#L105). I tried removing `ssh_keys`, and the droplet is created without uploading a key to the account.

However, if `ssh_keys` is not set, DigitalOcean sets the root password and emails it to the user. And `dstack` cannot connect to the droplet without the password, even though the project SSH key is actually installed via cloud-init. So I guess we'll have to keep using `ssh_keys` and uploading the key via the API, as the current implementation suggests

src/dstack/_internal/core/backends/digitalocean_base/compute.py
jvstme: Thanks. In `create_instance`, you can raise `BackendError` and it will be logged by dstack. And use `raise_invalid_credentials_error` in the backend configurator, so that the error is propagated to the user at backend configuration time

Bihan: Done

src/dstack/_internal/core/backends/digitalocean_base/models.py
Bihan: Done

src/dstack/_internal/core/backends/digitalocean_base/compute.py
Bihan: Done

src/dstack/_internal/core/backends/digitalocean_base/api_client.py
```diff
@@ -27,6 +27,14 @@ def validate_api_key(self) -> bool:
                 )
             raise e
 
+    def validate_project_name(self, project_name: str) -> bool:
+        if self.get_project_id(project_name) is None:
+            raise_invalid_credentials_error(
+                fields=[["config", "project_name"]],
```
jvstme: (nit) Not that it affects anything, but I think `fields` are supposed to be relative to the backend config root. And `project_name` is a root setting — it's `BaseDigitalOceanBackendConfigWithCreds.project_name`, not `BaseDigitalOceanBackendConfigWithCreds.config.project_name`

```suggestion
                fields=[["project_name"]],
```

src/dstack/_internal/core/backends/digitalocean_base/compute.py
```diff
@@ -155,9 +158,9 @@ def _get_image_for_instance(self, instance_offer: InstanceOfferWithAvailability)
             return "ubuntu-24-04-x64"
 
         gpu_count = len(instance_offer.instance.resources.gpus)
-        gpu_name = instance_offer.instance.resources.gpus[0].name
+        gpu_vendor = instance_offer.instance.resources.gpus[0].vendor
 
-        if gpu_name == "MI300X":
+        if gpu_vendor == "amd":
```
jvstme: (nit) Use the enum variant instead.

```suggestion
        if gpu_vendor == gpuhunt.AcceleratorVendor.AMD:
```

This ensures that the line will keep working properly even if `AcceleratorVendor` values change

src/dstack/_internal/core/backends/digitalocean_base/compute.py
```diff
@@ -88,15 +88,18 @@ def create_instance(
...
             backend_specific_commands = SETUP_COMMANDS + DOCKER_INSTALL_COMMANDS
         else:
             backend_specific_commands = SETUP_COMMANDS
 
-        project_id = self.api_client.get_project_id(self.config.project_name)
+        project_id = (
+            self.api_client.get_project_id(self.config.project_name)
+            if self.config.project_name
+            else None
+        )
```
jvstme: It is still worth checking that `get_project_id` didn't return `None`. This would mean that `project_name` no longer exists, in which case provisioning should fail, not be redirected to the default project.

```suggestion
        project_id = None
        if self.config.project_name:
            project_id = self.api_client.get_project_id(self.config.project_name)
            if project_id is None:
                raise BackendError(f"Project {self.config.project_name} does not exist")
```

src/dstack/_internal/core/backends/digitalocean_base/models.py
```diff
@@ -0,0 +1,47 @@
...
+    creds: Annotated[AnyBaseDigitalOceanCreds, Field(description="The credentials")]
+
+
+AnyBaseDigitalOceanBackendConfig = Union[
+    BaseDigitalOceanBackendConfig, BaseDigitalOceanBackendConfigWithCreds
+]
+
+
+class BaseDigitalOceanBackendFileConfigWithCreds(BaseDigitalOceanBackendConfig):
+    creds: Annotated[AnyBaseDigitalOceanCreds, Field(description="The credentials")]
```
jvstme: (nit) Unused since the latest commit

```suggestion
```

src/dstack/_internal/core/backends/digitalocean_base/models.py
```diff
@@ -0,0 +1,47 @@
...
+
+AnyBaseDigitalOceanCreds = BaseDigitalOceanAPIKeyCreds
+BaseDigitalOceanCreds = AnyBaseDigitalOceanCreds
+
+
+class BaseDigitalOceanBackendConfig(CoreModel):
+    type: Annotated[
+        Literal["amddevcloud", "digitalocean"],
+        Field(description="The type of backend"),
+    ] = "digitalocean"
```
jvstme: The reference is generated incorrectly - it says that `type` must be set to `digitalocean` for the `amddevcloud` backend.

reference


We may need to look into updating the reference generating script, `gen_schema_reference.py`, but the easiest way to fix this now is to remove the default.

```suggestion
    type: Annotated[
        Literal["amddevcloud", "digitalocean"],
        Field(description="The type of backend"),
    ]
```

jvstme (approved):
@Bihan, thanks for the update. I think we are almost ready to merge. The remaining things are:

- [x] Bump `gpuhunt` to `0.1.8` in `pyproject.toml`
- [x] Address my non-nitpicking comments on the latest update
- [x] Optionally address my nitpicking comments

If you decide to follow my code suggestions and there will be no details to discuss, feel free to merge without repeating the review. Thank you!

PR #3076, opened 2025-09-04
[CLI] Handle unrecognized arguments

### LINKED ISSUE #3073: [Bug]: CLI swallows unknown arguments
### Steps to reproduce

* ```
  dstack init --git-token gh_token
  ```
  `dstack` ignores `--git-token` (the correct option is `--token`), which may be repo scoped and read only, and falls back to, e.g., user's default SSH key, which gives the full access to all repos
* ```
  dstack apply --ruse
  ```
  `dstack` ignores `--ruse` (the user mistyped `--reuse`) and may spawn a new instance

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### dstack version

master

### Server logs

```shell

```

### Additional information

_No response_

un-def:
* All commands now reject unrecognized arguments
* Undocumented `${{ run.args }}` for tasks and services is still supported but requires `--` pseudo-argument:

  > If you have positional arguments that must begin with `-`
  > and don’t look like negative numbers, you can insert
  > the pseudo-argument `'--'` which tells `parse_args()`
  > that everything after that is a positional argument

  ```
  dstack apply --reuse -- --some=arg --some-option
                       ^^
  ```

Fixes: https://github.com/dstackai/dstack/issues/3073

---

@peterschmidt85 not sure if we want to drop `${{ run.args }}` altogether or document it as a feature.

[asciicast](https://asciinema.org/a/tEPx2ICbanEEUWR0SIPZLb7iI)

### DIFF
--- a/src/dstack/_internal/cli/commands/__init__.py
+++ b/src/dstack/_internal/cli/commands/__init__.py
@@ -1,20 +1,22 @@
 import argparse
 import os
+import shlex
 from abc import ABC, abstractmethod
-from typing import List, Optional
+from typing import ClassVar, Optional
 
 from rich_argparse import RichHelpFormatter
 
 from dstack._internal.cli.services.completion import ProjectNameCompleter
-from dstack._internal.cli.utils.common import configure_logging
+from dstack._internal.core.errors import CLIError
 from dstack.api import Client
 
 
 class BaseCommand(ABC):
-    NAME: str = "name the command"
-    DESCRIPTION: str = "describe the command"
-    DEFAULT_HELP: bool = True
-    ALIASES: Optional[List[str]] = None
+    NAME: ClassVar[str] = "name the command"
+    DESCRIPTION: ClassVar[str] = "describe the command"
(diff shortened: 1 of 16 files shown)


r4victor (approved):

peterschmidt85:
@un-def 
Was just testing `dstack init` and got an indirectly related question.

```
dstack init --help
Usage: dstack init [-h] [--project PROJECT] [-P REPO] [-t OAUTH_TOKEN] [--git-identity SSH_PRIVATE_KEY]

Options:
  -h, --help            Show this help message and exit
  --project PROJECT     The name of the project
  -P, --repo REPO       The repo to initialize. Can be a local path or a Git repo URL. Defaults to the current working directory.
  -t, --token OAUTH_TOKEN
                        An authentication token to access a private Git repo
  --git-identity SSH_PRIVATE_KEY
                        The private SSH key path to access a private Git repo
```

The command for some reason doesn't show `--ssh-identity` and `--remove` but respects and supports them. I wonder they aren't in the help message.

peterschmidt85 (approved):

Issue #3021, opened 2025-08-25
Implement backend-specific offers cache


r4victor:
Currently, offers cache is implemented via Compute.get_offers_cached(requirements), which caches backend offers given requirements. Since cache is not reused for different requirements, it's expensive to get offers for every existing fleet when provisioning a run (#3020). The solution is to make the offers cache backend-specific:

* We will call get_offers(requirements) for each fleet. Offer caching will not be shared at the get_offers level, but will be an implementation detail of the backend.
* Backends where offers do not depend on requirements will have a similar implementation: collect all offers and cache them, then apply the common offer filtering logic.
* Backends where offers depend on requirements will be able to cache not the dstack offers (since offers cannot be formed without requirements), but the API requests from which offers will later be formed using the requirements.

Essentially, the idea is to cache only the expensive operations (API calls) and allow doing fast operations (requirements filtering) many times.

I also suggest we move offers collection and filtering logic from gpuhunt to dstack as a part of this issue.

r4victor:
#3091 implemented requirements-independent offers cache on the dstack side for all backends expect cudo and vastai. Cudo and vastai require caching provider's API calls, which currently can be done in gpuhunt only. Until implemented, requesting cudo and vastai offers will be less efficient. We can postpone the implementation and do it when moving offer collection logic from gpuhunt to dstack.

Issue #3020, opened 2025-08-25
Consider backend offers when choosing optimal fleet for run


r4victor:
Current logic for selecting optimal fleet for the run (#2967) considers fleet existing instances. If fleets don't have enough instances and provisioning is required, then it's necessary to consider backend offers that can be provisioning with the fleet because fleet spec is combined with run spec to get requirements (#2969). To choose an optimal fleet, we need to get offers for every candidate fleet. This is currently expensive since offers cache is not reused for different requirements.

If this is implemented efficiently, we may also consider respecting fleet spec when returning offers in run plan, e.g. by returning offers list for every candidate fleet.

PR #3091, opened 2025-09-11
Implement requirements-independent offers cache

r4victor:
Part of #3021 

This PR reworks how dstack caches backend offers. Common `get_offers_cached()` is dropped. For most backends for which offers do not depend on requirements (or offers can be adjusted as in case with disk), it introduces `ComputeWithAllOffersCached` class that implements caching of all offers with availability and requirements post-filtering. This allows reusing cache when getting offers for different requirements. For cudo, vastai, kubernetes, the old behavior is preserved via `ComputeWithFilteredOffersCached`.

Upsides:
* Cache is reused frequently as it's requirements-independent. `dstack offer` and `dstack apply` become quick even when called with different requirements consequently.
* It's now possible to implement fleet selection considering backend offers for each fleet (#3020).

Downsides:
* When cache is cold, requesting offers may be slower since backend salways collect all offers. In my testing, I observed `time dstack apply -b aws --gpu H100` goes from ~15s to ~20s, `time dstack apply -b gcp --gpu H100` goes from ~10s to ~15s.

This implementation is for transitioning period. The plan is to experiment with fleet selection logic considering fleet offers, and see if it works well. Then, implement requirements-independent cache for cudo, vastai (requires caching provider's API calls, which currently can be done in gpuhunt only). Then move offers collection into dstack and simplify the code.

Tested offers for all backends except for cloudrift.


src/dstack/_internal/core/backends/template/compute.py.jinja
```diff
@@ -28,6 +29,7 @@ logger = get_logger(__name__)
 
 class {{ backend_name }}Compute(
     # TODO: Choose ComputeWith* classes to extend and implement
+    # ComputeWithAllOffersCached,
```
jvstme, line 32: (nit) I think `ComputeWithFilteredOffersCached` could be a good default for the template, since it works for all backends. Guidance on how to choose between `ComputeWithAllOffersCached` and `ComputeWithFilteredOffersCached` could also be helpful, unless we expect to move away from this model before anyone contributes the next backend.

jvstme (approved):

src/dstack/_internal/core/backends/template/compute.py.jinja
r4victor: My idea is to drop ComputeWithFilteredOffersCached soon – all providers should implement requirements-independent cache.

PR #3094, opened 2025-09-11
[TensorDock] Remove the `tensordock` from supported backeds #3092

### LINKED ISSUE #3092: [TensorDock] Remove the `tensordock` from supported backeds

peterschmidt85:


r4victor:
Ideally, show a warning if tensordock is configured so that it can be dropped completely in the future.

Also leave a comment in models.py that it's to be removed eventually.

r4victor (approved):

peterschmidt85:
> Ideally, show a warning if tensordock is configured so that it can be dropped completely in the future.

I thought the following is enough, no?

```
WARNING  dstack._internal.server.services.config:151 Backend tensordock not available and won't be configured. Check that backend dependencies are installed.
```

peterschmidt85:
> Also leave a comment in models.py that it's to be removed eventually.

Done

r4victor:
> > Ideally, show a warning if tensordock is configured so that it can be dropped completely in the future.
> 
> I thought the following is enough, no?
> 
> ```
> WARNING  dstack._internal.server.services.config:151 Backend tensordock not available and won't be configured. Check that backend dependencies are installed.
> ```

A bit confusing recommending "Check that backend dependencies are installed", but ok. We're dropping tensordock support without prior deprecations anyway.

PR #3097, opened 2025-09-12
[Internal] Project config support

peterschmidt85:
This PR extends project creation service with a customizable config that can be used in hooks.


src/dstack/_internal/server/services/projects.py
```diff
@@ -115,11 +116,20 @@ async def get_project_by_name(
     return project_model_to_project(project_model)
 
 
+class ProjectConfig(CoreModel):
```
r4victor: ProjectHookConfig to avoid confusion with other ProjectConfig?

r4victor (approved):

Issue #3088, opened 2025-09-10
[Feature]: GCP A4 instances


jvstme:
### Problem

GCP provides spot A4 instances with the NVIDIA B200 GPU. They are not yet supported in dstack.

### Solution

Support GCP spot A4 instances, including A4 clusters.

### Would you like to help us implement this feature by sending a PR?

Yes

r4victor:
@jvstme anything TBD on this issue or it can be closed?

PR #3099, opened 2025-09-13
Update NVIDIA driver in dstack OS images

jvstme:
Currently, `aws`, `gcp`, `azure`, and `oci` backends use our custom OS images. These images use 535 version of the CUDA driver. This version doesn't support newer generations of GPUs such as NVIDIA B200. 

This PR updates the scripts that build these custom OS images to update the CUDA version from 535 to 570.

This PR is required for #3100 (issue #3088).

Scope:

- Update from the 535 to the 570 family.
- Update to Ubuntu 24.04, since Ubuntu 22.04 does not have the `gcc` version required for building the 570 driver.
- Switch from proprietary to open kernel modules.
- Since pre-Turing GPUs aren't supported by NVIDIA open kernel modules, conditionally choose between old and new dstack OS images based on the GPU name.
- Adjust handling `apt` race conditions - the existing hack did not work on OCI's Ubuntu 24.04.
- Install `ufw` when building the image - it is missing in OCI's Ubuntu 24.04.

Notes: 

- This PR doesn't update `azure`'s [Grid drivers](https://github.com/dstackai/dstack/blob/4220247853206b8dfd661bdf277e2ffd9ce82f21/scripts/packer/provisioners/install-nvidia-grid-driver-for-azure.sh#L11) (used for A10).
- This PR doesn't update the `gcp`'s A3 OS [image script](https://github.com/dstackai/dstack/blob/update_nvidia_driver_in_base_images/.github/workflows/gcp-a3mega-image.yml)

Before/upon merging:
- [x] Build the 0.11rc2 images
- [x] Bump `base_image` in `version.py`
- [x] Test the relevant backends, especially Azure, which was only partially tested due to limitations of its staging images (see below)


r4victor (approved):

peterschmidt85:
Building 0.11rc2 via https://github.com/dstackai/dstack/actions/runs/17939443790

src/dstack/_internal/core/backends/aws/resources.py
```diff
@@ -35,9 +37,12 @@ def get_image_id_and_username(
         image_owner = DLAMI_OWNER_ACCOUNT_ID
         username = "ubuntu"
     else:
-        image_name = (
-            f"dstack-{version.base_image}" if not cuda else f"dstack-cuda-{version.base_image}"
-        )
+        if gpu_name is None:
```
peterschmidt85, line 40: Regarding AWS, just to confirm, the new image is only required for very few GPU types note covered by AWS DLAMI (e.g. T4),  right?

peterschmidt85 (commented):
Regarding Azure:

Not that it's a problem but for the Grid image (e.g., A10:4GB), we are still hard-coding the old CUDA version (550):https://github.com/dstackai/dstack/blob/4220247853206b8dfd661bdf277e2ffd9ce82f21/scripts/packer/provisioners/install-nvidia-grid-driver-for-azure.sh#L11

peterschmidt85 (approved):
After building 0.11rc2, I've tested:

1. AWS (T4) - all works
2. GCP (T4, L4) - all works
3. OCI (A10) - all works
4. Azure (T4, A10:4GB) - all works

Also, tested AWS (L4, L40S) but it used AWS DLAMI.

PR #3100, opened 2025-09-13
Support A4 instances with the B200 GPU on GCP

jvstme:
This implementation allows provisioning both individual A4 instances and clusters, but clusters do not yet support high-speed networking, since it requires a [different network setup](https://cloud.google.com/ai-hypercomputer/docs/create/create-vm#setup-network).

> [!NOTE]
> Do not merge before #3099.
> Also depends on https://github.com/dstackai/gpuhunt/pull/176 for testing.

#3088

### DIFF
--- a/src/dstack/_internal/core/backends/base/offers.py
+++ b/src/dstack/_internal/core/backends/base/offers.py
@@ -20,6 +20,7 @@ from dstack._internal.core.models.runs import Requirements
 SUPPORTED_GPUHUNT_FLAGS = [
     "oci-spot",
     "lambda-arm",
+    "gcp-a4",
 ]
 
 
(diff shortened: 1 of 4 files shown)


r4victor (approved):

peterschmidt85:
I've tested it using by changing `base_image` in `version.py` to `0.11rc2` (which is now used in `master`).

Just for the note: it makes only spot B200 instances work. Does it mean B200 are not available on-demand?

jvstme:
> Does it mean B200 are not available on-demand?

[Yes](https://cloud.google.com/compute/docs/accelerator-optimized-machines#consumption_option_availability_by_machine_type)

Issue #2562, opened 2025-04-24
[Feature]: Allow to use SSH via AWS Session Manager


peterschmidt85:
### Problem

Currently, both `dstack server` and `dstack` CLI require direct SSH access to VMs. 

Similar to #2554 for GCP, AWS recommends [using SSH via Session Manager](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html) to enforce the best security practices.

### Solution

1. Allow to enable use of Session Manager with the `aws` backend` configuration. This will require the `dstack server` to have `ssm:*` IAM permissions (e.g. to start sessions)
2. Allow the `dstack` CLI client users to use locally configured AWS credentials to start sessions via the SSH config.

### Would you like to help us implement this feature by sending a PR?

No

oborchers:
+1

Issue #2554, opened 2025-04-23
[Feature]: Support IAP tunneling when connecting to GCP VMs


colinjc:
### Problem

To follow security best practices we need to lock down SSH access to GCP VMs to only IAP authorized users. To do this, we can limit the source range to the IAP range, but dstack doesn't use the IAP tunnel. 

### Solution

When connecting to a GCP VMs allow tunnelling through IAP. https://cloud.google.com/compute/docs/connect/ssh-using-iap#gcloud

### Would you like to help us implement this feature by sending a PR?

No

PR #3123, opened 2025-09-19
Support client-side proxies for apply command

mrkn:
This is a concept implementation to support SSH proxy in `dstack apply` comand.
I think this can be a part of #2562 and #2554.

## Summary
- Add client-side SSH proxy configuration to dev/task/service configs
- Let `dstack apply` reuse those settings for ProxyJump, ProxyCommand, and AWS SSM

## Liimitations
- `dstack attach` still bypasses the new proxy controls, so detached runs cannot yet use them

## Next Steps
- Add cloud-specific helpers (e.g., GCP, Azure) on top of this structure

## Example configuration

```yaml
type: dev-environment
name: example
ide: vscode
attach:
  proxy:
    type: aws-ssm
    profile: sso-profile-name
```

### DIFF
--- a/src/dstack/_internal/cli/services/configurators/run.py
+++ b/src/dstack/_internal/cli/services/configurators/run.py
@@ -251,8 +251,10 @@ class BaseRunConfigurator(
                     )
                     # Map the attach.proxy settings to the original configuration
                     attach_proxy_config = SSHProxyConfig()
-                    if isinstance(conf, RunAttachConfiguration) and conf.attach is not None:
-                        attach: RunAttachParams = conf.attach
+                    if isinstance(conf, RunAttachConfiguration) and isinstance(
+                        conf.attach, RunAttachParams
+                    ):
+                        attach = conf.attach
                         if attach.proxy.type == "jump":
                             attach_proxy_config = SSHProxyJumpConfig(attach.proxy.proxy_jump)
                         elif attach.proxy.type == "command":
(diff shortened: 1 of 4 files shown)


peterschmidt85 (changes requested):
Thank you, @mrkn very much for the PR.

A few questions and comments:

1. I assume you've tested the PR end-to-end, have you? I mean submitting runs using the `aws` backend?
2. I assume this won't work if `dstack` doesn't create (and reuse) the default EC2 instance profile that is granted with `ssm` permissions.
3. I personally think that `ssm` should be a setting of the `aws` backend rather than a part of the user configuration. This means if `ssm` is enabled for the backend, `dstack` would always require `ssm` when attaching.
4. I agree that it should be possible to configure AWS credentials on the client side. I personally think, as the first initial version, we could simply reply on the default AWS credentials. This gives the user 100% control over that when using the `dstack` CLI.
5. Also, in my view, the attach logic cannot be outside of `run.attach`. The `run.attach` is the high-level API that should do anything what is required internally. Users may use it not via CLI but programmatically via `dstack` Python API (see [examples](https://dstack.ai/docs/reference/api/python/#usage-example)).
6. You also use `ssm` with jump hosts (`{run_name}-jump-host`) which is only required with the `kubernetes` backend (and has nothing to do with SSM).
7. I wonder if the PR also works under Windows (I guess not).

Please let me know if you need more assistance with it! We're actually happy to help as this functionality is definitely important.

peterschmidt85:
Hi @mrkn! Will close the PR for now. Please feel free to re-open if you decide to get back to it!

Issue #3126, opened 2025-09-23
[Meta] Improve `kubernetes` backend


un-def:
**Essential:**
* [x] Request resources according to the `dstack` configuration
* [x] Multi-node support (distributed tasks running on fleets with cluster placement)

**Strategic:**
* [x] AMD GPUs support
* [ ] Allow to configure multiple clusters per backend (e.g. per `region`)
* [ ] Auto-scaling support (ideally, find a way to support it for any clouds)

Improvements:
* [x] Update the jump pod: use a lightweight image, restrict SSH access (see TODOs in `_create_jump_pod_service`)
* [x] Test and update (if required) the gateway functionality on managed/self-hosted Kubernetes other than EKS (see TODO in `KubernetesCompute.create_gateway`)

PR #3141, opened 2025-10-01
Kubernetes: add multi-node support

un-def:
* Discover and set instance's internal_ip (PodIP)
* Fix region mismatch
* Add `privileged: true` support
* [runner] Set RLIMIT_MEMLOCK to unlimited. Fixes issues with InfiniBand/RDMA

Part-of: https://github.com/dstackai/dstack/issues/3126

### DIFF
--- a/src/dstack/_internal/core/backends/base/compute.py
+++ b/src/dstack/_internal/core/backends/base/compute.py
@@ -323,6 +323,7 @@ class ComputeWithCreateInstanceSupport(ABC):
 class ComputeWithPrivilegedSupport:
     """
     Must be subclassed to support runs with `privileged: true`.
+    All VM-based Computes (that is, Computes that use the shim) should subclass this mixin.
     """
 
     pass
--- a/src/dstack/_internal/core/backends/base/compute.py
+++ b/src/dstack/_internal/core/backends/base/compute.py
@@ -320,6 +320,14 @@ class ComputeWithCreateInstanceSupport(ABC):
             ]
 
 
+class ComputeWithPrivilegedSupport:
+    """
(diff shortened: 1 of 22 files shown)


src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -320,6 +320,14 @@ def _restrict_instance_offer_az_to_volumes_az(
             ]
 
 
+class ComputeWithPrivilegedSupport:
+    """
+    Must be subclassed to support runs with `privileged: true`.
+    """
+
+    pass
```
jvstme, line 329: (nit) I can suggest to mention in the docstring and possibly in `contriburing/BACKENDS.md` that this mixin should be subclassed in all VM-based backends.

Backend contributors don't necessarily know what is required for `privileged` to work and usually aim for a minimal implementation, so I'm afraid this mixin can be unintentionally omitted from new backends.

src/dstack/_internal/core/backends/kubernetes/compute.py
```diff
@@ -333,6 +348,30 @@ def run_job(
...
+        project_ssh_public_key: str,
+        project_ssh_private_key: str,
+    ):
+        pod = call_api_method(
+            self.api.read_namespaced_pod,
+            client.V1Pod,
+            name=provisioning_data.instance_id,
+            namespace=self.config.namespace,
+        )
+        pod_id = get_value(pod, ".status.pod_ip", str)
```
jvstme: (nit) `pod_id` -> `pod_ip`?

jvstme (approved):

PR #3154, opened 2025-10-06
[Hot Aisle] Support multi-GPU VMs

### LINKED ISSUE #3146: [Hot Aisle] Support multi-GPU VMs
Currently, we only support 1-GPU VMs as only this type of VMs was available at the time of implementation. Today, Hot Aisle offers also multi-GPU VMs.

peterschmidt85:
Fixes #3146

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -264,6 +264,7 @@ nav:
             - Python API: docs/reference/api/python/index.md
             - REST API: docs/reference/api/rest/index.md
         - Environment variables: docs/reference/environment-variables.md
+        - .dstack/profiles.yml: docs/reference/profiles.yml.md
         - Plugins:
             - Python API: docs/reference/plugins/python/index.md
             - REST API: docs/reference/plugins/rest/index.md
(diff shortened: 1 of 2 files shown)


peterschmidt85:
Perhaps we should also whitelist 8x instances too already - they come public soon.

src/dstack/_internal/core/backends/hotaisle/compute.py
```diff
@@ -66,6 +86,7 @@ def get_all_offers_with_availability(self) -> List[InstanceOfferWithAvailability
             locations=self.config.regions or None,
             catalog=self.catalog,
         )
+        logger.info(f"Offers: {offers}")
```
r4victor: Seems too verbose to log backend offers

mkdocs.yml
```diff
@@ -264,7 +264,6 @@ nav:
             - Python API: docs/reference/api/python/index.md
             - REST API: docs/reference/api/rest/index.md
         - Environment variables: docs/reference/environment-variables.md
-        - .dstack/profiles.yml: docs/reference/profiles.yml.md
```
r4victor, line 267: Why?

r4victor (approved):

PR #3166, opened 2025-10-08
[Docker] Update the CUDA version in the default Docker image to 12.8 (from 12.1)

### LINKED ISSUE #3163: [Docker] Update the CUDA version in the default Docker image to 12.8 (from 12.1)
Following our update of the CUDA drivers in the [OS image](https://github.com/dstackai/dstack/pull/3099), we need to update the Docker images too!

peterschmidt85:
Fixes #3163

- [x] Build 0.11rc3 on staging and do a minimum test
- [x] Build 0.11 on production
- [x] Update the image version in sources to 0.11

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -2,7 +2,7 @@
 ARG UBUNTU_VERSION
 
 # Build stage
-FROM nvidia/cuda:12.1.1-base-ubuntu${UBUNTU_VERSION}.04 AS builder
+FROM nvidia/cuda:12.8.1-base-ubuntu${UBUNTU_VERSION}.04 AS builder
 
 ENV NCCL_HOME=/opt/nccl
 ENV CUDA_HOME=/usr/local/cuda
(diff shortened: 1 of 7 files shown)


r4victor:
@peterschmidt85, IIRC, cuda 12.8 drops support for old architectures which includes V100, so these old GPUs should be handled if dstack still supports them without custom images.

r4victor:
Old architectures are said to only be deprecated but I remember having some issues with 12.8 and V100. Maybe related to pytorch or smth, so worth testing.

> Architecture support for Maxwell, Pascal, and Volta is considered feature-complete and will be frozen in an upcoming release.

https://docs.nvidia.com/cuda/archive/12.8.0/cuda-toolkit-release-notes/index.html#deprecated-or-dropped-features

peterschmidt85:
> @peterschmidt85, IIRC, cuda 12.8 drops support for old architectures which includes V100, so these old GPUs should be handled if dstack still supports them without custom images.

@r4victor Interestingly enough, when we use V100, we would automatically fallback to our previous OS image that comes with CUDA 12.2, and if we run our new Docker image with CUDA 12.8, V100 will work - I've just confirmed that.

I propose to merge this PR as is. cc @jvstme

peterschmidt85:
Building 0.11 nowo via https://github.com/dstackai/dstack/actions/runs/18349808238

peterschmidt85:
0.11 for Azure is finally built via https://github.com/dstackai/dstack/actions/runs/18377527747/job/52356839044

PR #3176, opened 2025-10-09
[Feature]: Store user SSH key on the server 

### LINKED ISSUE #2053: [Feature]: Store user SSH key on the server
### Problem

1. A user starts a run with `dstack apply`.
1. The user switches to another computer.
1. The user tries to attach to the run using `dstack attach`.

The user won't be able to attach and will get the `Can't connect to the remote host` error.

This happens because the SSH key for attaching is [generated in CLI](https://github.com/dstackai/dstack/blob/9e126a44ed859005d1c457f6d01e89019c411959/src/dstack/api/_public/repos.py#L172) and stored on the machine, so different machines will have different keys, even if the user is the same.

### Solution

Generate and store users' SSH keys in dstack-server and download them to the client machine when needed.
This ensures that all of a user's machines have the same key, and the only secret the user needs to access their runs from a new machine is the dstack API token.

### Workaround 1
- Generate a private+public key pair with `ssh-keygen` and deliver them to all machines where access to the run in desired.
- Before running `dstack apply`, init the repo with `dstack init --ssh-identity <path to private key>`
- To attach to the run, use `dstack attach --ssh-identity <path to private key> <run name>`
### Workaround 2
Manually copy `~/.dstack/ssh/id_rsa` and `~/.dstack/ssh/id_rsa.pub` from one machine to the rest.

Yes

peterschmidt85:
Fixes #2053

- This PR introduces two new properties to the user model: `ssh_private_key` and `ssh_public_key`.  These properties are automatically assigned when users are created or when a user submits a run without specifying a public key.
- When a run is submitted without a specified public key, dstack automatically assigns it to the run spec model from the user model.
- The public key is now optional in the run spec. However, it is assigned during server validation if it is empty. This ensures the public key in the run spec in the database is always set.
- The attachment logic now checks the run’s public key and attempts to use the user’s private key if it matches. Otherwise, it falls back to the legacy local SSH key.

> [!WARNING]
> The CLI still assigns the public key from the legacy local SSH key. However, if a run was submitted without an SSH key (for example, via the API and automatically assigned the user’s public key), the new CLI will be able to attach to it using the user’s key.

Backward compatibility:
- The new CLI (>=0.19.33) works with old servers (<0.19.33).
- The old CLI (<0.19.33) works with new servers (>=0.19.33).
- The new CLI (>=0.19.33) can attach to runs submitted by old servers (<0.19.33).

> [!WARNING]
> Since 0.19.40, the client will stop using the legacy local SSH key and will require that all users have updated their servers to at least 0.19.33.


src/dstack/_internal/core/models/users.py
```diff
@@ -38,3 +38,5 @@ class UserTokenCreds(CoreModel):
 
 class UserWithCreds(User):
     creds: UserTokenCreds
+    ssh_public_key: Optional[str] = None
```
jvstme: (nit) `ssh_public_key` could be part of the `User` model, since it is not secret

src/dstack/_internal/server/services/users.py
```diff
@@ -107,6 +112,7 @@ async def update_user(
     email: Optional[str] = None,
     active: bool = True,
 ) -> UserModel:
+    # TODO: Allow to update ssh_private_key and ssh_public_key
```
jvstme: (nit) Doesn't seem to be necessary.
```suggestion
```

src/dstack/_internal/cli/commands/offer.py
```diff
@@ -104,7 +104,6 @@ def _command(self, args: argparse.Namespace):
 
         run_spec = RunSpec(
             configuration=conf,
-            ssh_key_pub="(dummy)",
```
jvstme, line 107: (major) This breaks compatibility with older servers.

```shell
> dstack offer --project my-sky-project
Server validation error: 
{'detail': [{'loc': ['body', 'run_spec', 'ssh_key_pub'],
             'msg': 'none is not an allowed value',
             'type': 'type_error.none.not_allowed'}]}
```

src/dstack/api/_public/__init__.py
```diff
@@ -29,30 +28,23 @@ def __init__(
         self,
         api_client: APIClient,
         project_name: str,
-        ssh_identity_file: Optional[PathLike] = None,
```
jvstme, line 32: (major) This is a breaking change to the public Python API interface.

Consider keeping the `ssh_identity_file` parameter and the now-unused `self.ssh_identity_file` attribute, but raise a deprecation warning if it's not None.

src/dstack/api/_public/runs.py
```diff
@@ -72,13 +77,11 @@ def __init__(
         self,
         api_client: APIClient,
         project: str,
-        ssh_identity_file: Optional[PathLike],
```
jvstme, line 75: Although it's unlikely that Python API users construct `Run` objects manually, this is technically a breaking change too. I'd keep the parameter and raise a deprecation warning if it is ever not None

src/dstack/api/_public/runs.py
```diff
@@ -270,9 +273,33 @@ def attach(
...
+                with open(ssh_identity_file, "wb", opener=key_opener) as f:
+                    assert user.ssh_private_key
+                    f.write(user.ssh_private_key.encode())
+            else:
+                if config_manager.dstack_key_path.exists():
+                    # TODO: Remove since 0.19.35
+                    warn(
+                        f"Using legacy [code]{config_manager.dstack_key_path}[/code]."
+                        " Future versions will use the user SSH key from the server.",
+                    )
```
jvstme, line 303: Is the user supposed to do anything when they see this warning? If not, the warning is unnecessary and will probably only confuse the user, since they typically don't think about what keys `dstack` uses internally

src/dstack/api/_public/runs.py
```diff
@@ -465,6 +493,19 @@ def get_run_plan(
...
+        if ssh_identity_file:
+            ssh_key_pub = Path(ssh_identity_file).with_suffix(".pub").read_text()
+        else:
+            config_manager = ConfigManager()
+            if not config_manager.dstack_key_path.exists():
+                generate_rsa_key_pair(private_key_path=config_manager.dstack_key_path)
+            warn(
+                f"Using legacy [code]{config_manager.dstack_key_path.with_suffix('.pub')}[/code]."
+                " Future versions will use the user SSH key from the server.",
+            )
```
jvstme, line 511: Same as my previous comment

src/dstack/api/_public/runs.py
```diff
@@ -718,7 +761,6 @@ def get_plan(
             configuration_path=configuration_path,
             configuration=configuration,
             profile=profile,
-            ssh_key_pub=Path(self._client.ssh_identity_file + ".pub").read_text().strip(),
```
jvstme, line 721: This will break Python API users that use `get_plan` and a pre-0.19.33 server.

`get_plan` has been deprecated for 7 months, so this is hopefully a low-impact issue, but I'd still prefer not to introduce it.

src/dstack/api/_public/runs.py
```diff
@@ -465,6 +493,19 @@ def get_run_plan(
...
+        else:
+            config_manager = ConfigManager()
+            if not config_manager.dstack_key_path.exists():
+                generate_rsa_key_pair(private_key_path=config_manager.dstack_key_path)
+            warn(
+                f"Using legacy [code]{config_manager.dstack_key_path.with_suffix('.pub')}[/code]."
+                " Future versions will use the user SSH key from the server.",
+            )
+            ssh_key_pub = config_manager.dstack_key_path.with_suffix(".pub").read_text()
+            # TODO: Uncomment after 0.19.35
```
jvstme: (nit) 0.19.35 will be out in two weeks, and we are definitely not breaking backward compatibility that soon.

I think we've discussed that the best practice for describing backward compatibility logic is to specify for which past releases it is relevant, not in which future release it should be dropped.

There are more comments that reference 0.19.35 in the PR.

jvstme (approved):

src/dstack/api/_public/runs.py
peterschmidt85: Decided to keep the warning for now, but we can remove it in a separate PR once we’ve dogfooded it more.

jvstme: I think seeing it on every apply and attach and not being able to do anything about it will be a bit annoying. Anyways, it's up to you

PR #3177, opened 2025-10-10
Support specifying kubeconfig directly as yaml

### LINKED ISSUE #3167: [Feature]: server/config.yml load kubeconfig via data key
### Problem

Currently when using the kubernetes backend the configuration block expects a reference to a `kubeconfig` filename.
```yaml
projects:
- name: main
    backends:
    - type: kubernetes
      kubeconfig:
        filename: ~/.kube/config
```

However, that reference is not something that can be entered via the `dstack server` backend interface running in a docker container. 

### Solution

Noted in the code https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/core/backends/kubernetes/models.py#L39-L50 there is a reference to be able to specify a `data` key instead of `filename`. It is unclear to me if this already supported, but if it is then an example on that would be great. I am assuming it would work like this, but not clear if that is right.

```yaml
projects:
- name: main
    backends:
    - type: kubernetes
      kubeconfig:
        data:
         apiVersion: v1
         clusters:
           - cluster:
               certificate-authority-data: <redacted>
               server: https://cluster.k8s.io
             name: kubernetes
         contexts:
          - context:
               cluster: kubernetes
               namespace: development
               user: dstack-server
            name: dstack-server-context
         current-context: dstack-server-context
         kind: Config
         users:
         - name: dstack-server
           user:
            token: <redacted>
      proxy_jump:
        hostname: 204.12.171.137
        port: 32000
```
It looks like this is somewhat described in https://dstack.ai/docs/reference/server/config.yml/#kubernetes-kubeconfig although that assumes you have compact raw JSON blob of the config, while generally you deal with yaml files.

```bash
cat my-service-account-file.json | jq -c | jq -R
```

### Workaround

don't use the dstack server UI or if the `data` block is supported, use something similar to the following to convert the yaml.

```bash
kubectl config --kubeconfig ~/.kube/config  view --raw=true -o json |  jq -c | jq -R
```

Yes

r4victor:
Closes #3167 

It's now possible to specify kubeconfig for kubernetes backend directly as yaml via the `data` field:

```yaml
type: kubernetes
kubeconfig:
  data:
    apiVersion: v1
    clusters:
    - cluster:
        # ...
    contexts:
    - context:
        # ...
    # ...
```

Previously, `kubeconfig` contents have to be converted to a string first.

### DIFF
--- a/docs/docs/reference/server/config.yml.md
+++ b/docs/docs/reference/server/config.yml.md
@@ -287,10 +287,26 @@ to configure [backends](../../concepts/backends.md) and other [server-level sett
         show_root_heading: false
 
 ??? info "Specifying `data`"
-    To specify kubeconfig contents directly via `data`, convert it to a string:
+    To specify kubeconfig contents directly via `data`, you can convert it to a string:
 
     ```shell
     yq -o=json ~/.kube/config | jq -c | jq -R
+    ```
+
+    or copy kubeconfig contents under `data` as-is:
+
+    ```yaml
+    type: kubernetes
+    kubeconfig:
+      data:
+        apiVersion: v1
+        clusters:
+        - cluster:
+            # ...
+        contexts:
+        - context:
+            # ...
+        # ...
     ```
 
 ###### `projects[n].backends[type=kubernetes].proxy_jump` { #kubernetes-proxy_jump data-toc-label="proxy_jump" }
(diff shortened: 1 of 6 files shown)


r4victor:
Closing as you can already specify kubeconfig as yaml multiline string, e.g. 

```
type: kubernetes
kubeconfig:
  data: |
    apiVersion: v1
    clusters:
    - cluster:
        # ...
    contexts:
    - context:
        # ...
    # ...
```

PR #3178, opened 2025-10-10
Kubernetes: add AMD GPU support

un-def:
In addition, Pod.spec.affinity has been fixed (the inner structure was incorrect)

Part-of: https://github.com/dstackai/dstack/issues/3126


peterschmidt85 (approved):
I tested it on AMD Developer Cloud together with Dmitry on Friday.

@r4victor please have a look and merge!

PR #3187, opened 2025-10-13
Replace `Instance.termination_reason` values with codes 

### LINKED ISSUE #3182: [Internal]: Replace Instance.termination_reason values with codes
Currently, Instance.termination_reason uses human-readable messages rather then codes. 
In some cases, we need to use codes instead. That's why it's better to use current values in termination_reason_message and for termination_reason, use codes – ideally enums.

peterschmidt85:
Fixes #3182


src/dstack/_internal/server/background/tasks/process_running_jobs.py
```diff
@@ -383,7 +385,14 @@ async def _process_running_job(session: AsyncSession, job_model: JobModel):
...
                     # when CLI <= 0.19.8 is no longer supported
-                    job_model.termination_reason = JobTerminationReason.INTERRUPTED_BY_NO_CAPACITY
+                    if (
+                        job_model.instance is not None
+                        and job_model.instance.termination_reason
+                        == InstanceTerminationReason.NO_BALANCE.value
+                    ):
+                        job_model.termination_reason = JobTerminationReason.NO_BALANCE
+                    else:
+                        job_model.termination_reason = JobTerminationReason.INSTANCE_UNREACHABLE
```
r4victor: If you just replace INTERRUPTED_BY_NO_CAPACITY with INSTANCE_UNREACHABLE for spot instances, they will never be terminated due to no capacity. No capacity retries won't work. We can replace INTERRUPTED_BY_NO_CAPACITY with INSTANCE_UNREACHABLE only for on-demand.

src/dstack/_internal/server/models.py
```diff
@@ -615,6 +615,7 @@ class InstanceModel(BaseModel):
     # instance termination handling
     termination_deadline: Mapped[Optional[datetime]] = mapped_column(NaiveDateTime)
     termination_reason: Mapped[Optional[str]] = mapped_column(String(4000))
+    termination_reason_message: Mapped[Optional[str]] = mapped_column(String(4000))
```
r4victor, line 673: Not obvious at all what's in termination_reason vs termination_reason_message. If you look at other models, their termination_reason is EnumAsString, so you know the models have certain enum values. As of now, InstanceModel.termination_reason can be any string.

peterschmidt85: I thought of replacing termination_reason with EnumAsString but this won't be backward compatible and would require database migration. I suggest that for now we use String(4000) but much later replace with EnumAsString. I can leave a TODO.

src/dstack/_internal/server/background/tasks/process_running_jobs.py
peterschmidt85: I updated it. Please checlk!

src/dstack/_internal/server/models.py
peterschmidt85: Added TODO

src/dstack/_internal/server/models.py
```diff
@@ -614,7 +614,9 @@ class InstanceModel(BaseModel):
 
     # instance termination handling
     termination_deadline: Mapped[Optional[datetime]] = mapped_column(NaiveDateTime)
+    # TODO: Migrate to EnumAsString(InstanceTerminationReason, 100) after enough releases to ensure backward compatibility
     termination_reason: Mapped[Optional[str]] = mapped_column(String(4000))
```
r4victor: What happens after enough releases? Old instances continue to have non-enum values forever.

jvstme: Addressed by using `InstanceTerminationReason` straight away and converting old non-enum values to `InstanceTerminationReason` on reads

jvstme:
Updated the PR, addressed the pending review comment, and added small changes to termination reason naming. Also slightly reduced the scope of the PR. See commit messages for details and motivation

r4victor (approved):
👍

PR #3193, opened 2025-10-15
fix(VastAICompute): filter region before offer

DragonStuff:
Adds a filter for VastAI offers by config.regions, which uses `jp-japan` for example. At the moment, most provider computes pass locations=self.config.regions to pre-filter (AWS, Azure, GCP, OCI, Vultr, Runpod, Lambda, DigitalOcean, Cudo, etc.). Because the server later filters regions by exact equality to offer.region, region constraints still worked if you used the provider’s slug (e.g., jp-japan).

Essentially with this addition, the backend pre-filters by exact location slug early; which means fewer items processed.

PS: I'm thinking that there could be some kind of abstraction in the future that groups offers automatically. For us, picking inference within a certain region really matters for latency. We want to have our application's compute as close as possible to our GPU workloads to reduce latency.

Testing internally at the moment, marked as draft.

### DIFF
--- a/src/dstack/_internal/core/backends/vastai/compute.py
+++ b/src/dstack/_internal/core/backends/vastai/compute.py
@@ -56,46 +56,12 @@ class VastAICompute(
     def get_offers_by_requirements(
         self, requirements: Requirements
     ) -> List[InstanceOfferWithAvailability]:
-        def spot_filter(offer):
-            return not offer.instance.resources.spot
-
-        extra_filter = spot_filter
-
-        configured_regions = self.config.regions or []
-        if configured_regions:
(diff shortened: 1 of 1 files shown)


r4victor:
@DragonStuff, thanks for the PR. Regarding the grouping, we were thinking about introducing generic dstack locations so that users can refer to them instead of provider-specific locations but this a) involves too much work to add and maintain a mapping for each provider and b) it's hard to come up with good location list suitable for every provider.

Maybe we could allow regex-like patterns in region names for filtering to support filtering by `us-*` for vastai if this suits you. Feel free to share your ideas.

PR #3216, opened 2025-10-22
Use server-managed user SSH keys for new runs

### LINKED ISSUE #3199: [Feature]: Use the user's SSH key in `dstack` CLI
### Problem

#2053 added support for user SSH keys managed by the `dstack` server. However, the CLI still uses `~/.dstack/ssh/id_rsa` when applying runs, so the initial problem remains the same — users cannot attach to runs that they started from another machine.

### Solution

Use server-managed SSH keys when starting new runs from the CLI.

> [!NOTE]
> `~/.dstack/ssh/id_rsa` should still remain supported for compatibility with old runs and servers.

### Workaround

Use `dstack apply --ssh-identity ...` to enable access to the run from different machines.

Yes

jvstme:
This commit updates the CLI and the server to use
server-managed user SSH keys when starting new
runs. This allows users to attach to the run from
different machines, since the SSH key is
automatically replicated to all clients.

Implementation details:
- Server:
    - If the user key is missing, generate it when the user first calls `/get_my_user`.
- Client:
    - Before applying or getting a run plan, call `/get_my_user` to check if the user key is available. If it is, use it.
    - Cache the downloaded keys in `~/.dstack/ssh` to avoid repeated `/get_my_user` calls.
    - Switch from `warn` to logger messages, since this code is part of the Python API, so its output should be configurable.

Closes #3199

### DIFF
--- a/src/dstack/_internal/core/services/ssh/key_manager.py
+++ b/src/dstack/_internal/core/services/ssh/key_manager.py
@@ -1,0 +1,56 @@
+import os
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional
+
+from dstack._internal.core.models.users import UserWithCreds
+
+if TYPE_CHECKING:
+    from dstack.api.server import APIClient
+
+KEY_REFRESH_RATE = timedelta(minutes=10)  # redownload the key periodically in case it was rotated
+
+
+@dataclass
+class UserSSHKey:
+    public_key: str
+    private_key_path: Path
+
+
+class UserSSHKeyManager:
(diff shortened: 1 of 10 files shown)


src/dstack/_internal/core/services/ssh/key_manager.py
```diff
@@ -0,0 +1,56 @@
+import os
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional
+
+from dstack._internal.core.models.users import UserWithCreds
+
+if TYPE_CHECKING:
+    from dstack.api.server import APIClient
```
r4victor, line 10: Why put UserSSHKeyManager in core instead of api?

PR #3222, opened 2025-10-24
Switch to nebius sdk 0.3

### LINKED ISSUE #3171: [Bug]: nebius backend does not work with nebius sdk 0.2.73
### Steps to reproduce

1. Install latest nebius sdk: pip install nebius==0.2.73
2. See that nebius backend can be configured:
```
[15:03:45] WARNING  dstack._internal.server.services.config:151 Backend nebius  
                    not available and won't be configured. Check that backend   
                    dependencies are installed.  
```

### Actual behaviour

Apparently, [nebius sdk 0.2.73 release removed ](https://github.com/nebius/pysdk/commit/f2dbd25200ea749b9a7ea00451d888e548d0d299)`from nebius.aio.authorization.options import options_to_metadata`, which broke dstack nebius integration (https://github.com/dstackai/dstack/blob/688e9c8f10a967104da0a1df513fa9360fde608f/src/dstack/_internal/core/backends/nebius/resources.py#L11).

Ideally, nebius should fix backward compatibility so that older dstack versions continue to work.

### Expected behaviour

_No response_

### dstack version

master

### Server logs

```shell

```

### Additional information

_No response_

r4victor:
Closes #3171

The PR updates nebius backend to use the new API from nebius sdk 0.3.


src/dstack/_internal/core/backends/nebius/compute.py
```diff
@@ -379,9 +379,7 @@ def _wait_for_instance(sdk: SDK, op: SDKOperation[Operation]) -> None:
             op.status(),
         )
         time.sleep(WAIT_FOR_INSTANCE_UPDATE_INTERVAL)
-        resources.LOOP.await_(
-            op.update(per_retry_timeout=resources.REQUEST_TIMEOUT, metadata=resources.REQUEST_MD)
-        )
+        resources.LOOP.await_(op.update(per_retry_timeout=resources.REQUEST_TIMEOUT))
```
r4victor: @jvstme, op.update does not accept auth_options so not sure if it needs any. Do you know of any issues it may cause?

jvstme: > not sure if it needs any

I think it does and it's worth reporting to the maintainer that the parameter is missing.

`Operation.update` [calls](https://github.com/nebius/pysdk/blob/ad716cc89ea0715667e6c66f81d3017a8366af9f/src/nebius/aio/operation.py#L204) `OperationServiceClient.get`, which [does](https://github.com/nebius/pysdk/blob/ad716cc89ea0715667e6c66f81d3017a8366af9f/src/nebius/api/nebius/common/v1/__init__.py#L2094) accept `auth_options`.

> Do you know of any issues it may cause?

I'd expect the method call to [block for 15 minutes](https://github.com/nebius/pysdk?tab=readme-ov-file#test-token-renewal) if the credentials are invalid and `auth_options` is missing. For example, this can happen if the service account key expires right before we attempt the call

jvstme (approved):
It may be worth waiting until the issue from [this comment](https://github.com/dstackai/dstack/pull/3222#discussion_r2461120503) is resolved in the SDK. They usually resolve issues very quickly

src/dstack/_internal/core/backends/nebius/compute.py
r4victor: Thanks for confirmation. Filed a nebius issue.

PR #3223, opened 2025-10-24
Fix context usage in executor

matiasinsaurralde:
Sharing this small improvement - As we're already using `context` in `setupRepo` and `execJob` it makes sense to properly pass the context to the other functions.


r4victor (commented):
Overall passing context seems like an improvement but I have concerns about this particular case with `moveDir`. Suppose context is cancelled while restoreRepoDir – then cp could restore files only partially. This is problematic if repo is stored on a volume.

I think restoreRepoDir shouldn't be cancelled with the run context.

PR #3224, opened 2025-10-24
Fix context usage in shim package

matiasinsaurralde:
Same as #3223 but for the shim package.

Had to update the `Backend` interface to take the context 👍

### DIFF
--- a/runner/internal/shim/docker.go
+++ b/runner/internal/shim/docker.go
@@ -540,7 +540,7 @@ func unmountVolumes(ctx context.Context, taskConfig TaskConfig) error {
 			log.Info(ctx, "skipping", "mountpoint", mountPoint, "output", output)
 			continue
 		}
-		cmd = exec.CommandContext(ctx, "umount", "-qf", mountPoint)
+		cmd = exec.CommandContext(context.TODO(), "umount", "-qf", mountPoint)
 		if output, err := cmd.CombinedOutput(); err != nil {
 			log.Error(ctx, "failed to unmount", "mountpoint", mountPoint, "output", output)
 			failed = append(failed, mountPoint)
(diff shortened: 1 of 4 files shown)


r4victor (commented):
Cancelling some operations like "umount" can be problematic. Not sure it's an improvement.

matiasinsaurralde:
> Cancelling some operations like "umount" can be problematic. Not sure it's an improvement.

Makes sense - I've refactored the PR a bit, still passing context to `unmountVolumes` for logging, etc. But added `context.TODO` in the specific call tu `umount`: 

```go
exec.CommandContext(context.TODO(), "umount", "-qf", mountPoint)
```

I've inspected the other operations (there should be no impact if cancelled, they return errors and/or can be retried):
- `mountpoint` - just checks if the path is mounted.
- `lsblk` - query device info
- `mkfs.ext4` - formatting can be retried
- `mount` - can be retried

runner/internal/shim/docker.go
```diff
@@ -535,12 +535,12 @@ func unmountVolumes(ctx context.Context, taskConfig TaskConfig) error {
...
 	for _, volume := range taskConfig.Volumes {
 		mountPoint := getVolumeMountPoint(volume.Name)
-		cmd := exec.Command("mountpoint", mountPoint)
+		cmd := exec.CommandContext(ctx, "mountpoint", mountPoint)
 		if output, err := cmd.CombinedOutput(); err != nil {
 			log.Info(ctx, "skipping", "mountpoint", mountPoint, "output", output)
 			continue
 		}
-		cmd = exec.Command("umount", "-qf", mountPoint)
+		cmd = exec.CommandContext(context.TODO(), "umount", "-qf", mountPoint)
```
r4victor, line 543: So what's the benefit of CommandContext(context.TODO()?

PR #3226, opened 2025-10-25
Fix context usage in internal metrics package

matiasinsaurralde:
Same as #3223 and #3224 - In this case I've also updated the runner API so that the request context is passed over.


r4victor (approved):

r4victor:
@matiasinsaurralde, thx!

PR #3239, opened 2025-10-29
Drop hardcoded Hot Aisle VM specs

jvstme:
Use the spec object from gpuhunt offers instead.
This allows newly added instance types with
different CPU, RAM, disk, and GPU count
configurations to automatically become available
in dstack. However, limit the supported GPUs to
MI300X, since other GPUs and CPU-only VMs might
need to be tested by the dstack team before they
become available to users.

See also https://github.com/dstackai/gpuhunt/pull/194

**Before merging**:
- [x] Release and pin new gpuhunt version


peterschmidt85:
@Bihan pls review once you have a chance!

Bihan (approved):

peterschmidt85:
@jvstme feel free to merge - its reviewed!

PR #3267, opened 2025-11-06
Add SGLang Router Support

Bihan:
**Intro**
We want to make it possible to create a gateway which extends the gateway functionality with additional features (all sgl-router features such as cache aware routing, etc) while keeping all the standard gateway features (such as authentication, rate limits).

For the user, using such gateway should be very simple, e.g. setting router to sglang. - for the gateway configurations. The rest for the user should look the same - the same service endpoint, authentication and rate limits working, etc.
While the first change should only bring minimum features - allow to route replicas traffic through the router (dstack’s gateway/ngnix -> sglang-router -> replica workers), in the future this may be extended with router-specific scaling metrics, such as ttft, itl, dissagregated PD. See [TTFT/ITL Based Auto Scaling Issue](https://github.com/dstackai/dstack/issues/3293)).  

As the first experimental version, the most critical is to come up with the minimum changes that are tested thoroughly that would allow embedding the router without breaking any existing functionality.

**Key Changes**
1. Add `src/dstack/_internal/core/models/routers.py`
Define router types and configuration models. The RouterType enum identifies available routers. Each router has its own config model (SGLangRouterConfig, VLLMRouterConfig) with router-specific options. AnyRouterConfig allows to select the correct config class based on the type field.

2. Add `router: AnyRouterConfig`  in `GatewayConfiguration` and  in `GatewayComputeConfiguration`  
Ensure router config flows from user input → server → backend compute layer.

3. Update `gateway/pyproject.toml` to include router packages as optional dependencies

4. Update `get_dstack_gateway_commands()` in `src/dstack/_internal/core/backends/base/compute.py` to accept router config

5. Update `_update_gateway()` in `src/dstack/_internal/server/services/gateways/__init__.py` to extract router_config

6. Add abstract Router base class in `src/dstack/_internal/proxy/gateway/model_routers/base.py`
Handles lifecycle methods of router. 

7. Extend abstract Router base class and implement SGLangRouter in `src/dstack/_internal/proxy/gateway/model_routers/sglang.py`

8. Add router register `src/dstack/_internal/proxy/gateway/model_routers/__init__.py`
Implement the registry pattern (similar to dstack's backend configurators) for auto-discovery and lookup of available routers.

9. Update src/dstack/_internal/proxy/gateway/services/nginx.py

10. Update upstream block of src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2  to  forward request when router is defined.

11. Add new nginx config to src/dstack/_internal/proxy/gateway/resources/nginx/router_workers.jinja2 to make service replicas's available to TCP port.  Later we could avoid this extra proxying layer by switching from Unix sockets to TCP ports when opening SSH tunnels on the gateway.

**How Router Upgrade Works**

Steps
1. Let’s say Gateway currently has dstack-gateway 0.19.34 and sglang-router 0.2.1 and we are releasing new dstack-gateway 0.19.35 with new SGLang feature in latest SGLang version (sglang-router 0.2.2)
2. We bump up sglang-router version to 0.2.2 in gateway/pyproject.toml
3. dstack server restart -> init_gateways() called -> _update_gateway() -> update.sh executes
4. Flips to inactive venv, let’s say now version = “green”
5. Installs new version in green venv

**Gateway Service Restart and Gateway Instance Reboot**
Router has been tested to successfully reconnect to replicas after both a gateway service restart and a full gateway instance reboot.

**How to test**

Step 1
Apply Below Gateway Config

```
#gateway.dstack.yml
type: gateway
name: bihan-gateway


# Gateways are bound to a specific backend and region
backend: aws
region: eu-west-1

# This domain will be used to access the endpoint
domain: example.com
router:
  type: sglang
  policy: cache_aware
```

Step 2
Update DNS

Step 3
We want to test with multiple services therefore, apply below service configs. 

**Config1**

```
#sglang-service1.yml
type: service
name: sglang-service1

python: 3.12
nvcc: true

env:
  - HF_TOKEN
  - HF_HUB_DISABLE_XET=1
  - HF_HUB_ENABLE_HF_TRANSFER=0

commands:
  - pip install --upgrade pip
  - pip install uv
  - uv pip install sglang --prerelease=allow
  - python -m sglang.launch_server --model-path $MODEL_ID --host 0.0.0.0 --port 8000 --enable-metrics

port: 8000
model: sglang-service1.meta-llama/Llama-3.2-3B-Instruct

resources:
  gpu: 24GB

replicas: 0..2
scaling:
   metric: rps
   target: 1
```

**Config2**
```
#sglang-service2.yml
type: service
name: sglang-service2

python: 3.12
nvcc: true

env:
  - HF_TOKEN
  - HF_HUB_DISABLE_XET=1
  - HF_HUB_ENABLE_HF_TRANSFER=0

commands:
  - pip install --upgrade pip
  - pip install uv
  - uv pip install sglang --prerelease=allow
  - python -m sglang.launch_server --model-path $MODEL_ID --host 0.0.0.0 --port 8000 --enable-metrics

port: 8000
model: sglang-service2.meta-llama/Llama-3.2-3B-Instruct

resources:
  gpu: 24GB

replicas: 0..2
scaling:
   metric: rps
   target: 1
```

Step 3
To automate request and test autoscaling, you can use below script: `autoscale_test_sglang.py`

```
import asyncio
import aiohttp
import time
import json

# ==== Configuration ====
URL = "https://sglang-service1.example.com/v1/chat/completions" # <-- replace with your endpoint
TOKEN = "esdfds3263-c36d-41db-ba9b-0d31df4efb15e"   # <-- replace with your token
RPS = 2            # requests per second
DURATION = 1800        # duration in seconds
# =======================

HEADERS = {
    "Content-Type": "application/json",
    "Authorization": f"Bearer {TOKEN}"
}

PAYLOAD = {
    "model": "sglang-service1.meta-llama/Llama-3.2-3B-Instruct", #<--replace with your model
    "messages": [
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "What is Deep Learning?"}
    ]
}


async def send_request(session, idx):
    """Send a single request and print full response"""
    try:
        async with session.post(URL, headers=HEADERS, json=PAYLOAD) as resp:
            text = await resp.text()
            print(f"\n[{idx}] Status: {resp.status}")
            print(f"Response:\n{text}\n")
    except Exception as e:
        print(f"[{idx}] Error: {e}")


async def run_load_test():
    total_requests = RPS * DURATION
    interval = 1.0 / RPS

    async with aiohttp.ClientSession() as session:
        start_time = time.perf_counter()
        tasks = []

        for i in range(total_requests):
            tasks.append(asyncio.create_task(send_request(session, i + 1)))
            await asyncio.sleep(interval)

        await asyncio.gather(*tasks)

        elapsed = time.perf_counter() - start_time
        print(f"\n✅ Sent {total_requests} requests in {elapsed:.2f}s "
              f"(~{total_requests/elapsed:.2f} RPS)")


if __name__ == "__main__":
    asyncio.run(run_load_test())
```

Step 6
After updating token and service endpoint, run above script `python autoscale_test_sglang.py ` from your local machine.

Once the automated requests start hitting the service endpoint; dstack submits the job. When the service get's deployed and /health check from sglang-router responds with 200 as shown below, you will start to see response from the model.

As the automated requests continue, first dstack scales up to 2 jobs. If we stop the requests, dstack scales down to 0 jobs.

**Note:** 
1. This PR uses  "https://bihan-test-bucket.s3.eu-west-1.amazonaws.com/dstack_gateway-0.0.1-py3-none-any.whl"  for testing. Later once the PR is ready for merge, I will update it in `src/dstack/_internal/core/backends/base/compute.py`

2. For testing `gateway/pyproject.toml` has my fork as dependency. I will update it once the PR is ready for merge.


src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -1021,16 +1022,24 @@ def get_dstack_gateway_wheel(build: str) -> str:
...
+        gateway_package = f"dstack-gateway[{router.type}]"
+    else:
+        gateway_package = "dstack-gateway"
     return [
         "mkdir -p /home/ubuntu/dstack",
         "python3 -m venv /home/ubuntu/dstack/blue",
         "python3 -m venv /home/ubuntu/dstack/green",
-        f"/home/ubuntu/dstack/blue/bin/pip install {get_dstack_gateway_wheel(build)}",
+        f"/home/ubuntu/dstack/blue/bin/pip install {wheel}",
+        f"/home/ubuntu/dstack/blue/bin/pip install --upgrade '{gateway_package}'",
```
jvstme: (nit) It should be possible to install the correct extra with one pip command:

```shell
$ pip install "dstack-gateway[sglang] @ https://bihan-test-bucket.s3.eu-west-1.amazonaws.com/dstack_gateway-0.0.1-py3-none-any.whl"
```

This should be slightly faster than calling pip twice.

src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -449,11 +450,24 @@ async def _update_gateway(gateway_compute_model: GatewayComputeModel, build: str
...
+        gateway_package = "dstack-gateway"
+
     commands = [
         # prevent update.sh from overwriting itself during execution
         "cp dstack/update.sh dstack/_update.sh",
         f"sh dstack/_update.sh {get_dstack_gateway_wheel(build)} {build}",
         "rm dstack/_update.sh",
+        # Install gateway package with router extras to the active venv (blue or green)
+        # update.sh writes the active version to dstack/version
+        f"version=$(cat /home/ubuntu/dstack/version) && /home/ubuntu/dstack/$version/bin/pip install --upgrade '{gateway_package}'",
```
jvstme: (nit) It should be possible to install the correct extra by passing it to `update.sh`.

```shell
$ sh dstack/_update.sh "dstack-gateway[sglang] @ https://bihan-test-bucket.s3.eu-west-1.amazonaws.com/dstack_gateway-0.0.1-py3-none-any.whl"
```

This should be slightly faster than calling pip twice (here and inside `update.sh`). This will also allow the gateway app to import router-specific packages if it has to. Currently, it can't import them, because it is started by `update.sh` *before* the router extra is installed.

src/dstack/_internal/core/models/routers.py
```diff
@@ -0,0 +1,27 @@
...
+
+from pydantic import Field
+from typing_extensions import Annotated, Literal
+
+from dstack._internal.core.models.common import CoreModel
+
+
+class RouterType(str, Enum):
+    SGLANG = "sglang"
+    VLLM = "vllm"
```
jvstme: vLLM routers are not yet supported, yet vLLM settings will be visible to users in things like the automatically generated reference and IDE hints.

I'd suggest to remove anything vLLM-related to avoid confusing users.

src/dstack/_internal/core/models/routers.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class SGLangRouterConfig(CoreModel):
+    type: Literal["sglang"] = "sglang"
+    policy: str = "cache_aware"
```
jvstme: Currently, setting `policy` to invalid values, such as `policy: invalid`, is ignored. The gateway is then provisioned successfully, but starting a service fails with a hard to debug error.

```
Unexpected error: status code 500 when requesting http://localhost:3000/api/project/ilya/runs/apply. Check the server logs for backend issues, and the CLI
logs at (~/.dstack/logs/cli/latest.log) local CLI output
```

I can suggest to make this field a `typing.Literal` that only allows valid values.

```
  --policy {random,round_robin,cache_aware,power_of_two}
                        Load balancing policy to use. In PD mode, this is used for both prefill and decode unless overridden (default: cache_aware)
```

Then the user will see a detailed configuration validation error when trying to create the gateway.

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
```diff
@@ -0,0 +1,62 @@
+from typing import Dict, List, Optional, Type
+
+from dstack._internal.core.models.routers import AnyRouterConfig, RouterType
+from dstack._internal.utils.logging import get_logger
+
+from .base import Replica, Router, RouterContext
+
+logger = get_logger(__name__)
+
+"""This provides a registry of available router implementations."""
```
jvstme: (nit) Should be at the top of the module, otherwise it is not a docstring.

```python
>>> dstack._internal.proxy.gateway.model_routers.__doc__ is None
True
```

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+except ImportError as e:
+    logger.warning("SGLang router not available: %s", e)
+
+_ROUTER_TYPE_TO_CLASS_MAP: Dict[RouterType, Type[Router]] = {}
+
+for router_class in _ROUTER_CLASSES:
+    router_type_str = getattr(router_class, "TYPE", None)
+    if router_type_str is None:
+        logger.warning(f"Router class {router_class.__name__} missing TYPE attribute, skipping")
+        continue
```
jvstme: (nit) Missing `TYPE` variable indicates a bug, so it should raise an exception and, in this case, cause the application to crash. So I can suggest to:
- Access `router_class.TYPE` directly, without `getattr`.
- Define `Router.TYPE` using `typing.ClassVar`, so that type checkers (and devs) know that it should be set in subclasses.

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+    router_type = RouterType(router_type_str)
```
jvstme: (nit) You can use `RouterType` as the `Router.TYPE` type annotation, so that this conversion is not necessary.

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,366 @@
...
+
+            # Use Python from the active venv
+            venv_python = DSTACK_DIR_ON_GATEWAY / version / "bin" / "python3"
+
+            cmd = [
+                str(venv_python),
+                "-m",
+                "sglang_router.launch_router",
+                "--host",
+                "0.0.0.0",  # Bind to all interfaces (nginx connects via 127.0.0.1)
```
jvstme: **(major)**

If we only need `127.0.0.1`, bind only to `127.0.0.1` as an extra security measure to avoid exposing the SGLang API publicly.

We might also need to pass `--prometheus-host 127.0.0.1`, because currently sglang binds to `0.0.0.0:29000` and serves prometheus metrics there.

```shell
$ netstat -tlpn
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 127.0.0.1:10001         0.0.0.0:*               LISTEN      -                   
tcp        0      0 0.0.0.0:29000           0.0.0.0:*               LISTEN      2961/sglang::router 
tcp        0      0 127.0.0.1:8000          0.0.0.0:*               LISTEN      2923/python3        
tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN      -                   
tcp        0      0 0.0.0.0:3000            0.0.0.0:*               LISTEN      2961/sglang::router 
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      -                   
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      -                   
tcp6       0      0 :::80                   :::*                    LISTEN      -                   
tcp6       0      0 :::22                   :::*                    LISTEN      -
```

src/dstack/_internal/proxy/gateway/model_routers/base.py
```diff
@@ -0,0 +1,147 @@
...
+    def remove_replicas(self, replicas: List[Replica]) -> None:
+        """Unregister replicas from the router (actual API calls to remove workers).
+
+        Args:
+            replicas: The list of replicas to remove from router.
+
+        Raises:
+            Exception: If removing replicas fails.
+        """
+        ...
```
jvstme: (nit) Unused, can delete here and in `SglangRouter`

src/dstack/_internal/proxy/gateway/model_routers/base.py
```diff
@@ -0,0 +1,147 @@
...
+    def register_replicas(
+        self, domain: str, num_replicas: int, model_id: Optional[str] = None
+    ) -> List[Replica]:
+        """Register replicas to a domain (allocate ports/URLs for workers).
+
+        Args:
+            domain: The domain name for this service.
+            num_replicas: The number of replicas to allocate for this domain.
+            model_id: Optional model identifier (e.g., "meta-llama/Meta-Llama-3.1-8B-Instruct").
+                Required only for routers that support IGW (Inference Gateway) mode for multi-model serving.
```
jvstme: (nit) This method is supposed to return `list[Replica]`, and `Replica.model` is required. So I can suggest to make the `model_id` parameter required too. We can make it optional later if we will actually need it for some other router type

src/dstack/_internal/proxy/gateway/model_routers/base.py
```diff
@@ -0,0 +1,147 @@
...
+    """Abstract base class for router implementations (e.g., SGLang, vLLM).
+
+    A router manages the lifecycle of worker replicas and handles request routing.
+    Different router implementations may have different mechanisms for managing
+    replicas.
+    """
+
+    def __init__(
+        self,
+        router: Optional[AnyRouterConfig] = None,
```
jvstme: (nit) It seems a bit confusing to me that to construct a `Router` you need a `router`.

Maybe `config` is a better name for this parameter

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,366 @@
...
+
+            # Determine active venv (blue or green)
+            version_file = DSTACK_DIR_ON_GATEWAY / "version"
+            if version_file.exists():
+                version = version_file.read_text().strip()
+            else:
+                version = "blue"
+
+            # Use Python from the active venv
+            venv_python = DSTACK_DIR_ON_GATEWAY / version / "bin" / "python3"
```
jvstme: The path to the current interpreter should be available in `sys.executable`.

Reconstructing it from `~/dstack/version` is more error-prone, because we may change how the gateway glue-green deployment works (or switch to a completely different deployment model) and forget to update it here

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,366 @@
...
+                "--log-level",
+                self.context.log_level,
+                "--log-dir",
+                str(self.context.log_dir),
+            ]
+
+            if hasattr(self.config, "policy") and self.config.policy:
+                cmd.extend(["--policy", self.config.policy])
+
+            # Add additional required configs here
```
jvstme: (nit) What additional configs?

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,366 @@
...
+                "--port",
+                str(self.context.port),
+                "--enable-igw",
+                "--log-level",
+                self.context.log_level,
+                "--log-dir",
+                str(self.context.log_dir),
+            ]
+
+            if hasattr(self.config, "policy") and self.config.policy:
```
jvstme: (nit)

> `hasattr(self.config, "policy")`

`self.config` is of type `SGLangRouterConfig`, which means it always has an attribute called `policy`, so this check is redundant.

> `and self.config.policy`

This check can only fail if `self.config.policy` is `""`. I can't think of a reason why anyone would want to set `policy: ""` in YAML. So I'd suggest to forbid `policy: ""` on the model level (e.g., make `policy` a `Literal`) and drop this check.

src/dstack/_internal/proxy/gateway/services/model_routers/base.py
```diff
@@ -0,0 +1,147 @@
...
+        ...
+
+    @abstractmethod
+    def stop(self) -> None:
+        """Stop the router process.
+
+        Raises:
+            Exception: If the router fails to stop.
+        """
+        ...
```
jvstme, line 58: (nit) Unused, can delete here and in `SglangRouter`

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def is_running(self) -> bool:
+        """Check if the SGLang router is running and responding to HTTP requests."""
+        try:
+            result = subprocess.run(
+                ["curl", "-s", f"http://{self.context.host}:{self.context.port}/workers"],
+                capture_output=True,
+                timeout=5,
+            )
+            return result.returncode == 0
```
jvstme: Spawning a new OS process just to make an HTTP request looks inefficient. We also can't assume that `curl` is installed on the gateway, since we don't install it explicitly.

You can use `httpx` instead.

Same in other `SglangRouter` methods.

src/dstack/_internal/proxy/gateway/model_routers/base.py
```diff
@@ -0,0 +1,147 @@
...
+        """Initialize router with context.
+
+        Args:
+            router: Optional router configuration (implementation-specific)
+            context: Runtime context for the router (host, port, logging, etc.)
+        """
+        self.context = context or RouterContext()
+
+    @abstractmethod
+    def start(self) -> None:
```
jvstme: `Router.start` and most other `Router` method invocations can take up to several seconds, during which they will block the event loop and prevent the gateway from handling any requests.

To avoid that, you can:
- Make all `Router` methods `async`.
- Revise the `SglangRouter` implementation and replace all blocking calls (`time.sleep`, `subprocess.run`, etc) with non-blocking alternatives (`asyncio.sleep`, `httpx.AsyncClient`, etc).

src/dstack/_internal/proxy/lib/models.py
```diff
@@ -57,6 +58,8 @@ class Service(ImmutableModel):
     client_max_body_size: int  # only enforced on gateways
     strip_prefix: bool = True  # only used in-server
     replicas: tuple[Replica, ...]
+    router: Optional[AnyRouterConfig] = None
```
jvstme, line 61: If the router is configured globally for the gateway, I'm not sure it's a good choice to store a copy of the router config with each service.

For example, what if a newly registered service has a config that is different from existing services (e.g., because a new dstack release adds a new config property, or changes the default for a property, or the user updates the gateway configuration in-place)? Will we update the router configs of all previously registered services to match the new config? Or will different services have different router configs? If so, what are the semantics of that?

Considering the router config is supposed to be global for the gateway, a better place to store it is `gateway.models.GlobalProxyConfig`. And a better way to pass it to the gateway is `/api/config` rather than `/api/registry/services/register`.

(unless we want a separate router for each service)

src/dstack/_internal/proxy/lib/models.py
```diff
@@ -57,6 +58,8 @@ class Service(ImmutableModel):
...
+    model_id: Optional[str] = None
```
jvstme: The model to service relationship is already established via `ChatModel.run_name`, so adding `Service.model_id` to represent the exact same relationship adds redundancy to the stored data, which comes with potential data inconsistency issues.

Instead, I'd suggest to add a method like `GatewayProxyRepo.get_model_by_run` (similar to the existing `delete_models_by_run`) and use it whenever we need to retrieve the model ID for a service.

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -168,6 +217,26 @@ def write_global_conf(self) -> None:
...
         self.write_conf(conf, "00-log-format.conf")
 
+    def write_router_workers_conf(self, conf: ServiceConfig, allocated_ports: list[int]) -> None:
+        """Write router workers configuration file (generic)."""
+        # Pass ports to template
+        workers_config = generate_router_workers_config(conf, allocated_ports)
+        workers_conf_name = f"router-workers.{conf.domain}.conf"
+        workers_conf_path = self._conf_dir / workers_conf_name
+        sudo_write(workers_conf_path, workers_config)
+        self.reload()
```
jvstme: **(major)**

There is a method called `write_conf` that is worth using here. That method rolls back the config if Nginx fails to reload. Without it, you risk planting an invalid config, which will make all future Nginx reload attempts fail.

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -77,16 +85,47 @@ class Nginx:
...
+                try:
+                    await run_async(self.write_router_workers_conf, conf, allocated_ports)
+                except Exception as e:
+                    logger.exception(
+                        "write_router_workers_conf failed for domain=%s: %s", conf.domain, e
+                    )
+                    raise
+                finally:
+                    # Always update router state, regardless of nginx reload status
+                    await run_async(self._router.update_replicas, replicas)
```
jvstme: Why? If Nginx did not reload, it means that replicas are not available on TCP ports, which means there is no point in pointing SGLang to those ports

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -77,16 +85,47 @@ class Nginx:
...
     async def register(self, conf: SiteConfig, acme: ACMESettings) -> None:
         logger.debug("Registering %s domain %s", conf.type, conf.domain)
         conf_name = self.get_config_name(conf.domain)
-
         async with self._lock:
             if conf.https:
                 await run_async(self.run_certbot, conf.domain, acme)
             await run_async(self.write_conf, conf.render(), conf_name)
 
+            if isinstance(conf, ServiceConfig) and conf.router and conf.model_id:
```
jvstme: **(major)**

> `conf.model_id`

Currently, if you apply a service without `model`, it is provisioned successfully, but doesn't work — the service URL just leads to the router, where this service is not actually registered, so the router always returns 404.

There doesn't seem to be a way for the user to find out that setting `model` is required.

I can suggest to add validation for this case on the server (in [`_register_service_in_gateway`](https://github.com/dstackai/dstack/blob/05847cd86e7b66fee2e57a0ec84764abaf2ccf38/src/dstack/_internal/server/services/services/__init__.py#L72)) and reject registering services without `model` on gateways with `router`.

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,366 @@
...
+
+        if is_new_model:
+            # Store domain -> model_id mapping
+            self._domain_to_model_id[domain] = model_id
+
+            # Allocate ports for replicas
+            allocated_ports = []
+            for _ in range(num_replicas):
+                allocated_ports.append(self._next_worker_port)
+                self._next_worker_port += 1
```
jvstme: Sooner or later, `_next_worker_port` will reach `65536`, which will cause any updates to services and replicas to fail.

Even before that, it is possible that one of the allocated ports will already be taken by some other program on the host, e.g., some client bound to an ephemeral port. My testing shows that in that case Nginx will fail to bind to the port, and SGLang-router will be sending traffic to some random program listening on that port instead of Nginx.

I can suggest to at least allocate ports from a fixed range, such as `10001`-`11999`, so that we never reach ephemeral ports nor get out of valid port bounds.

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
```diff
@@ -4,9 +4,13 @@ limit_req_zone {{ zone.key }} zone={{ zone.name }}:10m rate={{ zone.rpm }}r/m;
 
 {% if replicas %}
 upstream {{ domain }}.upstream {
+    {% if router is not none %}
+    server 127.0.0.1:3000;  # SGLang router on the gateway
```
jvstme: **(major)**

This exposes all SGLang APIs, including `/workers`, to anyone who can request the service.

```shell
> curl https://sglang-service1.my-gateway.example/workers
{"workers":[{"id":"http://127.0.0.1:10006","url":"http://127.0.0.1:10006","model_id":"sglang-service1.meta-llama/Llama-3.2-3B-Instruct","priority":50,"cost":1.0,"worker_type":"regular","is_healthy":true,"load":0,"connection_mode":"Http","metadata":{"model_id":"sglang-service1.meta-llama/Llama-3.2-3B-Instruct"}},{"id":"http://127.0.0.1:10007","url":"http://127.0.0.1:10007","model_id":"sglang-service2.meta-llama/Llama-3.2-3B-Instruct","priority":50,"cost":1.0,"worker_type":"regular","is_healthy":true,"load":0,"connection_mode":"Http","metadata":{"model_id":"sglang-service2.meta-llama/Llama-3.2-3B-Instruct"}}],"total":2,"stats":{"prefill_count":0,"decode_count":0,"regular_count":2}}
```

This might be a security problem for `dstack` users, especially considering this allows to change workers and not just view them. And I also don't think we want to allow any users to interfere with the worker settings, because they are supposed to be managed by dstack.

I can suggest to forbid requesting `/workers` and other administrative SGLang APIs, if any. Or even introduce a whitelist of allowed routes, which is more secure, but may require more maintenance.

jvstme: **(major)**

Since all service domains just lead to the same router, you can request any service using any domain. For example, in the case from the PR description, you can access the `sglang-service1.meta-llama/Llama-3.2-3B-Instruct` model by requesting `sglang-service2.example.com` (instead of `sglang-service1.example.com`).

This causes various issues:
- It allows unauthorized access to services (if at least one service is `auth: false`, or if we support reusing gateways between projects). 
- It allows to bypass service rate limits.
- RPS autoscaling will be triggered for the service that owns the domain rather than the service that actually handles the load.
- It is confusing.

Not sure how we can address that. If SGLang provides alternatives to routing by model ID (e.g., routing by path or hostname), then we could use them. If not, maybe we could run a separate SGLang router for each service. What do you think?

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+
+_ROUTER_CLASSES: List[Type[Router]] = []
+
+try:
+    from dstack._internal.proxy.gateway.model_routers.sglang import SglangRouter
+
+    _ROUTER_CLASSES.append(SglangRouter)
+    logger.debug("Registered SglangRouter")
+except ImportError as e:
+    logger.warning("SGLang router not available: %s", e)
```
jvstme: (nit) We can import `SglangRouter` regardless of whether `sglang-router` is installed, so this `except` will never be triggered.

So it seems that this router registration mechanism is redundant (at least in the current version) and can be replaced by a bare `SglangRouter` import.

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+    """Factory function to create a router instance from router configuration."""
+    router_type = RouterType(router.type)
+    router_class = get_router_class(router_type)
+
+    if router_class is None:
+        available_types = [rt.value for rt in _AVAILABLE_ROUTER_TYPES]
+        raise ValueError(
+            f"Router type '{router_type.value}' is not available. "
+            f"Available types: {available_types}"
+        )
```
jvstme: (nit) For consistency with the rest of the `dstack-proxy` codebase, consider using the exceptions from [`proxy.lib.errors`](https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/proxy/lib/errors.py).

Same in a few other `raise` statements in this PR

jvstme (commented):
@Bihan, thanks for the PR, I tested it and the main use cases seem to work as expected. I’ve added some comments for things that may need attention and tagged them to make priorities easier to navigate

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
Bihan: @jvstme Thanks for pointing our this valid concern. Currently, SGLang only provides routing by model ID. I think we can go with separate SGLang router for each service.

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
Bihan: Resolved

Bihan: Resolved

src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -448,10 +450,22 @@ async def _update_gateway(gateway_compute_model: GatewayComputeModel, build: str
...
+    compute_config = GatewayComputeConfiguration.__response__.parse_raw(
+        gateway_compute_model.configuration
+    )
+
+    # Build package spec with extras and wheel URL (similar to compute.py)
+    wheel = get_dstack_gateway_wheel(build)
+    if compute_config.router:
+        gateway_package = f"dstack-gateway[{compute_config.router.type}] @ {wheel}"
+    else:
+        gateway_package = f"dstack-gateway @ {wheel}"
```
jvstme: (nit) This logic is duplicated here and in `backends/base/compute.py`. Consider moving it to `get_dstack_gateway_wheel` to avoid duplication and future errors

src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -1021,16 +1022,23 @@ def get_dstack_gateway_wheel(build: str) -> str:
         r.raise_for_status()
         build = r.text.strip()
         logger.debug("Found the latest gateway build: %s", build)
-    return f"{base_url}/dstack_gateway-{build}-py3-none-any.whl"
+    # return f"{base_url}/dstack_gateway-{build}-py3-none-any.whl"
+    return "https://bihan-test-bucket.s3.eu-west-1.amazonaws.com/dstack_gateway-0.0.1-py3-none-any.whl"
```
nikita-toffee-ai: I'm certain it's not supposed to be hard-coded, we need to get the dynamic URL back.

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,269 @@
...
+
+from .base import Router, RouterContext
+
+logger = get_logger(__name__)
+
+
+class SglangRouter(Router):
+    """SGLang router implementation with 1:1 service-to-router."""
+
+    TYPE = "sglang"
```
nikita-toffee-ai: Wouldn't it be better if we took the `sglang` literal value from the `RouterType` enum? That would be DRY.

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -0,0 +1,269 @@
...
+            # Determine active venv (blue or green)
+            version_file = DSTACK_DIR_ON_GATEWAY / "version"
+            if version_file.exists():
+                version = version_file.read_text().strip()
+            else:
+                version = "blue"
+
+            venv_python = DSTACK_DIR_ON_GATEWAY / version / "bin" / "python3"
+
+            prometheus_port = self.context.port + 10000
```
nikita-toffee-ai, line 50: Where is the `10000` magic number coming from?

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -0,0 +1,269 @@
...
+                "--log-dir",
+                str(self.context.log_dir),
+            ]
+
+            if hasattr(self.config, "policy") and self.config.policy:
+                cmd.extend(["--policy", self.config.policy])
+
+            subprocess.Popen(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+
+            time.sleep(2)
```
nikita-toffee-ai, line 74: Do you think clients would benefit from being able to customize this constant e.g. via an env?

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,269 @@
...
+            logger.error(f"Failed to start sglang-router-new: {e}")
+            raise
+
+    def stop(self) -> None:
+        try:
+            result = subprocess.run(
+                ["lsof", "-ti", f":{self.context.port}"], capture_output=True, timeout=5
+            )
+            if result.returncode == 0:
+                pids = result.stdout.decode().strip().split("\n")
```
nikita-toffee-ai: Wouldn't a platform-independent `os.linesep` be a better choice here?

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,269 @@
...
+    def update_replicas(self, replica_urls: List[str]) -> None:
+        """Update replicas for service, replacing the current set."""
+        # Query router to get current worker URLs
+        current_workers = self._get_router_workers()
+        current_worker_urls: set[str] = set()
+        for worker in current_workers:
+            url = worker.get("url")
+            if url and isinstance(url, str):
+                current_worker_urls.add(url)
+        target_worker_urls = set(replica_urls)
```
nikita-toffee-ai: We need to make sure there will be no path artifacts in set elements, like trailing slashes, otherwise, the set operations below could yield unexpected results.

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -0,0 +1,269 @@
...
+        # Remove workers
+        for worker_url in sorted(workers_to_remove):
+            success = self._remove_worker_from_router(worker_url)
+            if not success:
+                logger.warning("Failed to remove worker %s, continuing with others", worker_url)
+
+    def _get_router_workers(self) -> List[dict]:
+        try:
+            result = subprocess.run(
+                ["curl", "-s", f"http://{self.context.host}:{self.context.port}/workers"],
```
nikita-toffee-ai: Same as before, we should use `httpx` for HTTP queries.

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -77,14 +86,103 @@ class Nginx:
...
+                        # Allocate router port for new router
+                        router_port = self._allocate_router_port()
+                        conf.router_port = router_port
+
+                        # Create per-service log directory
+                        log_dir = Path(f"./router_logs/{conf.domain}")
+
+                        # Create router context with allocated port
+                        ctx = RouterContext(
+                            host="127.0.0.1",
```
nikita-toffee-ai: We've hard-coded `127.0.0.1` in quite a few places already, I think we should take it from the source model where it was once defined instead.

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -164,10 +293,149 @@ def certificate_exists(domain: str) -> bool:
...
+                    return True
+                except OSError:
+                    # If bind fails (e.g., Address already in use), port is not available
+                    return False
+        except Exception:
+            logger.debug("Error checking port %s availability", port)
+            return False
+
+    def _allocate_router_port(self) -> int:
+        """Allocate next available router port in fixed range (20000-24999).
```
nikita-toffee-ai: I would remove the exact range from the comment, because range ends are vars by nature as we see right down below.

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -164,10 +293,149 @@ def certificate_exists(domain: str) -> bool:
...
+                port = self._ROUTER_PORT_MIN  # Wrap around
+            attempts += 1
+
+        raise UnexpectedProxyError(
+            f"Router port range exhausted ({self._ROUTER_PORT_MIN}-{self._ROUTER_PORT_MAX}). "
+            "All ports in range appear to be in use."
+        )
+
+    def _allocate_worker_ports(self, num_ports: int) -> list[int]:
+        """Allocate worker ports globally in fixed range (10001-11999).
```
nikita-toffee-ai: I would remove the exact range from the comment, because range ends are vars by nature as we see right down below.

src/dstack/_internal/core/backends/base/compute.py
Bihan: Yes you are right. In production, it will not be hardcoded and the hardcoded URL will be replaced by `return f"{base_url}/dstack_gateway-{build}-py3-none-any.whl`

src/dstack/_internal/core/backends/base/compute.py
Bihan: Resolved.

src/dstack/_internal/server/services/gateways/__init__.py
Bihan: Resolved

src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -454,20 +454,19 @@ async def _update_gateway(gateway_compute_model: GatewayComputeModel, build: str
...
         "cp dstack/update.sh dstack/_update.sh",
-        f"sh dstack/_update.sh {get_dstack_gateway_wheel(build)} {build}",
+        # Pass the full package spec (with extras) to update.sh instead of just the wheel
+        f"sh dstack/_update.sh '{gateway_package}' {build}",
         "rm dstack/_update.sh",
-        # Install gateway package with router extras to the active venv (blue or green)
-        # update.sh writes the active version to dstack/version
-        f"version=$(cat /home/ubuntu/dstack/version) && /home/ubuntu/dstack/$version/bin/pip install --upgrade '{gateway_package}'",
     ]
     stdout = await connection.tunnel.aexec("/bin/sh -c '" + " && ".join(commands) + "'")
```
jvstme: **(major)**

Gateway update doesn't work now:

```
[19:44:03] WARNING  dstack._internal.server.services.gateways:423 Failed to update gateway 34.244.47.57:
```

This error message is not informative, but a quick debug shows that the problem is that we pass a malformed command to `tunnel.aexec`, where single quotes added on line 468 unintentionally close single quotes from line 471.

```
/bin/sh -c 'cp dstack/update.sh dstack/_update.sh && sh dstack/_update.sh 'dstack-gateway[sglang] @ https://bihan-test-bucket.s3.eu-west-1.amazonaws.com/dstack_gateway-0.0.1-py3-none-any.whl' latest && rm dstack/_update.sh'
```

And I assume the same issue is present in `core/backends/compute.py`, where single quotes from [`get_dstack_gateway_commands`](https://github.com/Bihan/dstack/blob/b8794f1d385a2a7e10504ee46452aa60804dc065/src/dstack/_internal/core/backends/base/compute.py#L1041) can interfere with single quotes from [`_get_gateway_commands`](https://github.com/Bihan/dstack/blob/b8794f1d385a2a7e10504ee46452aa60804dc065/src/dstack/_internal/core/backends/kubernetes/compute.py#L974) of the `kubernetes` backend, which most likely breaks k8s gateway provisioning.

src/dstack/_internal/core/models/routers.py
Bihan: Resolved

src/dstack/_internal/core/models/routers.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Here `lsof` always emits "\n" line endings, so splitting on "\n" matches the actual command output.

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
```diff
@@ -35,6 +59,7 @@ server {
         {% endif %}
     }
     {% endfor %}
+    {% endif %}
```
jvstme: **(major)**

This seems to break [rate limits](https://dstack.ai/docs/concepts/services/#rate-limits) — since rate limit locations are now only rendered in this `{% else %}` block, they won't be rendered on SGLang gateways.

To make rate limits work, yet forbid requests to anything but `/generate` and `/v1`, I can suggest to remove the `/generate` and `/v1` locations above, and build them dynamically along with rate limit locations in [`get_nginx_service_config`](https://github.com/dstackai/dstack/blob/2c97b3a8d7028a19840a60ada656b645b3b18e70/src/dstack/_internal/proxy/gateway/services/registry.py#L304)

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved. Current Implementation does not use service registration based on model_id.

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
Bihan: This is how I am implementing.
In `service.jinja2`
```
...
{% for location in locations %}
    location {{ location.prefix }} {
        {% if auth %}
        auth_request /_dstack_auth;
        {% endif %}

        try_files /nonexistent @$http_upgrade;

        {% if location.limit_req %}
        limit_req zone={{ location.limit_req.zone }}{% if location.limit_req.burst %} burst={{ location.limit_req.burst }} nodelay{% endif %};
        {% endif %}
    }
    {% endfor %}

    {# For SGLang router: block all requests except whitelisted locations added dynamically above #}
    {% if router is not none and router.type == "sglang" %}
    location ~ . {
        return 403;
    }
    {% endif %}
...
```

In [get_nginx_service_config](https://github.com/dstackai/dstack/blob/2c97b3a8d7028a19840a60ada656b645b3b18e70/src/dstack/_internal/proxy/gateway/services/registry.py#L304)

```
# For SGLang routers, add whitelisted endpoints as locations
    if service.router and service.router.type == "sglang":
        locations.append(LocationConfig(prefix="= /generate", limit_req=None))
        locations.append(LocationConfig(prefix="~ ^/v1/", limit_req=None))
```

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
jvstme: There are also some corner cases to consider in `get_nginx_service_config` when SGLang is used:
- If the user configures rate limits for locations outside of `= /generate` and `/v1/`, they should **not** be rendered, so that users cannot use them to expose sensitive URLs, such as `/workers`.
- If the user configures rate limits for something that is a prefix of `/generate` or `/v1/`  (e.g., `/`, `/gen`, `/v`, `/generate`, `/v1/`, etc), these limits should probably be included in the automatically added `= /generate` and `/v1/` locations.
- The `/` location should not be [automatically added](https://github.com/dstackai/dstack/blob/2c97b3a8d7028a19840a60ada656b645b3b18e70/src/dstack/_internal/proxy/gateway/services/registry.py#L327).

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/__init__.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/base.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/base.py
Bihan: Resolved. Not required now

src/dstack/_internal/proxy/gateway/model_routers/base.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: @jvstme There are other configs listed [here](https://docs.sglang.ai/advanced_features/router.html#configuration-reference). Additional options may be added in future iterations based on testing and user needs.

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/model_routers/base.py
Bihan: Now it is used.

src/dstack/_internal/server/services/gateways/__init__.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/lib/models.py
Bihan: Resolved. With current implementation model_id is removed from Service

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
Bihan: Resolved

src/dstack/_internal/server/services/gateways/__init__.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
Bihan: Prometheus port is offset by 10000 from router port to keep it in a separate range. Now included in comments

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/model_routers/base.py
```diff
@@ -79,67 +67,23 @@ def is_running(self) -> bool:
...
-            replicas: The list of replicas to add to router.
-
-        Raises:
-            Exception: If adding replicas fails.
-        """
-        ...
-
-    @abstractmethod
-    def remove_replicas(self, replicas: List[Replica]) -> None:
+    def remove_replicas(self, replica_urls: List[str]) -> None:
```
jvstme, line 70: (nit) This method looks redundant, because it is possible to remove replicas using `update_replicas`.

For example, [this](https://github.com/Bihan/dstack/blob/71cacf2a53d6a818a902dc2e3a1bfbf5f3b284da/src/dstack/_internal/proxy/gateway/services/nginx.py#L203) can be replaced with `run_async(router.update_replicas, [])`

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -1,366 +1,255 @@
...
 
 logger = get_logger(__name__)
 
 
 class SglangRouter(Router):
-    """SGLang router implementation using IGW (Inference Gateway) mode for multi-model serving."""
+    """SGLang router implementation with 1:1 service-to-router."""
 
-    TYPE = "sglang"
+    TYPE = RouterType.SGLANG
```
jvstme, line 23: (nit) Unused, can remove

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -1,366 +1,255 @@
...
-                subprocess.run(["pkill", "-f", "sglang::router"], timeout=5)
+                pids = result.stdout.decode().strip().split("\n")
+                for pid in pids:
+                    if pid:
+                        logger.info(
+                            "Stopping sglang-router-new process (PID: %s) on port %s",
+                            pid,
+                            self.context.port,
+                        )
+                        subprocess.run(["kill", pid], timeout=5)
```
jvstme: (nit) Use `os.kill`

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -1,366 +1,255 @@
...
-            if result.returncode == 0:
-                response = json.loads(result.stdout.decode())
-                workers = response.get("workers", [])
-                # Filter by model_id
-                workers = [w for w in workers if w.get("model_id") == model_id]
-                return workers
-            return []
+            with httpx.Client(timeout=5.0) as client:
+                response = client.get(f"http://{self.context.host}:{self.context.port}/workers")
+                if response.status_code == 200:
```
jvstme, line 191: (nit) Log an error if status is not 200

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -34,10 +35,9 @@ class SiteConfig(BaseModel):
 
     def render(self) -> str:
         template = read_package_resource(f"{self.type}.jinja2")
-        return jinja2.Template(template).render(
-            **self.dict(),
-            proxy_port=PROXY_PORT_ON_GATEWAY,
-        )
```
jvstme, line 34: (nit) Does this change do anything?

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -85,46 +85,102 @@ class Nginx:
...
+                        # Create new router instance for this service
+                        router = get_router(conf.router, context=ctx)
+
+                        # Store mappings
+                        self._router_port_to_domain[router_port] = conf.domain
+                        self._domain_to_router[conf.domain] = router
+
+                        # Start router if not running
+                        if not await run_async(router.is_running):
+                            await run_async(router.start)
```
jvstme: (nit) This check seems to be redundant — we've just created the router, so we know it's not running yet. It always prints an unnecessary error:

```
dstack._internal.proxy.gateway.model_routers.sglang - ERROR - Error checking sglang router status on port 20003: [Errno 111] Connection refused
```

And this is the only place where the `Router.is_running` method is used, which means the method itself is also redundant and can be removed from the `Router` base class (but it can remain a private `SglangRouter` method, considering it is called from `SglangRouter.start`)

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -85,46 +85,104 @@ class Nginx:
...
+        self._domain_to_router: Dict[str, Router] = {}
+        self._ROUTER_PORT_MIN: int = 20000
+        self._ROUTER_PORT_MAX: int = 24999
+        self._WORKER_PORT_MIN: int = 10001
+        self._WORKER_PORT_MAX: int = 11999
+        self._next_router_port: int = self._ROUTER_PORT_MIN
+        # Tracking of worker ports to avoid conflicts across router instances
+        self._allocated_worker_ports: set[int] = set()
+        self._domain_to_worker_ports: Dict[str, list[int]] = {}
+        self._next_worker_port: int = self._WORKER_PORT_MIN
```
jvstme, line 99: (nit) There's a lot of code duplication for allocating router and worker ports, especially in the `_allocate_router_port` and `_allocate_worker_ports` below. This makes the implementation error-prone and difficult to maintain.

I can suggest to move port allocation logic to a separate class. You can then have two instances of that class — one for routers and one for workers.

```python
self._worker_port_allocator = PortAllocator(MIN_WORKER_PORT, MAX_WORKER_PORT)
self._router_port_allocator = PortAllocator(MIN_ROUTER_PORT, MAX_ROUTER_PORT)
...
worker_ports = self._worker_port_allocator.allocate(count=worker_count)
router_port = self._router_port_allocator.allocate(count=1)[0]
...
self._worker_port_allocator.deallocate(worker_ports)
self._router_port_allocator.deallocate([router_port])
```

src/dstack/_internal/proxy/gateway/services/nginx.py
jvstme: (nit) Clean up resources in case of exceptions — deallocate the port, remove the router from the mapping

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -85,46 +85,102 @@ class Nginx:
...
+
+                        # Start router if not running
+                        if not await run_async(router.is_running):
+                            await run_async(router.start)
+
+                    # Discard old worker ports if domain already has allocated ports (required for scaling case)
+                    if conf.domain in self._domain_to_worker_ports:
+                        old_worker_ports = self._domain_to_worker_ports[conf.domain]
+                        for port in old_worker_ports:
+                            self._allocated_worker_ports.discard(port)
```
jvstme: (nit) This deallocates ports while they are still in use, which can lead to an inconsistent state if there is an error writing the Nginx config with new ports.

Only deallocate old ports after writing the Nginx config.

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -85,46 +85,102 @@ class Nginx:
...
+
+                    allocated_ports = self._allocate_worker_ports(len(conf.replicas))
+                    self._domain_to_worker_ports[conf.domain] = allocated_ports
```
jvstme: (nit) This overwrites the old ports in `_domain_to_worker_ports` while they are still in use, which can lead to an inconsistent state if there is an error writing the Nginx config with new ports.

Only update `_domain_to_worker_ports` after writing the Nginx config.

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -85,46 +85,102 @@ class Nginx:
...
+                    # Write router workers config
+                    try:
+                        if conf.replicas:
+                            await run_async(self.write_router_workers_conf, conf, allocated_ports)
+                    except Exception as e:
+                        # Discard allocated worker ports on error
+                        for port in allocated_ports:
+                            self._allocated_worker_ports.discard(port)
+                        if conf.domain in self._domain_to_worker_ports:
+                            del self._domain_to_worker_ports[conf.domain]
```
jvstme: (nit) This deallocates all ports, even though the replicas are still using the old ports, which leads to an inconsistent state.

Don't deallocate, just keep the old ports as they were before we attempted an update

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -85,46 +85,102 @@ class Nginx:
...
+
+                    # Update replicas to router (actual HTTP API calls to add workers)
+                    try:
+                        await run_async(router.update_replicas, replica_urls)
+                    except Exception as e:
+                        # Free allocated worker ports on error
+                        for port in allocated_ports:
+                            self._allocated_worker_ports.discard(port)
+                        if conf.domain in self._domain_to_worker_ports:
+                            del self._domain_to_worker_ports[conf.domain]
```
jvstme: (nit) This deallocates all ports, even though the replicas are still using them, which leads to an inconsistent state.

Don't deallocate, just log an error

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -213,6 +290,125 @@ def certificate_exists(domain: str) -> bool:
...
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+                try:
+                    sock.bind(("127.0.0.1", port))
+                    # If bind succeeds, port is available
+                    return True
+                except OSError:
+                    # If bind fails (e.g., Address already in use), port is not available
+                    return False
+        except Exception:
+            logger.debug("Error checking port %s availability", port)
```
jvstme: (nit) Doesn't look expected, so should be `warning`, not `debug`

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -213,6 +290,125 @@ def certificate_exists(domain: str) -> bool:
...
+
+        Args:
+            num_ports: Number of worker ports to allocate
+
+        Returns:
+            List of allocated worker port numbers
+        """
+        allocated = []
+        port = self._next_worker_port
+        max_attempts = (self._WORKER_PORT_MAX - self._WORKER_PORT_MIN + 1) * 2  # Allow wrap-around
```
jvstme, line 361: (nit) Why `* 2`? Wrapping around doesn't mean going through the range twice

src/dstack/_internal/proxy/gateway/services/registry.py
```diff
@@ -321,13 +329,39 @@ async def get_nginx_service_config(
         limit_req_zones.append(
             LimitReqZoneConfig(name=zone_name, key=key, rpm=round(rate_limit.rps * 60))
         )
-        locations.append(
-            LocationConfig(
-                prefix=rate_limit.prefix,
-                limit_req=LimitReqConfig(zone=zone_name, burst=rate_limit.burst),
+        if is_sglang:
+            for path in sglang_whitelisted_paths:
+                if rate_limit.prefix == path or path.startswith(rate_limit.prefix):
```
jvstme: (nit) `str_a == str_b` is a special case of `str_b.startswith(str_a)`.


```suggestion
                if path.startswith(rate_limit.prefix):
```

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
```diff
@@ -1,366 +1,255 @@
...
             else:
-                logger.debug("No sglang-router process found to stop")
+                result = subprocess.run(
+                    ["pgrep", "-f", f"sglang.*--port.*{self.context.port}"],
```
jvstme: (nit) Neither `lsof`, nor `pgrep` is guaranteed to be installed. Use `psutil` instead

src/dstack/_internal/proxy/gateway/services/model_routers/__init__.py
jvstme, line 1: (nit) I'd suggest to move the `model_routers` directory to `proxy/gateway/services/model_routers`.

The subdirectories in `proxy/gateway` are supposed to represent architectural tiers:

- `proxy/gateway/repo` - the data tier
- `proxy/gateway/services` - the logic tier
- `proxy/gateway/routers` and `proxy/gateway/schemas` - the presentation tier

The model routers implementation is part of the logic tier.

src/dstack/_internal/proxy/gateway/model_routers/base.py
```diff
@@ -0,0 +1,91 @@
...
+
+class RouterContext(BaseModel):
+    """Context for router initialization and configuration."""
+
+    class Config:
+        frozen = True
+
+    host: str = "127.0.0.1"
+    port: int = 3000
+    log_dir: Path = Path("./router_logs")
```
jvstme: (nit) These defaults here are no longer relevant after switching to per-service routers and dynamic port allocation.

I can suggest to remove the defaults to avoid misleading the devs. And then explicit `RouterContext` will also become required in [`Router.__init__`](https://github.com/dstackai/dstack/blob/71cacf2a53d6a818a902dc2e3a1bfbf5f3b284da/src/dstack/_internal/proxy/gateway/model_routers/base.py#L32)

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
```diff
@@ -4,9 +4,15 @@ limit_req_zone {{ zone.key }} zone={{ zone.name }}:10m rate={{ zone.rpm }}r/m;
 
 {% if replicas %}
 upstream {{ domain }}.upstream {
+    {% if router is not none %}
+    {% if router.type == "sglang" and router_port is not none %}
```
jvstme: (nit) Can replace these 3 conditions with 1 — `router_port is not none`?

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -0,0 +1,255 @@
...
+                    if response_data.get("status") == "accepted":
+                        logger.info(
+                            "Added worker %s to sglang router on port %s",
+                            worker_url,
+                            self.context.port,
+                        )
+                        return True
+                    else:
+                        logger.error("Failed to add worker %s: %s", worker_url, response_data)
+                        return False
```
jvstme, line 223: For some reason I'm always getting this error now:

```
dstack._internal.proxy.gateway.model_routers.sglang - ERROR - Failed to add worker http://127.0.0.1:10001: status 202, {"status":"accepted","worker_id":"http://127.0.0.1:10001","message":"Worker addition queued for background processing"}
```

Could it be because the gateway wheel from S3 depends on sglang-router 0.2.3?

```shell
$ dstack/blue/bin/pip freeze | grep sglang
dstack @ https://github.com/Bihan/dstack/archive/refs/heads/add_sglang_router_support.tar.gz
sglang-router==0.2.3
```

**UPD**: I tried downgrading to 0.2.2 and it didn't help, so maybe this is unrelated

src/dstack/_internal/server/services/gateways/__init__.py
jvstme: The k8s part doesn't seem to be resolved (or pushed) yet

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
Bihan: I've identified an issue affecting SGLang Router versions 0.2.2 and 0.2.3. The worker registration retry mechanism stops at 10 attempts regardless of the --worker-startup-timeout-secs setting. Version 0.2.2 was working until recently. I'm investigating the root cause and will provide an update once I have more details.

src/dstack/_internal/proxy/gateway/model_routers/base.py
Bihan: Yes router_port is only set when a router exists and only sglang router exists currently.

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
Bihan: Yes router_port is only set when a router exists and only sglang router exists currently.

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -0,0 +1,255 @@
...
+                            worker_url,
+                            self.context.port,
+                        )
+                        return True
+                    else:
+                        logger.error("Failed to remove worker %s: %s", worker_url, response_data)
+                        return False
+                else:
+                    logger.error(
+                        "Failed to remove worker %s: status %d, %s",
```
jvstme, line 261: (nit) I'm seeing this error in `systemctl -u dstack.gateway`:

```
dstack._internal.proxy.gateway.model_routers.sglang - ERROR - Failed to remove worker http://127.0.0.1:10003: status 202, {"status":"accepted","worker_id":"http://127.0.0.1:10003","message":"Worker removal queued for background processing"}
```

To reproduce:
- Start a service with one replica.
- Scale it to two replicas.

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Already removed.

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
Bihan: Resolved. This was happening with all Add/Remove call. This happened because I was expecting status 200, while Add/Remove worker apis responded with 202.

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/base.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/model_routers/__init__.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Resolved

src/dstack/_internal/proxy/gateway/model_routers/sglang.py
Bihan: Resolved

PR #3275, opened 2025-11-09
 [Feature]: Make proxy service HTTP timeout configurable

### LINKED ISSUE #3274: [Feature]: Make proxy service HTTP timeout configurable
### Title
Make the proxy service HTTP timeout configurable

### Problem / Motivation
The in-server proxy currently hard-codes a 60-second timeout when forwarding requests to service replicas. Long‑running model inference calls can exceed that, causing 504s even though replicas are still working. Users need a way to raise or lower that limit based on their workloads.

### Solution

- Introduce an environment variable (e.g., `DSTACK_SERVICE_CLIENT_TIMEOUT`) that controls the timeout used by `get_service_replica_client/ServiceConnection`.
- Default it to the existing `60s`, so behavior doesn’t change unless configured.
- Document the new knob in the server environment variable reference.

### Workaround

There isn’t one; the timeout is hard-coded today.

Yes

earandap:
###  Summary
  - Allow the proxy service clients to read their timeout from `DSTACK_SERVICE_CLIENT_TIMEOUT`, falling back to the previous 60s default so users can tune it without code changes.
  - Document the new environment variable in the server reference so the knob is discoverable.
  
 Closes #3274


r4victor:
Thanks!

PR #3276, opened 2025-11-11
[runner] Allow to compile runner on macOS (for development purposes only)

peterschmidt85:


runner/.justfile
```diff
@@ -56,7 +56,12 @@ build-shim-binary:
     cd {{source_directory()}}/cmd/shim
     if [ -n "$shim_os" ] && [ -n "$shim_arch" ]; then
         echo "Building shim for $shim_os/$shim_arch"
-        GOOS=$shim_os GOARCH=$shim_arch go build -ldflags "-X 'main.Version=$version' -extldflags '-static'"
+        if [ "$shim_os" = "linux" ] && [ "$(uname -s)" != "Linux" ]; then
+            echo "WARNING: Cross-compiling to Linux, disabling CGO (DCGM unavailable)"
+            CGO_ENABLED=0 GOOS=$shim_os GOARCH=$shim_arch go build -ldflags "-X 'main.Version=$version' -extldflags '-static'"
+        else
+            GOOS=$shim_os GOARCH=$shim_arch go build -ldflags "-X 'main.Version=$version' -extldflags '-static'"
```
un-def: 1. I'd add an explicit `CGO_ENABLED=1` to override the var if it's already set externally.
2. There should be no `-extldflags '-static'` when CGO is enabled (static linking of `libdl` is very tricky and won't work OOTB; `libdl` is used by DCGM Go wrapper to load `libdcgm.so`).

un-def (approved):

PR #3304, opened 2025-11-19
[Runpod] Require CUDA 12.8+ on the host 

### LINKED ISSUE #3299: [Runpod] Require CUDA 12.8+ on the host
Original request:

```
### Problem

There is no option currently to specify the cuda version of the instance on runpod backend.

### Solution

Use allowed_cuda_versions parameter on the runpod api to have the option.

### Workaround

_No response_

No
```

Updated description:

### Problem

Currently, gpuhut returns all offers, including the offers that come with too old CUDA version, e.g. 11.x. Because dstack doesn't allow to filter that, it can be that sometimes a user is using 12.4 CUDA version inside the Docker image but the host can have, e.g. 11.x version.

### Solution

Filter out the offers withe the hosts that have too old CUDA version

peterschmidt85:
Fixes #3299

> [!WARNING]
> Note, this PR requires https://github.com/dstackai/gpuhunt/pull/199

### DIFF
--- a/src/dstack/_internal/core/backends/runpod/api_client.py
+++ b/src/dstack/_internal/core/backends/runpod/api_client.py
@@ -3,6 +3,7 @@ from datetime import timedelta
 from typing import Any, Dict, List, Optional
 
 import requests
+from gpuhunt.providers.runpod import RunpodProvider
 from requests import Response
 
 from dstack._internal.core.errors import BackendError, BackendInvalidCredentialsError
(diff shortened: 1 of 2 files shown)


src/dstack/_internal/core/backends/runpod/api_client.py
jvstme, line 1: Should we also pass the CUDA version when creating clusters? See `_generate_create_cluster_mutation`

jvstme (approved):

src/dstack/_internal/core/backends/runpod/api_client.py
jvstme: If necessary, we can add it later in a separate PR, of course

peterschmidt85: I pass it here: https://github.com/dstackai/dstack/blob/4f9cfeeaeae77f885a8cb4d63a816d04ae3dd51e/src/dstack/_internal/core/backends/runpod/api_client.py#L479

Please confirm

jvstme: This line is within `_generate_pod_deployment_mutation`, which is only used for individual pods. My comment is about clusters.

Anyways, you merged the `gpuhunt` PR without adding a filter for cluster offers, so it's better not to add it in `dstack` now

jvstme (approved):

Issue #3290, opened 2025-11-13
[Feature]: Events (formerly Audit Logs)


jvstme:
https://docs.google.com/document/d/1TChZyLjETnSF5mkTXEXBuO2m-FL-E8ceTI1lkHongSg

## Implementation plan
- [x] Event storage and emission framework
- [x] `/api/events/list` API
- [ ] CLI
  - [x] `dstack event`
  - [x] `dstack event --watch`
  - [ ] Event log when provisioning resources
- [ ] Event coverage
  - [ ] Runs
    - [x] Creation
    - [x] Status changes
    - [x] Deletion
    - [x] In-place update
    - [ ] Service scaling
    - [ ] Rolling deployment
    - [x] Service registration
  - [ ] Jobs
    - [x] Creation
    - [x] Status changes
    - [x] In-place update
    - [x] Reachable/unreachable changes
    - [ ] Probe status
    - [x] Replica registration
  - [ ] Fleets
    - [x] Creation
    - [x] Status changes
    - [ ] In-place update
  - [ ] Instances
    - [x] Creation
    - [x] Status changes
    - [x] Reachable/unreachable changes
    - [x] Health changes
    - [ ] shim/runner updates
  - [x] Users
    - [x] Creation
    - [x] Update
    - [x] Deletion
  - [ ] Projects
    - [x] Creation
    - [x] Update
    - [x] Deletion
    - [ ] Membership changes
  - [ ] Volumes
    - [x] Creation
    - [x] Status changes
    - [x] Deletion
    - [ ] Attachment/detachment
      - [ ] To/from existing instance
      - [ ] To/from instance being created/deleted (RunPod)
  - [x] Gateways
  - [x] Secrets
  - [ ] Backends

PR #3325, opened 2025-11-26
Events emission framework and API

jvstme:
- Add the framework for event emission
- Add the events API
- Add a few sample events about project / user / fleet / instance / run / job creation

Hidden behind the `DSTACK_FF_EVENTS` feature flag.

Part of #3290

Next steps:
- CLI
- UI
- Actual event coverage
- More event targets (volumes, gateways, backends, secrets; possibly repos and files)


src/dstack/_internal/server/services/events.py
```diff
@@ -0,0 +1,327 @@
...
+                )
+    order_by = (EventModel.recorded_at.desc(), EventModel.id)
+    if ascending:
+        order_by = (EventModel.recorded_at.asc(), EventModel.id.desc())
+    query = (
+        select(EventModel)
+        .order_by(*order_by)
+        .limit(limit)
+        .options(
+            joinedload(EventModel.targets),
```
r4victor, line 373: IIUC, targets can potentially be from different projects. The access filter checks that an event has at least one target in the user's project but the response includes all the targets. Thus, event responses can leak targets from projects that users don't have access to.

jvstme: This may be expected in some cases.

Example: job assigned to instance, where the job and the instance are in different projects (will be possible once we allow sharing instances between projects, which is on the Q4 roadmap).
- Users in the job's project may need to know what instance it is assigned to.
- Users in the instance's project may need to know what jobs are assigned to it.

When this is not expected, it may be possible to split the event into several events with different targets. I don't think we'll have such cases anytime soon, but I described this behavior in the `events.emit` docstring.

src/dstack/_internal/core/models/events.py
```diff
@@ -18,16 +18,50 @@ class EventTargetType(str, Enum):
...
-    type: str  # Holds EventTargetType; str for adding new types without breaking compatibility
-    project_id: Optional[uuid.UUID]
-    id: uuid.UUID
-    name: str
+    type: Annotated[
+        str,  # not using EventTargetType to allow adding new types without breaking compatibility
+        Field(
+            description=(
+                f"Type of the target entity."
+                f" One of: {', '.join([f'`{t}`' for t in EventTargetType])}"
```
r4victor: You can use list_enum_values_for_annotation().

r4victor (approved):

PR #3335, opened 2025-11-28
[Feature]: Add JSON output option for CLI commands

### LINKED ISSUE #3322: [Feature]: Add JSON output option for CLI commands
### Problem

Currently, dstack CLI commands output data in a human-readable text format that is difficult to parse programmatically. When trying to extract specific information or integrate dstack commands into scripts and automation workflows, parsing the text output with tools like `awk`, `grep`, or `sed` is unreliable and cumbersome. Different commands may have varying output formats, making it challenging to build robust automation around dstack CLI operations.

### Solution

Add a global `--json` flag (or `--output json`) to dstack CLI commands that would output structured data in JSON format. This would:
1. Provide a stable, machine-readable output format for all commands
2. Make it easy to parse command output in scripts using standard JSON parsers
3. Enable reliable automation and integration with other tools
4. Follow common CLI best practices (similar to tools like `docker`, `kubectl`, `gh`, etc.)

```shell
dstack ps --json
dstack offer --json
dstack gateway list --json
```

### Workaround

Currently, the only option is to parse text output using shell utilities, for example:

```shell
gateway_list=$(dstack gateway list --verbose 2>&1 || echo "")

result=$(echo "${gateway_list}" | awk -v prefix="${GATEWAY_PREFIX}" '
          $1 ~ "^" prefix && $1 != "NAME" {
            gateway_name = $1
            domain = ""

            # Search through all fields for something that looks like a domain name
            # Domain should contain dots but NOT match IP address pattern
            for (i = 2; i <= NF; i++) {
              field = $i
              # Check if field contains dot and is not an IP address
              if (field ~ /\./ && field !~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/) {
                # Skip truncated values and header-like values
                if (field !~ /…$/ && field !~ /\.\.\.$/ && field !~ /^[A-Z]+$/) {
                  domain = field
                  break
                }
              }
            }

            if (domain != "") {
              print gateway_name "," domain
              exit
            }
          }
        ')
```

This approach is fragile, breaks when output formatting changes, and requires different parsing logic for each command.

No. I would gladly help out the dstack team with this one, but realistically I won't have time for that.

peterschmidt85:
Fixes #3322

This is a not-backward compatible change. It refactors existing `--json` code to make it eaasy to maintain and keep it extensible and backward-compatible in the future.


peterschmidt85:
@r4victor Please check if this approach is OK. If so, I can implement similar approach for `dstack gateway list` and `dstack ps`.

src/dstack/_internal/cli/models/offer.py
```diff
@@ -0,0 +1,19 @@
+from typing import List, Literal
+
+from dstack._internal.core.models.common import CoreConfig, generate_dual_core_model
+from dstack._internal.core.models.runs import Requirements
+from dstack._internal.server.schemas.gpus import GpuGroup
```
r4victor: cli cannot import from server. Models should be moved to core if they are used both in cli and server.

src/dstack/_internal/cli/models/offer.py
```diff
@@ -0,0 +1,19 @@
...
+
+class OfferCommandOutputConfig(CoreConfig):
+    json_dumps = pydantic_orjson_dumps_with_indent
+
+
+class OfferCommandOutput(generate_dual_core_model(OfferCommandOutputConfig)):
+    """JSON output model for `dstack offer` command with GPU grouping."""
+
+    project: str
+    requirements: Requirements
```
r4victor: A note on using Requirements – it's internal model used in job_spec/compute interface – you'll see more and more "technical" fields being added there with time, so it may be best to expose a different model if you need a user-facing model.

src/dstack/_internal/cli/utils/gpu.py
```diff
@@ -18,49 +24,14 @@ def print_gpu_json(gpus, run_spec, group_by_cli, api_project):
         reservation=run_spec.configuration.reservation,
```
r4victor, line 24: reservation can be specified via profile

r4victor (commented):
I suppose print_offers_json should be refactored in the same manner?

peterschmidt85:
@r4victor Fixed, please review

src/dstack/_internal/cli/utils/gpu.py
peterschmidt85: Should it be `run_spec.configuration.reservation or run_spec.merged_profile.reservation`?

peterschmidt85:
^ Supported `--json` for `dstack ps` and `dstack gateway`

r4victor (approved):

PR #3342, opened 2025-12-04
Make autocreated fleets opt-in

### LINKED ISSUE #3059: Forbid creating new fleets on run apply
Part of #2973 

Currently, dstack fleets are optional for cloud backends, which means users have two ways of provisioning via dstack: using fleet configurations or run configurations. The proposal is to make provisioning UX fleet-first: Run configurations won’t be able to create new fleets but will only use fleets explicitly configured by users. Thus, all provisioning use cases will be unified with the same UX.

Runs that don't need clusters may use one default elastic fleet:

```yaml
type: fleet
name: default-fleet
nodes: 0..
```

Multi-node tasks require creating cluster fleets explicitly.

Plan

1. Display a CLI warning if a run creates fleet informing that future dstack version will require creating fleets explicitly.
2. Forbid auto-created fleets and display a CLI error if there is no eligible fleet for the run. For non-multi-node runs, also suggest creating default elastic fleet.

r4victor:
Closes #3059 

The PR replaces opt-out feature flag DSTACK_FF_AUTOCREATED_FLEETS_DISABLED with opt-in DSTACK_FF_AUTOCREATED_FLEETS_ENABLED so that the default UX is fleet-first without autocreated fleets but legacy behavior can be restored by setting DSTACK_FF_AUTOCREATED_FLEETS_ENABLED.

**Breaking changes**

This is a breaking change that will be released in 0.20.0.

### DIFF
--- a/src/dstack/_internal/cli/services/configurators/run.py
+++ b/src/dstack/_internal/cli/services/configurators/run.py
@@ -64,7 +64,6 @@ from dstack._internal.utils.logging import get_logger
 from dstack._internal.utils.nested_list import NestedList, NestedListItem
 from dstack._internal.utils.path import is_absolute_posix_path
 from dstack.api._public.runs import Run
-from dstack.api.server import APIClient
 from dstack.api.utils import load_profile
 
 _KNOWN_AMD_GPUS = {gpu.name.lower() for gpu in gpuhunt.KNOWN_AMD_GPUS}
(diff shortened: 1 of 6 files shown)


peterschmidt85:
Screenshot 2025-12-04 at 13 15 12

@r4victor 
Do you think we could in case of no fleets configured explicitely tell about it?

peterschmidt85:
> Do you think we could in case of no fleets configured explicitely tell about it?

Oh, I see that we show a short error message at the end:

Screenshot 2025-12-04 at 13 16 49

Maybe just adjust the message to make it more ellaborative and give a link to Fleets page. 

@r4victor Shall I adjust the message?

r4victor:
@peterschmidt85, submit a separate PR.

PR #3355, opened 2025-12-08
Clean up low- and high-level API clients

un-def:
* Deprecate and ignore RunCollection.get_run_plan() repo_dir argument
* Ensure UsersAPIClient.get_my_user() returns UserWithCreds
* Ensure UserSSHKeyManager.get_user_key() returns user SSH key stored on the server
* Drop SSH key-related support code for pre-0.20.0 servers

### DIFF
--- a/src/dstack/api/_public/runs.py
+++ b/src/dstack/api/_public/runs.py
@@ -9,7 +9,7 @@ from contextlib import contextmanager
 from copy import copy
 from datetime import datetime
 from pathlib import Path
-from typing import BinaryIO, Dict, Iterable, List, Optional
+from typing import BinaryIO, Dict, Iterable, List, Optional, Union
 from urllib.parse import urlencode, urlparse
 
 from websocket import WebSocketApp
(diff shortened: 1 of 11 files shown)


jvstme (approved):

src/dstack/api/_public/runs.py
```diff
@@ -497,20 +503,11 @@ def get_run_plan(
...
-                if not config_manager.dstack_key_path.exists():
-                    generate_rsa_key_pair(private_key_path=config_manager.dstack_key_path)
-                logger.warning(
-                    f"Using legacy [code]{config_manager.dstack_key_path.with_suffix('.pub')}[/code]."
-                    " You will only be able to attach to the run from this client."
-                    " Update the [code]dstack[/] server to [code]0.19.34[/]+ to switch to user keys"
-                    " automatically replicated to all clients.",
-                )
-                ssh_key_pub = config_manager.dstack_key_path.with_suffix(".pub").read_text()
+            # Ensure we have a fresh key locally
```
peterschmidt85: @un-def Sorry, just to confirm, why do we need the key locally when submitting a run? I thought we only need it when attaching to a run, no?

un-def: Right, both `/get_plan` and `/apply` endpoints generate the user key if missing, we don't need to trigger key generation manually. I've removed this redundant API call, added `/get_plan`/`apply` tests, and simplified server code a bit : https://github.com/dstackai/dstack/pull/3355/commits/329e993de43f6512874c6eb7b93719cf0ebbb8f7

PR #3356, opened 2025-12-08
[Docs] Reflect the 0.20 changes related to `working_dir` and `repo_dir`

peterschmidt85:


peterschmidt85:
@un-def Please review. If this is OK, then I'll add the same for tasks and services

docs/docs/concepts/dev-environments.md
```diff
@@ -340,7 +340,7 @@ type: dev-environment
 name: vscode    
 
 files:
-  - ../examples  # Maps `examples` (the parent directory of `.dstack.yml`) to `/workflow/examples`
+  - ../examples  # Maps the parent directory of `.dstack.yml` to `<working dir>/../examples`
```
un-def, line 343: Actually, `examples` is a sibling of the parent dir:

```
<grandparent>/     -> ".."
  examples/        -> "../examples"
  <parent>/        -> "."
    .dstack.yml
    <sibling>/      -> "./sibling" or just "sibling"
```

I'd rather avoid such complicated examples, esp. with the omitted container path. I don't think someone really wants to put the `examples` dir alongside `<working dir>`. If you need an example with `..`, consider `../examples:./examples` (puts `examples` _inside_ `<working dir>`).

docs/docs/concepts/dev-environments.md
```diff
@@ -393,16 +390,30 @@ The local path can be either relative to the configuration file or absolute.
     name: vscode    
 
     repos:
-      # Mounts the parent directory of `examples` (must be a Git repo)
-      #   to `/my-repo`
+      # Clones the repo in the parent directory (`examples/..`) to `/my-repo`
```
un-def, line 393: “Parent directory” is somewhat ambiguous. I understand that you mean "the parent directory [of the configuration directory]”, where “configuration directory” is a directory where `.dstack.yml` is located, but not sure if it's clear enough for the reader.

docs/docs/concepts/dev-environments.md
```diff
@@ -366,24 +366,21 @@ type: dev-environment
 name: vscode    
 
 repos:
-  # Mounts the parent directory of `examples` (must be a Git repo)
-  #   to `/workflow` (the default working directory)
+  # Clones the repo from the parent directory (`examples/..`) to `<working dir>`
```
un-def, line 369: > to `<working dir>`

Then it should be `..:.`

un-def (approved):

docs/docs/concepts/dev-environments.md
peterschmidt85: > The container path is optional. If not specified, it will be automatically calculated:

My idea has been to give exact examples of calculating the path if it's not specified, and especially pointing to the non-trivial fact that if they use `..` without specified a path, it will be concatenated.

docs/docs/concepts/dev-environments.md
peterschmidt85: Yes, I often mix up the current directory (`.`) and the parent directory (`..`) myself. If that's OK with you, I plan to address this separately.

docs/docs/concepts/dev-environments.md
peterschmidt85: @un-def Why? I suppose for `repos`, `..` is the same as `..:.` (`..:<working dir>`).

un-def: Sorry, I overlooked that it's `repos`, not `files`, you are right

PR #3365, opened 2025-12-10
Add the `dstack event` command for viewing events

jvstme:
Note: `dstack event --watch` is not part of this PR and will be added later.

Part of #3290

### DIFF
--- a/src/dstack/_internal/cli/services/events.py
+++ b/src/dstack/_internal/cli/services/events.py
@@ -1,0 +1,63 @@
+import uuid
+from collections.abc import Iterator
+from dataclasses import asdict, dataclass
+from datetime import datetime
+from typing import Optional
+
+from rich.text import Text
+
+from dstack._internal.cli.utils.common import console
+from dstack._internal.core.models.events import Event, EventTargetType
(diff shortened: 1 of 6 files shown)


src/dstack/_internal/cli/services/events.py
```diff
@@ -0,0 +1,63 @@
...
+from collections.abc import Iterator
+from dataclasses import asdict, dataclass
+from datetime import datetime
+from typing import Optional
+
+from rich.text import Text
+
+from dstack._internal.cli.utils.common import console
+from dstack._internal.core.models.events import Event, EventTargetType
+from dstack._internal.server.schemas.events import LIST_EVENTS_DEFAULT_LIMIT
```
r4victor, line 11: Pls avoid importing from server in cli. Move to core if needed both for server and cli.

r4victor (approved):

src/dstack/_internal/cli/services/events.py
r4victor: Although api already imports from `schemas` everywhere, so never mind.

PR #3376, opened 2025-12-12
[Docs] Merge GCP clusters examples

peterschmidt85:
This is required as we agreed to add more cloud-specific examples under `Clusters`.


examples/clusters/gcp/README.md
r4victor, line 1: The merged guide tells how to run nccl tests but not how to run any other tasks.

Issue #3309, opened 2025-11-20
Events UI


jvstme:
Implement the "UI" section from the parent issue design document.

See the "Main concepts" and "API" sections for context.

## Implementation plan

- [x] Events page
- [ ] References from resource pages to the events page
  - [x] Run page
  - [x] Job page
  - [x] Fleet page
  - [x] Project page (the `within_projects` filter)
  - [x] User page
    - [x] Reference to user events (the `target_users` filter)
    - [x] Reference to user activity (the `actors` filter)

peterschmidt85:
@olgenn Must be misunderstanding - we need Events tab - similar to Run/Fleet pages. And the tab will contain a table and a link.

peterschmidt85:
We don't need a link only.

peterschmidt85:
Plus, for the Project page, we need to rework it from sections to Tabs.

olgenn:
Okey. I will add it

PR #3392, opened 2025-12-16
Was implemented Event list for job, run and fleet

### LINKED ISSUE #3399: [Bug]: Duplicate events shown in the UI
### Steps to reproduce

1. Create at least 21 events.
2. View them all on the Events page by scrolling down.

### Actual behaviour

Every 20-event batch (except the top 20 events) is duplicated.

For example, this screenshot shows the start of a duplicated batch. You can see that instead of descending gradually, timestamps suddenly just to a point of time that was already shown above.

Image

Additionally, every request to `/api/events/list` is duplicated with the same parameters as you are scrolling.

https://github.com/user-attachments/assets/e6a7c257-3a80-4171-8fe3-ebd14f226ee8

### Expected behaviour

No events are duplicated.

### dstack version

0.20.0

### Server logs

```shell

```

### Additional information

The same issue is present in event tables added on Run, Job, and Fleet pages in #3392.

olgenn:
#3309
Close #3399

### DIFF
--- a/frontend/src/hooks/useInfiniteScroll.ts
+++ b/frontend/src/hooks/useInfiniteScroll.ts
@@ -85,7 +85,9 @@ export const useInfiniteScroll = <DataItem, Args extends InfinityListArgs>({
             console.log(e);
         }
 
-        isLoadingRef.current = false;
+        setTimeout(() => {
+            isLoadingRef.current = false;
+        }, 10);
     };
 
     useLayoutEffect(() => {
@@ -108,6 +110,7 @@ export const useInfiniteScroll = <DataItem, Args extends InfinityListArgs>({
         const scrollPositionFromBottom = element.scrollHeight - (element.clientHeight + element.scrollTop);
 
         if (scrollPositionFromBottom < SCROLL_POSITION_GAP) {
+            console.log('test', element.scrollHeight);
             getMore().catch(console.log);
         }
     }, [disabledMore, getMore]);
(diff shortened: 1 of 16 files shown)


jvstme (commented):
@olgenn, thanks for the PR. Based on my testing and our discussion with the team, I can suggest the following changes:
1. Add a button above the event tables, for example “Full view”, that opens the Events page with a preconfigured filter.
1. On the Events tab of the run page, use the `within_runs` filter (instead of `target_runs`). This is more useful for most cases. The remaining cases will be covered by navigating to the Events page via a link. Navigation via the button should also set `within_runs`.
1. Similarly, on the fleet page, use the `within_fleets` filter (instead of `target_fleets`). Navigation via the button should also set `within_fleets`.
1. Show the Events tab on the run page regardless of the number of jobs (it is currently shown only for single-job runs).
1. Fix #3399

olgenn:
@jvstme fixed

frontend/src/hooks/useInfiniteScroll.ts
```diff
@@ -106,6 +110,7 @@ export const useInfiniteScroll = <DataItem, Args extends InfinityListArgs>({
         const scrollPositionFromBottom = element.scrollHeight - (element.clientHeight + element.scrollTop);
 
         if (scrollPositionFromBottom < SCROLL_POSITION_GAP) {
+            console.log('test', element.scrollHeight);
```
jvstme: (nit) Do we need this in production?

jvstme (approved):
🔥

frontend/src/hooks/useInfiniteScroll.ts
olgenn: I removed it

jvstme (approved):

PR #3406, opened 2025-12-19
[UX] Make "No fleets" run status more explicit #3405

### LINKED ISSUE #3405: [UX] Make "No fleets" run status more explicit
Problem: 
Currently `No offers` is shown in both cases, when `dstack` can't find matching offers and when no fleets found. Those are two totally different situations that require different action.

Solution:

**1. Warning**
On `dstack apply`, if the run plan has no offers, check if fleets exist, and if no fleets exist, instead of 

```
No matching instance offers available. Possible reasons: https://dstack.ai/docs/guides/troubleshooting/#no-offers
```

Print:

```
The project has no fleets. Create one before submitting a run: https://dstack.ai/docs/concepts/fleets"
```

**2. Status**
In case job termination reason is no capacity and termination reason message starts with `No fleet found`, in `dstack ps` and UI, instead of `no offers`, show `no fleets`.

peterschmidt85:
Open

### DIFF
--- a/src/dstack/_internal/cli/utils/common.py
+++ b/src/dstack/_internal/cli/utils/common.py
@@ -32,6 +32,12 @@ NO_OFFERS_WARNING = (
     " https://dstack.ai/docs/guides/troubleshooting/#no-offers"
     "[/]\n"
 )
+NO_FLEETS_WARNING = (
+    "[warning]"
+    "The project has no fleets. Create one before submitting a run:"
+    " https://dstack.ai/docs/concepts/fleets"
+    "[/]\n"
+)
 
 
 def cli_error(e: DstackError) -> CLIError:
--- a/frontend/src/libs/run.ts
+++ b/frontend/src/libs/run.ts
@@ -39,7 +39,11 @@ export const getStatusIconType = (
 export const getStatusIconColor = (
     status: IRun['status'] | TJobStatus,
     terminationReason: string | null | undefined,
+    statusMessage: string,
 ): StatusIndicatorProps.Color | undefined => {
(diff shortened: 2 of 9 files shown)


src/dstack/_internal/cli/utils/common.py
```diff
@@ -32,6 +32,12 @@
     " https://dstack.ai/docs/guides/troubleshooting/#no-offers"
     "[/]\n"
 )
+NO_FLEETS_WARNING = (
+    "[warning]"
+    "The project has no fleets. Create one before submitting a run:"
```
r4victor, line 37: You set "no fleets" if there is no suitable fleet for the run but error message says "The project has no fleets". So you'll see this message even if the project has fleets but they are not suitable, e.g. there is a default fleet but no fleet for multi-node.

r4victor (changes requested):
The code relies on specific text in job_model.termination_reason_message and job status text – there should be a comment that changing them would break things. Otherwise, there is no way to know that.

src/dstack/_internal/cli/utils/common.py
peterschmidt85: @r4victor 
No fleets is only shown if the project indeed has no fleets at all https://github.com/dstackai/dstack/blob/cd8f73a036617bb2551df783d5b6d98ee89742f6/src/dstack/_internal/cli/services/configurators/run.py#L112

peterschmidt85:
> The code relies on specific text in job_model.termination_reason_message and job status text – there should be a comment that changing them would break things. Otherwise, there is no way to know that.

You're right. Added comment.

PR #3425, opened 2025-12-25
[UX] Add an API that returns projects that lack active fleets

peterschmidt85:
The `/api/projects/list` API now supports the `only_no_fleets` body parameter. If `only_no_fleets` is `True`: returns only projects where the user is a member and that have no active fleets.

Example:

```
curl -X POST http://127.0.0.1:3000/api/projects/list_only_no_fleets -H "Authorization: Bearer cea07601-88a2-4709-87b8-d2fc29bd9f70" -H "Content-Type: application/json" -H "X-API-VERSION: latest" -d '{}' | jq
```


src/dstack/_internal/server/routers/projects.py
```diff
@@ -43,13 +43,23 @@ async def list_projects(
...
 
     `members` and `backends` are always empty - call `/api/projects/{project_name}/get` to retrieve them.
     """
     if body is None:
         # For backward compatibility
         body = ListProjectsRequest()
+    if body.only_no_fleets:
+        return CustomORJSONResponse(
+            await fleets.list_projects_with_no_active_fleets(session=session, user=user)
+        )
```
r4victor: If this is a different function, it won't be working with other request parameters like include_not_joined. I'd just make a new endpoint.

r4victor:
How is this going to be used exactly?

peterschmidt85:
> How is this going to be used exactly?

UI will invoke this API to understand where to show warnings on the Runs/Fleets/Project settings pages. The user will be able to dismiss these warnings.

src/dstack/_internal/server/routers/projects.py
peterschmidt85: Yes, I thought of it. Will make it a separate endpoint then.

PR #3488, opened 2026-01-22
[UX] Make `dstack project` and `dstack project set-default` interactive for default project selection

peterschmidt85:
If invoked in the interactive mode, both `dstack project` and `dstack project set-default` allow the user to select/change the default project interactively via ꜛꜜ Enter


r4victor (approved):

peterschmidt85:
@r4victor FYI, replaced the library to support Windows too

PR #3497, opened 2026-01-26
Set INSTANCE_UNREACHABLE for unreachable on-demand instances

r4victor:
Previously, if an instance is killed (e.g. a memory issue) while the job was running, the run got the `interrupted` status message.

This PR changes this for on-demand instances from `interrupted` to `error` along with a proper error message `instance unreachable`. Example:

```
Submit a new run? [y/n]: y
 NAME           BACKEND        RESOURCES             INSTA…  PRICE  STAT…  PROB…  SUBM…  ERROR
 stale-dingo-1  gcp            cpu=2 mem=8GB         e2-st…  $0.0…  error         14:10  instance un…
                (us-east5)     disk=100GB

stale-dingo-1 provisioning completed (failed)
Error (Instance unreachable)
```


peterschmidt85:
Also, ideally, ensure that the status_message will be calculated and shown right

PR #3500, opened 2026-01-26
Add gateway lifecycle events

jvstme:
- Gateway created
- Gateway status changed
- Gateway deleted
- Gateway set as default
- Gateway unset as default
- Gateway wildcard domain changed

#3290
Stacked on #3499

### DIFF
--- a/src/dstack/_internal/server/services/gateways/__init__.py
+++ b/src/dstack/_internal/server/services/gateways/__init__.py
@@ -419,6 +419,7 @@ async def get_project_gateway_model_by_name_for_update(
             res = await session.execute(
                 select(GatewayModel)
                 .where(GatewayModel.id.in_([gateway_id]), *filters)
+                .options(selectinload(GatewayModel.gateway_compute))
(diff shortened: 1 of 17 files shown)


src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -343,6 +392,39 @@ async def get_project_gateway_model_by_name(
...
+        yield None
+    else:
+        async with get_locker(get_db().dialect_name).lock_ctx(
+            GatewayModel.__tablename__, [gateway_id]
+        ):
+            # Refetch after lock
+            res = await session.execute(
+                select(GatewayModel)
+                .where(GatewayModel.id.in_([gateway_id]), *filters)
+                .options(selectinload(GatewayModel.gateway_compute))
```
r4victor: Better to use joinedload for one-to-one relationships to avoid an extra query with selectinload.  with_for_update(of=GatewayModel) to lock only the main table.

r4victor (approved):

PR #3503, opened 2026-01-27
Switch UI to pagination-based projects and users API 

### LINKED ISSUE #3490: Switch UI to pagination-based projects and users API
#3489 implemented cursor-based pagination API for `/api/project/list` and `/api/users/list`. The UI should switch to the new API for performance.

All the UI functionality is supported via the new API (displaying total count, filtering by name, etc). The only UI change is it won't longer be possible to open a specific page – it'll be prev/next navigation as it's usually done with cursor-based pagination.

olgenn:
Close #3490

### DIFF
--- a/frontend/webpack/dev.js
+++ b/frontend/webpack/dev.js
@@ -4,7 +4,7 @@ const { srcDir } = require('./env');
 const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
 const CircularDependencyPlugin = require('circular-dependency-plugin');
 
-const port = 3002;
+const port = 3000;
 
 module.exports = {
     mode: 'development',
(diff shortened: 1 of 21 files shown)


frontend/webpack/dev.js
```diff
@@ -4,7 +4,7 @@ const { srcDir } = require('./env');
 const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
 const CircularDependencyPlugin = require('circular-dependency-plugin');
 
-const port = 3000;
+const port = 3002;
```
r4victor: How about reading the port from the env var to avoid changing the code.

r4victor (changes requested):
Projects/Users pages no longer display total number of resources – this is useful to know, especially on Sky. You can get it by specifying `return_total_count: true`. Not that response type changes in that case to return `total_count` besides the list.

I'd also replace infinite scroll pagination with prev/next pages pagination – scroll pagination triggers many requests quickly by preloading, which may not be necessary, but you can leave it as is since we already do scroll pagination on other pages.

r4victor (approved):

PR #3521, opened 2026-01-29
[Feature]: Show probe statuses in the UI 

### LINKED ISSUE #3204: [Feature]: Show probe statuses in the UI
### Problem

Service [probe](https://dstack.ai/docs/concepts/services/#probes) statuses are only shown in the CLI, so it's difficult or impossible to track the service health and the rolling deployment progress from the UI.

### Solution

In the "Jobs" section of the run page, show probe statuses for each job similar to how they are [shown](https://github.com/dstackai/dstack/blob/d4e0c75e67ead27ae99cb076a6a6aa1bf58876fc/src/dstack/_internal/cli/utils/run.py#L277) in the CLI.

Image

If there is only one job and the "Jobs" section is hidden, show probe statuses in the "General" section.

### Workaround

_No response_

Yes

olgenn:
Close #3204

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -12,7 +12,7 @@
         "@cloudscape-design/chat-components": "^1.0.62",
         "@cloudscape-design/collection-hooks": "^1.0.74",
         "@cloudscape-design/component-toolkit": "^1.0.0-beta.120",
-        "@cloudscape-design/components": "^3.0.1091",
+        "@cloudscape-design/components": "^3.0.1188",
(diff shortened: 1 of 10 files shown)


peterschmidt85 (changes requested):
I suggest at least to use `StatusIndicator`: with `not-started` and `success`?

olgenn:
@peterschmidt85 Done

olgenn:
I also added underline for links

jvstme (approved):
Works as expected 👍

PR #3524, opened 2026-01-29
[Services] Add default probes if model is set 

### LINKED ISSUE #3522: [Services] Add default `probes` if `model` is set
The following `probes` can be used by default if the user specified `model` but hasn't specified `probes`:

```
probes:
  - type: http
    method: post
    url: /v1/chat/completions
    headers:
      - name: Content-Type
        value: application/json
    body: |
      {
        "model": "<model property name>",
        "messages": [{"role": "user", "content": "hi"}],
        "max_tokens": 1
      }
    timeout: 30s
```

peterschmidt85:
Fixes #3522

### DIFF
--- a/src/dstack/_internal/core/models/configurations.py
+++ b/src/dstack/_internal/core/models/configurations.py
@@ -56,6 +56,8 @@ DEFAULT_PROBE_READY_AFTER = 1
 DEFAULT_PROBE_METHOD = "get"
 MAX_PROBE_URL_LEN = 2048
 DEFAULT_REPLICA_GROUP_NAME = "0"
+DEFAULT_MODEL_PROBE_TIMEOUT = 30
+DEFAULT_MODEL_PROBE_URL = "/v1/chat/completions"
 
 
 class RunConfigurationType(str, Enum):
@@ -851,9 +853,9 @@ class ServiceConfigurationParams(CoreModel):
     ] = None
     rate_limits: Annotated[list[RateLimit], Field(description="Rate limiting rules")] = []
     probes: Annotated[
-        list[ProbeConfig],
+        Optional[list[ProbeConfig]],
         Field(description="List of probes used to determine job health"),
-    ] = []
+    ] = None  # None = omitted (may get default when model is set); [] = explicit empty
 
     replicas: Annotated[
         Optional[Union[List[ReplicaGroup], Range[int]]],
@@ -895,7 +897,9 @@ class ServiceConfigurationParams(CoreModel):
         return v
 
     @validator("probes")
-    def validate_probes(cls, v: list[ProbeConfig]) -> list[ProbeConfig]:
+    def validate_probes(cls, v: Optional[list[ProbeConfig]]) -> Optional[list[ProbeConfig]]:
+        if v is None:
+            return v
         if has_duplicates(v):
             # Using a custom validator instead of Field(unique_items=True) to avoid Pydantic bug:
             # https://github.com/pydantic/pydantic/issues/3765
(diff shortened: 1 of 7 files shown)


src/dstack/_internal/core/models/configurations.py
```diff
@@ -932,6 +936,35 @@ def validate_replicas(
                 )
         return v
 
+    @root_validator()
+    def set_default_probes_for_model(cls, values):
```
jvstme: This sets the default on the client side, and we [prefer](https://github.com/dstackai/dstack/blob/6452158a1cfd33d8700719d8c769f697087d5422/src/dstack/_internal/server/services/runs/spec.py#L65) server-side defaults.

Consider handling the default probe in [`JobConfigurator._probes`](https://github.com/dstackai/dstack/blob/58701213f28a7e42787344f10197ca6a85bab3cd/src/dstack/_internal/server/services/jobs/configurators/base.py#L395) instead, so that the configuration always holds the initial value (`None`) and only the job spec holds the default probe.

src/dstack/_internal/core/models/configurations.py
```diff
@@ -932,6 +936,35 @@ def validate_replicas(
...
+        model = values.get("model")
+        probes = values.get("probes")
+        if model is not None and probes is None:
```
jvstme: (nit)

> `model is not None`

This check also passes for models that are declared using the [full syntax](https://dstack.ai/docs/reference/dstack.yml/service/#model_1) that are not OpenAI-compatible or not chat models (once we support those). Such models don't necessarily provide `/v1/chat/completions`, so the probe will not work.

A better check would be `isinstance(configuration.model, OpenAIChatModel)`.

src/dstack/_internal/core/models/configurations.py
```diff
@@ -851,9 +853,9 @@ class ServiceConfigurationParams(CoreModel):
     ] = None
     rate_limits: Annotated[list[RateLimit], Field(description="Rate limiting rules")] = []
     probes: Annotated[
-        list[ProbeConfig],
+        Optional[list[ProbeConfig]],
         Field(description="List of probes used to determine job health"),
```
jvstme: Consider documenting the default probe either here or in `model`. And optionally in `concepts/services.md`

src/dstack/_internal/core/models/configurations.py
```diff
@@ -851,9 +853,13 @@ class ServiceConfigurationParams(CoreModel):
...
     rate_limits: Annotated[list[RateLimit], Field(description="Rate limiting rules")] = []
     probes: Annotated[
-        list[ProbeConfig],
-        Field(description="List of probes used to determine job health"),
-    ] = []
+        Optional[list[ProbeConfig]],
+        Field(
+            description="The list of probes to determine service health. "
+            "If `model` is set, defaults to a `/v1/chat/completions` probe. "
+            "Set explicitly to override."
```
jvstme: (nit)

image


```suggestion
            "Set explicitly to override"
```

src/dstack/_internal/server/services/jobs/configurators/base.py
```diff
@@ -394,7 +398,13 @@ def _service_port(self) -> Optional[int]:
...
     def _probes(self) -> list[ProbeSpec]:
         if isinstance(self.run_spec.configuration, ServiceConfiguration):
-            return list(map(_probe_config_to_spec, self.run_spec.configuration.probes))
+            probes = self.run_spec.configuration.probes
+            if probes is not None:
+                return list(map(_probe_config_to_spec, probes))
+            # Generate default probe if model is set
+            model = self.run_spec.configuration.model
+            if model is not None:
+                return [_default_model_probe_spec(model.name)]
```
jvstme: (nit) My comment from the previous review is still relevant. Potentially, not every model is an OpenAI-compatible chat model, so not every model needs the default probe

src/dstack/_internal/server/services/jobs/configurators/base.py
```diff
@@ -394,7 +398,13 @@ def _service_port(self) -> Optional[int]:
 
     def _probes(self) -> list[ProbeSpec]:
         if isinstance(self.run_spec.configuration, ServiceConfiguration):
-            return list(map(_probe_config_to_spec, self.run_spec.configuration.probes))
+            probes = self.run_spec.configuration.probes
+            if probes is not None:
+                return list(map(_probe_config_to_spec, probes))
+            # Generate default probe if model is set
+            model = self.run_spec.configuration.model
+            if model is not None:
```
jvstme: (nit) This [comment](https://github.com/dstackai/dstack/pull/3524#discussion_r2754016370) is still relevant.

```suggestion
            if isinstance(model, OpenAIChatModel):
```




(this relies on `model: str` being [normalized](https://github.com/dstackai/dstack/blob/2acb6dee3db3960427ca346b7fa466b22e1fbcf4/src/dstack/_internal/core/models/configurations.py#L882) to `OpenAIChatModel`)

jvstme (approved):

jvstme (approved):

PR #3530, opened 2026-02-02
Add probe `until_ready` configuration option

jvstme:
This option allows to stop executing the probe
once it reaches the `ready_after` threshold of
successful executions. This is useful for heavier
probes that only need to run in the startup phase
and not during regular replica operation.


r4victor (commented):
Seems to be missing `until_ready` client exclude for compatibility with older servers.

r4victor (approved):

PR #3532, opened 2026-02-02
Events UI #3309 

olgenn:
Added links to event list from project and user details pages

image
image


olgenn:
@jvstme We need to add anything else?

jvstme (approved):
> We need to add anything else?

No, I think this covers the current requirements 👍

Although I see @peterschmidt85 in #3309 suggests to rework project and user events into tabs (probably still with the "Full view" button, like with other entities). But the current PR would already be useful in this week's release, so I think merging it as-is and reworking into tabs in a separate PR is also an option

peterschmidt85:
> > We need to add anything else?
> 
> No, I think this covers the current requirements 👍
> 
> Although I see @peterschmidt85 in #3309 suggests to rework project and user events into tabs (probably still with the "Full view" button, like with other entities). But the current PR would already be useful in this week's release, so I think merging it as-is and reworking into tabs in a separate PR is also an option

Just quickly chatted with @olgenn. Agreed to start with tabs instead of adding too many links to the top menu (especially since their visual style differs from our common approach). We'll work with @olgenn on this together.

peterschmidt85 (changes requested):
The goal is to make the Events tabs look consistent across pages.

peterschmidt85 (approved):

PR #3533, opened 2026-02-03
[CLI]: `dstack event --watch`

jvstme:
This adds the new `--watch` option to the `dstack event` command, which allows tracking events in realtime.

#3290


r4victor (approved):
🔥

PR #3548, opened 2026-02-05
Kubernetes: improve offers

### LINKED ISSUE #3481: Kubernetes: improve offers
Some of the problems:

* Node taints are ignored
* A run with `cpu: 2..4` requirement doesn't match a node with 8 CPUs (this makes sense for VM backends, but not for K8s)
* Resources allocated for other pods are not taken into account

un-def:
* Skip nodes with untolerated taints
* Take into account already allocated resources
* Set offer resources to the lower limit of the resource requirements ranges

Closes: https://github.com/dstackai/dstack/issues/3481

### DIFF
--- a/src/dstack/_internal/core/backends/kubernetes/resources.py
+++ b/src/dstack/_internal/core/backends/kubernetes/resources.py
@@ -1,0 +1,363 @@
+import dataclasses
+from collections.abc import Mapping
+from decimal import Decimal
+from enum import Enum
+from typing import Callable, Optional, Union, cast
+
+from gpuhunt import KNOWN_AMD_GPUS, KNOWN_NVIDIA_GPUS, AcceleratorVendor
+
+# XXX: kubernetes.utils is missing in the stubs package
+from kubernetes import utils as _kubernetes_utils  # pyright: ignore[reportAttributeAccessIssue]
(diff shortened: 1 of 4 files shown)


src/dstack/_internal/core/backends/kubernetes/resources.py
```diff
@@ -0,0 +1,363 @@
...
+            )
+            return []
+        return [gpu] * gpu_count
+    logger.debug("Node %s: no available GPU resource found", node_name)
+    return []
+
+
+def _get_nodes_allocated_resources(api: CoreV1Api) -> dict[str, KubernetesResources]:
+    nodes_allocated_resources: dict[str, KubernetesResources] = {}
+    for pod in api.list_pod_for_all_namespaces().items:
```
r4victor, line 343: "list" permission for "pods" needs to be added to the docs: https://github.com/dstackai/dstack/blob/db413d80fcd862172ea89121677639fe4ab6d034/docs/docs/concepts/backends.md?plain=1#L1052

r4victor:
I think we need to document somewhere in the docs how Kubernetes offers work since they're now very different from all other backends.

r4victor (approved):

PR #3558, opened 2026-02-10
Add pd disaggregated inference

Bihan:
**Testing Steps**
1. Create (CPU node) in K8s cluster 

2. Create gateway in the CPU node using below config
```
type: gateway
name: bihan-gateway

backend: kubernetes
region: any

domain: bihan-gateway.dstack.ai
router: sglang
```

3. Create GPU-node with 3 instances (1 Prefill, 1 Decode and 1 for testing scaling) in the same K8s cluster where gateway node exists. 
Note: See [design doc](https://docs.google.com/document/d/1trNdpCSSQdKufOMZIYcgH94QfvZZhrlfj_cBeWecpUc/edit?tab=t.0#heading=h.dpuoswhdv11h) for details on why the gateway and workers are required to be on the same network.

4. Apply below prefill-decode service configuration
```
type: service
name: prefill-decode
image: lmsysorg/sglang:latest

env:
  - HF_TOKEN
  - MODEL_ID=meta-llama/Llama-3.2-1B-Instruct

replicas:
  - count: 1..2
    scaling:
      metric: rps
      target: 3
    commands:
      - |
          python -m sglang.launch_server \
            --model-path $MODEL_ID \
            --disaggregation-transfer-backend mooncake \
            --host 0.0.0.0 \
            --port 8000 \
            --disaggregation-bootstrap-port 8998 \
            --log-level debug \
            > worker-server.log 2>&1
    resources:
      gpu: 1

  - count: 1
    commands:
      - |
          python -m sglang.launch_server \
            --model-path $MODEL_ID \
            --disaggregation-transfer-backend mooncake \
            --host 0.0.0.0 \
            --port 8000 \
            --log-level debug \
            > worker-server.log 2>&1
    resources:
      gpu: 1

port: 8000
model: meta-llama/Llama-3.2-1B-Instruct

probes:
  - type: http
    url: /health_generate
    interval: 15s

router:
  type: sglang
  policy: round_robin
  pd_disaggregation: true
```
5. When `rps>=3` prefill replica scales to 2.  

**Note:** For testing you need to assign [wheel](https://github.com/Bihan/dstack/blob/a90173b7612b5b695c9e380874c10cfe8e7d0f48/src/dstack/_internal/core/backends/base/compute.py#L1045) to `https://bihan-test-bucket.s3.eu-west-1.amazonaws.com/dstack_gateway-0.0.1-py3-none-any.whl`


src/dstack/_internal/core/models/configurations.py
```diff
@@ -885,6 +886,14 @@ class ServiceConfigurationParams(CoreModel):
             )
         ),
     ] = None
+    router_config: Annotated[
```
jvstme: (nit) `router_config` -> `router` for brevity and consistency with gateway configurations?

src/dstack/_internal/core/models/configurations.py
```diff
@@ -885,6 +886,14 @@ class ServiceConfigurationParams(CoreModel):
...
+        Optional[AnyRouterConfig],
+        Field(
+            description=(
+                "Router configuration for the service. Currently supports routing policy and pd_disaggregation. "
```
jvstme: (nit) The supported properties (routing policy and pd_disaggregation) should already be visible in the `AnyRouterConfig` reference, duplicating them here may lead to inconsistencies when adding or removing properties in the future

src/dstack/_internal/core/models/configurations.py
jvstme: Any new properties should be excluded from client requests for compatibility with older servers.

See `get_run_spec_excludes`.

src/dstack/_internal/core/models/routers.py
```diff
@@ -19,6 +19,30 @@ class SGLangRouterConfig(CoreModel):
             description="The routing policy. Options: `random`, `round_robin`, `cache_aware`, `power_of_two`"
         ),
     ] = "cache_aware"
+    pd_disaggregation: Annotated[
```
jvstme, line 40: Any new properties should be excluded from client requests for compatibility with older servers.

See `get_run_spec_excludes` and `_get_gateway_configuration_excludes`.

jvstme, line 40: This adds pd_disaggregation to both service configurations and gateway configurations. I'd advocate for adding it to service configurations only.

- Whether or not the service is configured to use PD disaggregation is clearly a service property, because it depends on the replica groups configuration. I don't think many users would want to configure that property at the gateway level, making assumptions about what services are going to run on that gateway in the future.

- Having two places for the same property complicates the interface — you'd have to explain in the docs how these places are related to each other, when and how one setting overrides the other, etc.

- Having the property at the gateway level can potentially complicate further development — that way, you can only tell whether a service is using PD disaggregation if the service is already registered, and you need to fetch the GatewayModel object from the database to do so. For example, this would complicate adding the default probe for services with PD disaggregation.

src/dstack/_internal/proxy/gateway/repo/state_v1.py
```diff
@@ -98,6 +98,7 @@ def parse_replica(replica: dict) -> Replica:
         ssh_destination=replica["ssh_host"],
         ssh_port=replica["ssh_port"],
         ssh_proxy=ssh_proxy,
+        internal_ip=replica.get("internal_ip"),
```
jvstme: (nit) Redundant, old `state.json` files never had the `internal_ip` property

```suggestion
```

src/dstack/_internal/proxy/lib/models.py
```diff
@@ -27,6 +27,7 @@ class Replica(ImmutableModel):
     # Optional outer proxy, a head node/bastion
     ssh_head_proxy: Optional[SSHConnectionParams] = None
     ssh_head_proxy_private_key: Optional[str] = None
+    internal_ip: Optional[str] = None
```
jvstme, line 30: (nit) Since the `Replica` model is used both on the gateway and in the in-server proxy, it's also worth populating `Replica.internal_ip` in [`ServerProxyRepo.get_service`](https://github.com/Bihan/dstack/blob/a90173b7612b5b695c9e380874c10cfe8e7d0f48/src/dstack/_internal/server/services/proxy/repo.py#L106).

Otherwise, if we'll ever need to use `internal_ip` in the common proxy logic, it will be difficult to spot that it's only set on the gateway, which may lead to bugs.

src/dstack/_internal/server/services/services/__init__.py
jvstme, line 1: Currently, it seems to be possible to run a service that defines `router_config.type == "sglang"` on a non-SGLang gateway or even without a gateway. I assume this either won't work (if the service requires PD disaggregation) or will just lead to unexpected behavior, since `router_config` will be ignored.

I'd suggest to enforce that services with `router_config.type == "sglang"` only run on SGLang gateways by raising a relevant exception in `_register_service_in_server` and  `_register_service_in_gateway`.

src/dstack/_internal/proxy/gateway/services/model_routers/base.py
```diff
@@ -89,3 +89,25 @@ def update_replicas(self, replica_urls: List[str]) -> None:
...
+            bootstrap_port: Bootstrap port for prefill workers (optional).
+
+        Returns:
+            True if the worker was accepted, False otherwise.
+        """
+        raise NotImplementedError
+
+    async def register_worker(self, url: str) -> bool:
+        """Register worker with one attempt (no polling). Returns True if ready and added."""
+        raise NotImplementedError
```
jvstme: (nit) These methods don't need to be in the `Router` base class, they are only ever called by `SglangRouter.update_replicas` and never by external callers. Consider removing them from `Router` and making private in `SglangRouter` (prefixed with `_`)

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -140,7 +141,7 @@ def remove_replicas(self, replica_urls: List[str]) -> None:
         for replica_url in replica_urls:
             self._remove_worker_from_router(replica_url)
 
-    def update_replicas(self, replica_urls: List[str]) -> None:
+    async def update_replicas(self, replica_urls: List[str]) -> None:
```
jvstme: The method is now async, but it still calls `_get_router_workers` and `_remove_worker_from_router`, which are synchronous and perform blocking I/O using `httpx.Client`. This can block the gateway's event loop for a long time, during which the gateway will be inoperable.

I'd suggest to revert the method back to synchronous and call it outside of the event loop using `run_async`, as was done before.

Or make all `SglangRouter` methods async and ensure there are no blocking calls (no `httpx.Client`, `subprocess.Popen`, etc). But that'd be too many unrelated changes, so I'd suggest to stick to the synchronous interface in this PR.

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -224,21 +233,65 @@ def _add_worker_to_router(self, worker_url: str) -> bool:
...
+                    worker_id = worker.get("id")
+                    if worker_id and isinstance(worker_id, str):
+                        break
+            if not worker_id:
+                logger.exception("No worker id found for url %s", worker_url)
+                return False
             with httpx.Client(timeout=5.0) as client:
                 response = client.delete(
-                    f"http://{self.context.host}:{self.context.port}/workers/{encoded_url}"
+                    f"http://{self.context.host}:{self.context.port}/workers/{worker_id}"
```
jvstme, line 297: What is this change for?

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -224,21 +233,65 @@ def _add_worker_to_router(self, worker_url: str) -> bool:
...
+            current_workers = self._get_router_workers()
+            worker_id = None
+            for worker in current_workers:
+                url = worker.get("url")
+                if url and isinstance(url, str) and url == worker_url:
+                    worker_id = worker.get("id")
+                    if worker_id and isinstance(worker_id, str):
+                        break
+            if not worker_id:
+                logger.exception("No worker id found for url %s", worker_url)
```
jvstme: (nit) `logger.exception` [should](https://docs.python.org/3/library/logging.html#logging.Logger.exception) only be called from an exception handler. You can use `logger.error`

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -144,33 +147,37 @@ async def register(self, conf: SiteConfig, acme: ACMESettings) -> None:
...
-                    ]
-
-                    # Write router workers config
-                    try:
+                    if conf.router.pd_disaggregation:
+                        # PD path: replica_urls from internal_ip (router talks directly to workers)
+                        replica_urls = [
+                            f"http://{replica.internal_ip}:{replica.port}"
+                            for replica in conf.replicas
+                            if replica.internal_ip
```
jvstme: This ignores replicas that don't have an internal IP (most backends don't set the internal IP). I assume such replicas will be shown as `running` and `registered` in dstack, but the service won't work, which will be difficult to troubleshoot.

Raise `ProxyError` if `pd_disaggregation` is enabled and some replicas don't have an internal IP?

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
```diff
@@ -224,21 +233,65 @@ def _add_worker_to_router(self, worker_url: str) -> bool:
...
         except Exception:
-            logger.exception("Error adding worker %s", worker_url)
+            logger.exception("Error adding worker %s", url)
+            return False
+
+    async def register_worker(self, url: str) -> bool:
+        server_info_url = f"{url}/server_info"
+        try:
+            async with httpx.AsyncClient(timeout=10) as client:
+                resp = await client.get(server_info_url)
```
jvstme: (nit) Avoid this request and assume `worker_type = "regular"` if `pd_disaggregation` is disabled?

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -189,12 +196,12 @@ async def unregister(self, domain: str) -> None:
...
-                if domain in self._domain_to_worker_ports:
-                    worker_ports = self._domain_to_worker_ports[domain]
-                    replica_urls = [
-                        f"http://{router.context.host}:{port}" for port in worker_ports
-                    ]
-                    await run_async(router.remove_replicas, replica_urls)
+                if domain in self._domain_to_worker_urls:
+                    worker_urls = self._domain_to_worker_urls[domain]
+                    await run_async(router.remove_replicas, worker_urls)
+                    self._discard_ports(worker_urls)
```
jvstme: No ports are allocated for services with `pd_disaggregation`, but `_discard_ports` is still called unconditionally. This can potentially deallocate ports that are actually allocated to some other unrelated service

src/dstack/_internal/proxy/gateway/services/registry.py
```diff
@@ -258,7 +260,12 @@ async def apply_service(
         service, repo, service_conn_pool
     )
```
jvstme: (nit) This opens an SSH tunnel unconditionally, but for services with `pd_disaggregation` the gateway doesn't actually need an SSH tunnel, because it communicates to replicas directly over internal IP

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
Bihan: Latest `sglang_router (0.3.2)` does not support `/{worker_url} `endpoint for deletion.

src/dstack/_internal/core/models/configurations.py
Bihan: Done

src/dstack/_internal/core/models/configurations.py
Bihan: Done

src/dstack/_internal/core/models/configurations.py
Bihan: Done

src/dstack/_internal/core/models/routers.py
Bihan: Done

Bihan: Done

src/dstack/_internal/proxy/gateway/repo/state_v1.py
Bihan: Done

src/dstack/_internal/proxy/lib/models.py
Bihan: Done

src/dstack/_internal/server/services/services/__init__.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/model_routers/base.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
Bihan: Thanks for pointing it out. Done

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/model_routers/sglang.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/registry.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/nginx.py
```diff
@@ -185,7 +189,7 @@ async def register(self, conf: SiteConfig, acme: ACMESettings) -> None:
 
         logger.info("Registered %s domain %s", conf.type, conf.domain)
 
-    async def unregister(self, domain: str) -> None:
+    async def unregister(self, domain: str, service: models.Service) -> None:
```
jvstme: (nit) If it now accepts `service`, it no longer has to accept `domain`. The domain is available as `service.domain_safe`

src/dstack/_internal/proxy/gateway/services/registry.py
```diff
@@ -237,6 +237,13 @@ async def register_model_entrypoint(
     logger.info("Entrypoint %s is now registered in project %s", domain, project_name)
 
 
+def _uses_pd_disaggregation(service: models.Service) -> bool:
+    """PD disaggregation: router talks to replicas via internal_ip, no SSH tunnels needed."""
+    return (
+        service.router is not None and getattr(service.router, "pd_disaggregation", False) is True
```
jvstme: (nit) Attribute access via `getattr` is not indexed by the IDE. I think it's almost never a good idea to use it, unless there's no other option.

You could do `service.router.pd_disaggregation` like in [other places](https://github.com/Bihan/dstack/blob/8560bcb0bc2e030e277b42c41686022130742e74/src/dstack/_internal/proxy/gateway/services/nginx.py#L207).

If you want to make it extra clear that the attribute exists, you could also do `isinstance(service.router, SGLangRouterConfig)`, although it seems to be redundant until we have other router types.

src/dstack/_internal/core/models/gateways.py
```diff
@@ -63,8 +63,8 @@ class GatewayConfiguration(CoreModel):
         ),
     ] = None
     router: Annotated[
-        Optional[AnyRouterConfig],
-        Field(description="The router configuration"),
+        Optional[RouterType],
```
jvstme: This change isn't backward compatible — if I have a previously created gateway with `router: type: sglang`, dstack fails to start new services:

```
[09:10:10] INFO     dstack._internal.server.services.events:205 Emitting event: Run submitted. Status: SUBMITTED. Event targets:                        
                    run(99e05f)pretty-insect-1. Actor: user(aeafc8)admin                                                                                
ERROR:    Exception in ASGI application
  + Exception Group Traceback (most recent call last):
[...]
  File "pydantic/main.py", line 364, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for GatewayConfigurationResponse
router
  value is not a valid enumeration member; permitted: 'sglang' (type=type_error.enum; enum_values=[<RouterType.SGLANG: 'sglang'>])
```

And to delete the gateway:

```shell
$ dstack gateway delete -y sglang-gateway
Unexpected error: status code 500 when requesting http://localhost:3000/api/project/main/gateways/get. Check the server logs for backend issues, and the
CLI logs at (~/.dstack/logs/cli/latest.log) local CLI output
```

Apart from backward compatibility, having `router: type: sglang` is useful in case we ever support specifying any gateway-level router properties, such as the SGLang package version.

```yaml
router:
  type: sglang
  version: 0.3.2
```

I can suggest to introduce two different pydantic models for service-level (`type`, `policy`, `pd_disaggregation`) and gateway-level (`type`, `policy`) router configuration. Before we can drop the gateway-level `router.policy` in a future version, we mention that it is deprecated in its `description`

src/dstack/_internal/server/services/services/__init__.py
```diff
@@ -91,8 +92,17 @@ async def _register_service_in_gateway(
...
+        and run_spec.configuration.router.type == RouterType.SGLANG
+    ):
+        if gateway_configuration.router != RouterType.SGLANG:
+            raise ServerClientError(
+                f"Service requires a SGLang gateway but gateway '{gateway.name}' "
+                "does not have the SGLang router configured."
+            )
     service_https = _get_service_https(run_spec, gateway_configuration)
-    router = gateway_configuration.router
+    router = run_spec.configuration.router
```
jvstme: Does that mean the SGLang router will not be enabled if the service config does not specify `router: type: sglang`? I'm not sure if this is expected or not.

- On one hand, I like these semantics more:
  - To use the SGLang router, you explicitly request it in the service configuration. This prevents misconfiguration cases when a non-SGLang service gets assigned to a gateway with SGLang enabled.
  - This allows to use the same gateway for both SGLang and non-SGLang services, without the need to create two separate gateways.
- On the other hand, this change of semantics is not backward-compatible — if someone is already using SGLang-enabled gateways, they expect SGLang to be enabled automatically.

Maybe the safest option for now is to keep the old semantics (automatically enable SGLang if specified in the gateway configuration). Later, we can discuss the change of semantics with the team (cc @peterschmidt85) and decide if it's worth it

Bihan: @jvstme 
cc @peterschmidt85 

Maybe we keep GatewayConfiguration's `router` to `Optional[AnyRouterConfig] `with properties `type` and `policy` without `pd_disaggregation`  as before. This will eliminate backward compatibility issue. 

Create new `GatewayConfig` class  as 
```
class GatewayConfig(CoreModel):
    name: Annotated[
        Optional[str],
        Field(description="The name of the gateway. Omit for the default gateway."),
    ] = None
    pd_disaggregation: Annotated[
        bool,
        Field(
            description=(
                "Enable PD disaggregation mode. Requires a gateway with SGLang router. "
                "Uses internal IP instead of SSH tunnels for replica connections."
            ),
        ),
    ] = False
```

Update `ServiceConfigurationParams`'s gateway field to

  ```
  gateway: Annotated[
      Optional[Union[bool, str, GatewayConfig]],
      Field(
          description=(
              "Gateway configuration. Use `false` to run without a gateway. "
              "Use `true` to run with the default gateway. "
              "Use a string for the gateway name. "
              "Use an object for name + options (e.g. pd_disaggregation). "
              "Omit to use the default gateway if one exists, otherwise no gateway."
          ),
      ),
      ] = None
```
YAML configs with gateway..

```
# Boolean shortcuts (unchanged)
gateway: true   # default gateway
gateway: false  # no gateway

# String shortcut
gateway: "my-gateway"

# Object form
gateway:
  name: "my-gateway"
  pd_disaggregation: true

# Default gateway + pd_disaggregation
gateway:
  pd_disaggregation: true

```
Validation in `_register_service_in_server` would be

```
if isinstance(run_spec.configuration.gateway, GatewayConfig):
    raise ServerClientError(
        "Service is configured with gateway options but no gateway is available. "
        "Please configure a gateway or set gateway to false to run without a gateway."
    )
```

Validation in `_register_service_in_gateway
`
```
gateway_configuration = get_gateway_configuration(gateway)
service_gateway = run_spec.configuration.gateway
if (
    isinstance(service_gateway, GatewayConfig)
    and service_gateway.pd_disaggregation
):
    if gateway_configuration.router != RouterType.SGLANG:
        raise ServerClientError(
            f"Service requires a SGLang gateway (pd_disaggregation) but gateway '{gateway.name}' "
            "does not have the SGLang router configured."
        )
```

Finally router construction in _register_service_in_gateway becomes
```
router = None
if gateway_configuration.router == RouterType.SGLANG:
    service_gateway = run_spec.configuration.gateway
    pd_disaggregation = (
        isinstance(service_gateway, GatewayConfig) and service_gateway.pd_disaggregation
    )
    policy = gateway_configuration.router_policy  # or gateway_configuration.router_config.policy
    router = ServiceRouterConfig(
        type=RouterType.SGLANG.value,
        policy=policy,
        pd_disaggregation=pd_disaggregation,
    )
```

File: src/dstack/_internal/server/services/services/__init__.py [hunk 18]
```diff
@@ -92,17 +127,22 @@ async def _register_service_in_gateway(
...
+    service_pd_disaggregation = (
+        service_router is not None
+        and isinstance(service_router, SGLangRouterConfig)
+        and service_router.pd_disaggregation
+    )
+    if service_pd_disaggregation and not _gateway_has_sglang_router(gateway_configuration):
+        raise ServerClientError(
+            "Service requires gateway with SGLang router for pd_disaggregation but gateway "
+            f"'{gateway.name}' does not have the SGLang router configured."
+        )
```
jvstme: From the `ServiceConfiguration.router` [description](https://github.com/dstackai/dstack/blob/56286c47fb60e2ab2c5b10792d8f66b36a2c7168/src/dstack/_internal/core/models/configurations.py#L896):

> Router configuration for the service. Requires a gateway with matching router enabled

So this error here should be triggered for any service with `router: type: sglang`, not only for those with `pd_disaggregation: true`. For example, this configuration should also fail if assigned to a non-SGLang gateway:

```yaml
type: service
port: 80
image: my-image
router:
  type: sglang
```

src/dstack/_internal/core/models/routers.py
```diff
@@ -25,4 +43,4 @@ class SGLangRouterConfig(CoreModel):
     ] = False
 
 
-AnyRouterConfig = SGLangRouterConfig
+AnyRouterConfig = Union[SGLangRouterConfig, GatewayRouterConfig]
```
jvstme: Why unite these two? One of these is the gateway-level router config, and the other is the service-level router config. They are supposed to have different properties (except `policy`, for compatibility reasons), and I can't think of any context where we'd need `union[gateway-level, service-level]`.

Uniting them will likely result in unexpected behavior. For example, if you add `GatewayRouterConfig.version`, it will also appear in the service-level router config ([annotated](https://github.com/dstackai/dstack/blob/56286c47fb60e2ab2c5b10792d8f66b36a2c7168/src/dstack/_internal/core/models/configurations.py#L892C5-L892C12) as `AnyRouterConfig`), which is not expected.

I think the class structure we need is:

```python
class SGLangServiceRouterConfig:
    type: ...
    policy: ...
    pd_disaggregation: ...
class SGLangGatewayRouterConfig:
    type: ...
    policy: ...  # for compatibility
    # In the future:
    # version: ...
AnyServiceRouterConfig = SGLangServiceRouterConfig  # In the future: Union[SGLangServiceRouterConfig, VLLMServiceRouterConfig, ...]
AnyGatewayRouterConfig = SGLangGatewayRouterConfig  # In the future: Union[SGLangGatewayRouterConfig, VLLMGatewayRouterConfig, ...]
```

src/dstack/_internal/core/models/configurations.py
```diff
@@ -888,6 +889,14 @@ class ServiceConfigurationParams(CoreModel):
             )
         ),
     ] = None
+    router: Annotated[
+        Optional[AnyRouterConfig],
+        Field(
+            description=(
+                "Router configuration for the service. Requires a gateway with matching router enabled. "
```
jvstme, line 895: (nit)


```suggestion
                "Router configuration for the service. Requires a gateway with matching router enabled"
```

src/dstack/_internal/core/models/routers.py
Bihan: Actually, I am still in WIP. I pushed because gateway needs to download from this branch. This is exactly what I am doing.

Bihan: Done

src/dstack/_internal/server/services/services/__init__.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/nginx.py
Bihan: Done

src/dstack/_internal/proxy/gateway/services/registry.py
Bihan: Done

jvstme (approved):

PR #3560, opened 2026-02-10
Migrate service model base url

peterschmidt85:
## Summary

Migrate `service.model.base_url` from the deprecated `gateway.{domain}` model aggregation endpoint to the service endpoint for `openai`-format models.

- **`service.model.base_url`**: For `openai` format, set to `service_url + prefix` (e.g. `https://run-name.domain.com/v1`). For other formats, keep the existing `gateway.{domain}` URL.
- **Default probe URL**: Derive from model's `prefix` field instead of hardcoding `/v1`.
- **CORS on gateway service endpoints**: Add CORS support (preflight handling + response headers) to the `service.jinja2` nginx template, enabled only for services with `openai`-format models. Strips upstream CORS headers to avoid duplicates.
- **`cors_enabled` on `Service` model**: Store CORS flag per-service for correctness with multiple services. Includes a startup migration for gateways with pre-existing state.


Bihan (approved):
These changes do not affect SGLang router–based services, before or after [PR](https://github.com/dstackai/dstack/pull/3558)

peterschmidt85:
@jvstme Let me know once you have a look. In principle, we can also merge as is.

src/dstack/_internal/proxy/lib/models.py
```diff
@@ -59,6 +59,7 @@ class Service(ImmutableModel):
     strip_prefix: bool = True  # only used in-server
     replicas: tuple[Replica, ...]
     router: Optional[AnyRouterConfig] = None
+    cors_enabled: bool = False  # only used on gateways; enabled for openai-format models
```
jvstme, line 62: (nit) Whether CORS is enabled is a decision the gateway takes based on existing data points, not a new data point, so I'd prefer not to persist it in the repo.

But if there isn't a different and easy way to implement that decision, I don't mind keeping it as is

src/dstack/_internal/proxy/gateway/services/registry.py
```diff
@@ -305,7 +309,7 @@ async def stop_replica_connections(
 
 
 async def get_nginx_service_config(
-    service: models.Service, replicas: Iterable[ReplicaConfig]
+    service: models.Service, replicas: Iterable[ReplicaConfig], cors_enabled: bool = False
```
jvstme: (nit) No need to accept `cors_enabled` as a parameter, it is already available in `service.cors_enabled`

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
jvstme, line 1: This CORS policy is more permissive than what we have on the `gateway.*` endpoint, because this policy allows cross-origin requests to the entire service, not just the model. Consider setting a stricter policy by returning the CORS headers only from `/v1` (or a different configured OpenAI API prefix). Should be relatively easy to implement with an `if ($request_uri ~ ^/<prefix>)` block

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
```diff
@@ -60,6 +80,15 @@ server {
     location @ {
         set $dstack_replica_hit 1;
         {% if replicas %}
+        {% if cors_enabled %}
+        proxy_hide_header 'Access-Control-Allow-Origin';
+        proxy_hide_header 'Access-Control-Allow-Methods';
+        proxy_hide_header 'Access-Control-Allow-Headers';
+        proxy_hide_header 'Access-Control-Allow-Credentials';
```
jvstme, line 87: Note that this prevents users from configuring their own CORS policy for the service. For example, users may want to set a stricter `Access-Control-Allow-Origin`, which is a requirement if they also set `Access-Control-Allow-Credentials` (which is [forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS/Errors/CORSNotSupportingCredentials) when `Access-Control-Allow-Origin` is `*`).

But I'm not sure if there's an easy way to set headers conditionally, based on whether a header was returned by the upstream.

Maybe this limitation won't have any realistic impact if we only set our CORS policy on `/v1` (or another configured OpenAI API prefix)

jvstme (approved):

src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
peterschmidt85: Yes, if you don't strongly insist, I'd prefer to implement it separately from the main PR.

PR #3575, opened 2026-02-16
[runner] Don't bind to public addresses

### LINKED ISSUE #3078: Do not bind `dstack-runner` to the public IP address
### Problem

Currently, `dstack-runner` always [listens](https://github.com/dstackai/dstack/blob/b6aa7e89781d59ac2caebf3590afbf42a3a43d4a/runner/cmd/runner/main.go#L41) on all network interfaces, including the instance's public IP address if the container runs in [host network mode](https://github.com/dstackai/dstack/blob/b6aa7e89781d59ac2caebf3590afbf42a3a43d4a/src/dstack/_internal/core/models/common.py#L118).

In general, this is not an issue, as access to the public IP should be restricted by the firewall. For cloud instances, dstack automatically [configures](https://github.com/dstackai/dstack/blob/b6aa7e89781d59ac2caebf3590afbf42a3a43d4a/src/dstack/_internal/core/backends/base/compute.py#L671) the firewall. For SSH fleets, firewall configuration is the responsibility of the fleet administrator.

However, to reduce the risk of unintended exposure in case of a firewall misconfiguration, it is recommended to avoid binding dstack-runner to the public IP address altogether.

### Solution

- If the container is running in host network mode, bind only to localhost.
- If the container is running in bridge network mode, continue binding to all interfaces — otherwise, the container port cannot be forwarded to the host's localhost.

### Workaround

_No response_

Yes

un-def:
Closes: https://github.com/dstackai/dstack/issues/3078

### DIFF
--- a/runner/cmd/runner/main.go
+++ b/runner/cmd/runner/main.go
@@ -31,6 +31,7 @@ func main() {
 
 func mainInner() int {
 	var tempDir string
+	var httpAddress string
 	var httpPort int
 	var sshPort int
 	var sshAuthorizedKeys []string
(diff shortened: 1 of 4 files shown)


jvstme (approved):
👍

PR #3613, opened 2026-02-25
[GCP] Make `tpu` an opt-in option in the backend configuration

peterschmidt85:
Currently, TPU offers are included by default for all GCP backends, even though most users don't use TPUs. This creates noise in offer listings and can lead to accidental TPU provisioning. Additionally, TPU support is still maturing,

This PR adds a `tpu` option to the GCP backend config (default `false`). When set to `true`, single-host TPU offers are included in offer listings and available for provisioning. This mirrors the pattern already established by RunPod's `community_cloud` option.


r4victor:
@peterschmidt85, `dstack apply` already shows only nvidia offers by default, so users already need to explicitly specify TPU for it to be provisioned – they won't be provisioning TPUs accidentally.

Issue #3473, opened 2026-01-19
Stop calling `api/projects/list` on every UI page


r4victor:
Currently, UI call `api/projects/list` when loading practically any UI page (e.g. Runs, Fleets, Instances, etc) even when it doesn't seem to be necessary. This endpoint can be heavy for global admins (especially on Sky) since it returns all projects on the server.

The main reason call `api/projects/list` seems to be `useProjectFilter` that allows filtering resources by project name. For this, we could introduce api endpoint to list project names only (perhaps by prefix, with a limit).

Also:
* [`useCheckingForFleetsInProjects` should not `useGetProjectsQuery`](https://github.com/dstackai/dstack/blob/a07ef352779fad5aa2fdc0334314becc940ddd47/frontend/src/hooks/useCheckingForFleetsInProjectsOfMember.ts#L8) since there is already `useGetOnlyNoFleetsProjectsQuery`.
* [TutorialPanel should not list all projects](https://github.com/dstackai/dstack/blob/a07ef352779fad5aa2fdc0334314becc940ddd47/frontend/src/layouts/AppLayout/TutorialPanel/hooks.ts#L47).
* [Select only project names for events project filtering](https://github.com/dstackai/dstack/blob/a07ef352779fad5aa2fdc0334314becc940ddd47/frontend/src/pages/Events/List/hooks/useFilters.ts#L68).

Overall, ensure the UI never lists all projects (besides Projects page currently). Introduce more efficient API if needed.

**Ensure the same for `api/users/list`**.

PR #3621, opened 2026-02-26
[UI]: Implement dynamic property filter options

olgenn:
#3473

### DIFF
--- a/frontend/src/pages/Runs/List/hooks/useFilters.ts
+++ b/frontend/src/pages/Runs/List/hooks/useFilters.ts
@@ -16,6 +16,8 @@ const filterKeys: Record<string, RequestParamsKeys> = {
     USER_NAME: 'username',
 };
 
+const limit = 100;
+
 export const useFilters = () => {
     const [searchParams, setSearchParams] = useSearchParams();
     const [onlyActive, setOnlyActive] = useLocalStorageState('run-list-filter-only-active', true);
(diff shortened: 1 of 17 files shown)


frontend/src/pages/Runs/List/hooks/useFilters.ts
```diff
@@ -102,6 +81,46 @@ export const useFilters = ({ localStorePrefix }: Args) => {
...
+        setFilteringOptions([]);
+
+        if (!filteringText.length) {
+            return Promise.resolve();
+        }
+
+        setFilteringStatusType('loading');
+
+        if (filteringProperty?.key === filterKeys.PROJECT_NAME) {
+            await getProjects({ name_pattern: filteringText })
```
r4victor: The UI should set some sensible limit in filters, e.g. 100 – the default server limit is large (2000) for backward compatibility.

r4victor (approved):

peterschmidt85:
@olgenn If Victor tested it, we can merge

jvstme (commented):
@olgenn, I tested event filters, found these issues:
- The `Target projects`, `Within projects`, `Target users`, and `Actors` filters on the `Events` page switch from names to IDs in a weird way (as you've mentioned yourself). This is a regression, so it'd be good to fix it. But if there isn't a simple fix, I personally don't mind keeping it as is.

   Image

- The `Events` tabs on Project and User pages don't work. They attempt to set `within_projects`, `actors`, and `target_users` to names instead of IDs, which fails with error 422.

olgenn:
@jvstme I fixed second point

jvstme (approved):

Issue #3644, opened 2026-03-10
[Feature]: SSH reverse proxy


un-def:
### Problem

* `dstack` CLI app is required for `ssh` and IDE support. Currently, the user can submit the run via web UI, but cannot connect to it
* SSH fleet instances behind a bastion host are not accessible to the user unless they have a private key to the bastion host, see https://github.com/dstackai/dstack/issues/2472 
* (hypothetical) private instances (instances without public IPs) are not accessible to the user unless they are connected to the same private network

### Solution

https://docs.google.com/document/d/1AMDnZ0TSqdZI5SApW3faPMyI_WZPzPINCQHoXEMspDs

### Would you like to help us implement this feature by sending a PR?

Yes

r4victor:
@un-def, should we consider this completed? If there are any UI TODO, I'd create a separate issue.

un-def:
https://github.com/dstackai/dstack/issues/3768

PR #3646, opened 2026-03-10
Add API for SSH proxy

un-def:
Part-of: https://github.com/dstackai/dstack/issues/3644


src/dstack/_internal/server/services/jobs/__init__.py
```diff
@@ -267,6 +267,10 @@ def get_job_runtime_data(job_model: JobModel) -> Optional[JobRuntimeData]:
     return JobRuntimeData.__response__.parse_raw(job_model.job_runtime_data)
 
 
+def get_job_spec(job_model: JobModel) -> JobSpec:
+    return JobSpec.__response__.parse_raw(job_model.job_spec_data)
```
r4victor, line 271: Worth replacing all `RunSpec.__response__.parse_raw` calls with the helper.

r4victor (approved):

Issue #3626, opened 2026-03-02
[Feature]: Cross-project SSH fleet sharing


jvstme:
### Problem

It is not possible to share SSH instances across multiple `dstack` projects.

### Solution

https://docs.google.com/document/d/1kU4eQDXZz9ovQ-1V6etrV6wSuIhm-wC9cwqz6hUgeCQ

### Implementation plan

- [x] Submitting jobs to imported fleets
- [x] Viewing imported fleets in API, CLI, UI
- [x] Filtering events by imported fleets and instances
- [x] Exports API & CLI
- [x] Imports API & CLI
- [x] Evicting jobs if fleet is no longer imported
- [x] Referencing imported fleets in CLI and run configurations via `<project>/<fleet>`
- [x] Removing fleets from exports on deletion
- [x] (?) Removing exports / imports on project deletion
- [ ] (?) Export / import events
- [x] Docs

PR #3659, opened 2026-03-12
Add Exports concept page and CLI reference

peterschmidt85:
## Summary
- Add a dedicated Exports concept page under Concepts > More
- Add `dstack export` CLI reference page with all subcommands (list, create, update, delete)
- Add cross-links from Fleets and Projects concept pages
- Add Exports to nav in mkdocs.yml
- Add Tenant isolation section to Fleets docs describing the current host access model, with a cross-reference in Exports

## Test plan
- [ ] Verify the new pages render correctly with `mkdocs serve`
- [ ] Check all cross-links between exports, fleets, and projects pages
- [ ] Verify `#GENERATE#` placeholders produce correct CLI help output

#3626

### DIFF
--- a/docs/docs/concepts/exports.md
+++ b/docs/docs/concepts/exports.md
@@ -11,8 +11,6 @@ the specified importer projects can see and use it as if it were their own.
 !!! warning "Experimental"
     Exports are an experimental feature.
     Currently, only [SSH fleets](fleets.md#ssh-fleets) can be exported.
-    Exports do not yet support configuration files and `dstack apply`.
-    Use the CLI commands below to manage exports.
 
 An export is created in the exporter project and specifies the resources to export and the
 importer projects that will gain access to them.
--- a/docs/docs/concepts/exports.md
+++ b/docs/docs/concepts/exports.md
@@ -134,6 +134,9 @@ $ dstack fleet list
 
 Imported fleets can be used for runs just like the project's own fleets.
 
+!!! info "Tenant isolation"
+    Exported fleets share the same access model as regular fleets. See [Tenant isolation](fleets.md#tenant-isolation) for details.
+
 !!! info "What's next?"
     1. Check the [`dstack export` CLI reference](../reference/cli/dstack/export.md)
     2. Learn how to manage [fleets](fleets.md)
--- a/docs/docs/concepts/exports.md
+++ b/docs/docs/concepts/exports.md
@@ -23,7 +23,7 @@ for running tasks, dev environments, and services. Imported resources appear wit
 
 !!! info "Required project role"
     The user creating or updating an export must have the project admin role on both the exporter project and
-    each importer project. Alternatively, a global admin can add any project as an importer.
+    any importer project they add. Alternatively, a global admin can add any project as an importer.
 
 ## Manage exports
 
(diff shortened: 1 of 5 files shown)


docs/docs/concepts/exports.md
```diff
@@ -0,0 +1,142 @@
...
+# Exports
+
+Exports allow making resources from one project available to other projects. When a project exports a resource,
+the specified importer projects can see and use it as if it were their own.
+
+!!! warning "Experimental"
+    Exports are an experimental feature.
+    Currently, only [SSH fleets](fleets.md#ssh-fleets) can be exported.
+    Exports do not yet support configuration files and `dstack apply`.
+    Use the CLI commands below to manage exports.
```
jvstme: We [intentionally chose](https://docs.google.com/document/d/1kU4eQDXZz9ovQ-1V6etrV6wSuIhm-wC9cwqz6hUgeCQ/edit?disco=AAABvYKbMEI) CLI commands over configuration files. We can revisit this decision in the future, but currently exports are not expected to support configuration files, so I wouldn't say that they do not "yet" support them.


```suggestion
```

docs/docs/concepts/exports.md
```diff
@@ -0,0 +1,142 @@
...
+An export is created in the exporter project and specifies the resources to export and the
+importer projects that will gain access to them.
+
+Once an export is created, the importer projects can see the exported resources in their resource lists and use them
+for running tasks, dev environments, and services. Imported resources appear with a project prefix
+(e.g., `team-a/my-fleet`) to distinguish them from the project's own resources.
+
+!!! info "Required project role"
+    The user creating or updating an export must have the project admin role on both the exporter project and
+    each importer project. Alternatively, a global admin can add any project as an importer.
```
jvstme: > each importer project

Technically, only on those importer projects that they are adding. For example, if the export already has importer A and you adding importer B, you only need to be the admin in importer B.


```suggestion
    The user creating or updating an export must have the project admin role on both the exporter project and
    any importer project they add. Alternatively, a global admin can add any project as an importer.
```

docs/docs/concepts/exports.md
jvstme, line 1: I would also add a note about the lack of instance-level isolation between users, since this may be important for some customers. Something along the lines of:

> Every importer gains full SSH access to the fleet's instances. Only share fleets with trusted teams.

jvstme (approved):
👍

docs/docs/concepts/fleets.md
```diff
@@ -468,6 +468,20 @@ ssh_config:
 !!! info "Reference"
     The fleet configuration file supports additional options, including [`instance_types`](../reference/dstack.yml/fleet.md#instance_types), [`max_price`](../reference/dstack.yml/fleet.md#max_price), [`regions`](../reference/dstack.yml/fleet.md#max_price), among others. For the complete list, see the [reference](../reference/dstack.yml/fleet.md).
 
+## Tenant isolation
```
peterschmidt85, line 471: @un-def please review if possible

un-def: A small correction: 

> For distributed workloads, containers use host network mode

The current default mode is `HOST_WHEN_POSSIBLE`:

https://github.com/dstackai/dstack/blob/733a17c7746dda39112f85cec1a7bf601778c656/src/dstack/_internal/server/settings.py#L151-L152

That is, the host mode is used for any type of workload, not only distributed, and the bridge mode is only used if the blocks feature is enabled for the instance and the job doesn't occupy all blocks.

PR #3668, opened 2026-03-16
Add per-job hourly log quota enforced on runner

peterschmidt85:
## Summary

- Adds a per-job hourly log quota (default 50MB) enforced on the runner side, preventing runaway log costs (e.g., the $194+/day CloudWatch incident from excessive training job logs)
- Quota is configurable via `DSTACK_SERVER_LOG_QUOTA_PER_JOB_HOUR` env var (bytes, 0 disables)
- Jobs exceeding the quota are killed with status `error` and reason `log quota exceeded`
- Byte counting happens post-ANSI-stripping, matching what gets stored in CloudWatch

### Go (runner)
- `types.go`: Add `TerminationReasonLogQuotaExceeded` constant
- `schemas.go`: Add `LogQuotaHour` field to `JobSpec`
- `logs.go`: Add quota tracking to `appendWriter` with out-of-band signaling via channel (needed because ansistrip is async and swallows downstream errors)
- `executor.go`: Add `copyOutputWithQuota()` method, wire quota via `SetJob()`, handle quota error in `Run()`
- `executor_test.go`: Add `TestExecutor_LogQuota`

### Python (server)
- `runs.py`: Add `LOG_QUOTA_EXCEEDED` to `JobTerminationReason` (maps to `FAILED`), add `log_quota_hour` to `JobSpec`
- `settings.py`: Add `SERVER_LOG_QUOTA_PER_JOB_HOUR` setting (default 50MB)
- `runner.py`: Add `log_quota_hour` to `SubmitBody` include set
- `base.py`: Add `_log_quota_hour()` method, wire into `_get_job_spec()`

## Test plan

- [x] `TestExecutor_LogQuota` passes
- [x] All Go tests pass (`go test ./...`)
- [x] All Python tests pass (2222 passed)
- [x] E2E with remote backend: 1000-byte quota — job terminates immediately with `log quota exceeded`
- [x] E2E with remote backend: 50MB quota — job runs ~5 min, terminates at ~52MB with `log quota exceeded`
- [x] `dstack ps -v` shows status `error` and error `log quota exceeded`
- [x] `dstack logs` shows partial logs captured before termination


src/dstack/_internal/core/models/runs.py
```diff
@@ -275,6 +278,9 @@ class JobSpec(CoreModel):
     privileged: bool = False
     single_branch: Optional[bool] = None
     max_duration: Optional[int]
+    log_quota_hour: Optional[int] = None
```
r4victor: JobSpec is sent from client to server (as a part of RunPlan.current_resource), so new fields must be excluded for client backward compatibility with older servers. Or maybe not include it in JobSpec at all, e.g. send in SubmitBody?

runner/internal/runner/executor/logs.go
```diff
@@ -1,29 +1,63 @@
...
+func (w *appendWriter) QuotaExceeded() <-chan struct{} {
+	return w.quotaExceeded
+}
+
 func (w *appendWriter) Write(p []byte) (n int, err error) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 
+	if w.quota > 0 {
+		now := time.Now().Unix()
```
un-def: This implementation is susceptible to wall clock changes. This is nos critical though, as we count traffic per hour and the event where there are several clock adjustments changing `hourStart` (= resetting `bytesInHour`) during one hour is highly unlikely. Nonetheless, I'd rather store monotonic start time in the stuct (e.g., `timeStarted time.Time`) and calculate `currentHour` as `math.Floor((time.Now() - w.timeStarted).Hours())`

src/dstack/_internal/core/models/runs.py
peterschmidt85: Resolved — moved `log_quota_hour` from `JobSpec` to `SubmitBody` (server→runner only). It's no longer part of `RunPlan.current_resource`, so no client backward compatibility issue.

runner/internal/runner/executor/logs.go
peterschmidt85: Resolved — switched from wall clock (`time.Now().Unix()`) to monotonic time (`time.Since(w.timeStarted).Hours()`), which is immune to NTP adjustments.

un-def (approved):

PR #3685, opened 2026-03-23
Add per-instance data transfer metering

peterschmidt85:
## Summary

Adds per-instance outbound data transfer metering to track billable network traffic.

- The shim starts an iptables-based meter (`dstack-nm` chain) at boot that counts outbound bytes to external IPs, excluding private/VPC traffic (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, 169.254.0.0/16)
- Cumulative bytes are reported via `GET /api/instance/health` and stored on `InstanceModel.data_transfer_bytes`
- Server reads bytes during periodic health checks (~60s) and captures a final reading before instance termination
- Exposed via the Instance API for downstream billing integration
- Metering is best-effort: if iptables is unavailable, the shim logs a warning and continues without metering

### Files changed

**Shim:** `netmeter/` package (iptables chain setup, 10s polling, atomic `Bytes()` read), started at shim boot in `main.go`, reported via `InstanceHealthResponse.data_transfer_bytes`

**Server:** `InstanceModel.data_transfer_bytes` column + Alembic migration, extraction in instance health check (both pipeline_tasks and scheduled_tasks), final read in termination path, `Instance` API model

## Test plan

- [x] Unit tests: iptables output parsing, `Bytes()` read
- [x] golangci-lint, ruff: 0 issues
- [x] Go + Python tests: all pass
- [x] E2E on AWS: task uploaded ~20MB, `data_transfer_bytes` = 22.1 MB (includes apt-get, Docker pull overhead)
- [x] Final read at termination: value captured before instance destroyed
- [x] Backward compatible: old shims → field is None/0, old servers → ignore extra field

### DIFF
--- a/src/dstack/_internal/server/schemas/instances.py
+++ b/src/dstack/_internal/server/schemas/instances.py
@@ -37,7 +37,10 @@ class InstanceCheck(CoreModel):
     def has_health_checks(self) -> bool:
         if self.health_response is None:
             return False
-        return self.health_response.dcgm is not None
+        return (
+            self.health_response.dcgm is not None
+            or self.health_response.data_transfer_bytes is not None
+        )
 
 
 class GetInstanceHealthChecksRequest(CoreModel):
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ uv.lock
 
 profiling_results.html
 docs/docs/reference/api/rest/openapi.json
+.agents/skills
+.claude/skills
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,3 @@ uv.lock
 
 profiling_results.html
 docs/docs/reference/api/rest/openapi.json
-.agents/skills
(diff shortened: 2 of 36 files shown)


src/dstack/_internal/server/schemas/instances.py
```diff
@@ -37,7 +37,10 @@ def get_health_status(self) -> HealthStatus:
     def has_health_checks(self) -> bool:
         if self.health_response is None:
             return False
-        return self.health_response.dcgm is not None
+        return (
+            self.health_response.dcgm is not None
+            or self.health_response.data_transfer_bytes is not None
```
r4victor, line 42: I don't think having data_transfer_bytes should mean there are health check present – we wouldn't want to create health checks only because of data_transfer_bytes. Overall, having data_transfer_bytes as a part of health checks is confusing.

Issue #3683, opened 2026-03-20
[Bug]: Multiple replica jobs assigned to same instance due to busy_blocks race


Bihan:
### Steps to reproduce

1. Create an SSH fleet with multiple hosts (e.g. 4 CPU hosts):

```
type: fleet
name: simple-cpu-fleet

ssh_config:
  user: bihan
  identity_file: ~/.ssh/id_rsa
  hosts:
    - 89.169.102.8
    - 89.169.120.141
    - 89.169.121.216
    - 89.169.103.68
```
2. Create a service with replicas: 4 :
```
type: service
name: simple-http
python: 3.12

commands:
  - python3 -m http.server 8000

replicas: 4
port: 8000
resources:
  cpu: 4
```
3. After the service is running, inspect containers on each host: sudo docker ps

### Actual behaviour

Several replica jobs are placed on the same instance. As below

```
sudo docker ps
CONTAINER ID   IMAGE                                 COMMAND                  CREATED          STATUS          PORTS     NAMES
db3de0c31670   dstackai/base:0.12-base-ubuntu22.04   "/bin/sh -c '( : && …"   52 seconds ago   Up 49 seconds             simple-http-0-2-5594b688
8b1b74135c49   dstackai/base:0.12-base-ubuntu22.04   "/bin/sh -c '( : && …"   52 seconds ago   Up 49 seconds             simple-http-0-1-500c5c8d
37dbddee69cf   dstackai/base:0.12-base-ubuntu22.04   "/bin/sh -c '( : && …"   52 seconds ago   Up 49 seconds             simple-http-0-3-b09943b7
```

### Expected behaviour

Each instance runs one job.

### dstack version

commit ff712e0f

### Server logs

```shell

```

### Additional information

This also occurs with cloud fleet instances after they become idle. When idle cloud instances are reused for new jobs, the same race condition causes multiple jobs to be assigned to the same instance instead of being spread across available instances.

eskayML:
this looks like a classic case of a race condition in the replica assignment logic. i'm guessing it's due to the fact that the `busy_blocks` list is not properly synchronized across instances, allowing multiple instances to claim the same block at the same time. a possible fix could be to introduce a locking mechanism to ensure that only one instance can claim a block at a time. should i open a PR for this?

Bihan:
[...]

PR #3690, opened 2026-03-25
Fix Issue 3683 multiple jobs assigned to same instance

Bihan:
This PR fixes bug [#3683](https://github.com/dstackai/dstack/issues/3683).

**Problem**
When assigning a job to an existing instance, the server used a read → add in Python → write on `busy_blocks` (e.g. load the row, compute `busy_blocks += offer.blocks`). Under concurrency, two jobs read the same value, both add their share, and both write—so capacity was oversubscribed and multiple jobs ended up on one instance.

**Fix**
Changed how capacity is reserved: instead of a non-atomic Python update, the server issues a single SQL UPDATE that:
Increments `busy_blocks` by the job’s blocks, only if there is remaining capacity (`total_blocks is NULL or busy_blocks + blocks <= total_blocks`).

### DIFF
--- a/src/dstack/_internal/server/background/pipeline_tasks/jobs_submitted.py
+++ b/src/dstack/_internal/server/background/pipeline_tasks/jobs_submitted.py
@@ -98,6 +98,7 @@ from dstack._internal.server.services.instances import (
     get_instance_offer,
     get_instance_provisioning_data,
     switch_instance_status,
+    try_atomic_busy_blocks_increment,
 )
 from dstack._internal.server.services.jobs import (
     check_can_attach_job_volumes,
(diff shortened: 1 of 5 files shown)


r4victor:
@Bihan Have you tried reproducing this problem with pipelines? I suspect it should already be fixed in pipelines and we don't need fixing the scheduled task since it'll be dropped right after the coming release.

r4victor:
And there is nothing wrong with doing busy_blocks += offer.blocks in Python if the resource is locked and the instances should be locked in this case. I think the original bug is related to lock being held correctly but busy_blocks is stale (read from unlock value).

Let me know if the problem is reproducible with pipelines and we'll be looking for the root cause in the case. If it's not, then there is no need for the fix.

Bihan:
> @Bihan Have you tried reproducing this problem with pipelines? I suspect it should already be fixed in pipelines and we don't need fixing the scheduled task since it'll be dropped right after the coming release.

@r4victor Do you mean with this merged [#3686](https://github.com/dstackai/dstack/pull/3686) there will be no issues?

r4victor:
@Bihan, #3686 shouldn't be related. I suspect JobSubmitted pipeline fixed it (that you try to change) (https://github.com/Bihan/dstack/blob/master/src/dstack/_internal/server/background/pipeline_tasks/jobs_submitted.py)

You can try set DSTACK_FF_PIPELINE_PROCESSING_ENABLED to enable pipelines and try to reproduce. scheduled tasks will be dropped after the coming release.

Bihan:
> @Bihan, #3686 shouldn't be related. I suspect JobSubmitted pipeline fixed it (that you try to change) (https://github.com/Bihan/dstack/blob/master/src/dstack/_internal/server/background/pipeline_tasks/jobs_submitted.py)
> 
> You can try set DSTACK_FF_PIPELINE_PROCESSING_ENABLED to enable pipelines and try to reproduce. scheduled tasks will be dropped after the coming release.

@r4victor With pipelines there is no above issue. Jobs are allocated to unique instances as expected. I think for now we can close this PR.

Issue #3278, opened 2025-11-12
[Bug]: Blocks param for SSH fleet does not work at top yaml level


pawelkopec:
### Steps to reproduce

I was trying to create SSH fleet with Tenstorrent chips with `blocks: auto` so that we can deploy workloads which use less than all chips on the node.

When I create the fleet with `dstack apply -f tt-fleet.yaml` with `blocks: auto` defined at the top yaml level
```
type: fleet
name: tt-fleet
blocks: auto

ssh_config:
  user: kopec
  identity_file: ~/.ssh/id_ed25519
  hosts:
    - <IP>
```

then dev env creation fails with `No matching instance offers available` as if the configured node was not available.

However when I move `blocks: auto` to the node level like so
```
type: fleet
name: tt-fleet

ssh_config:
  user: kopec
  identity_file: ~/.ssh/id_ed25519
  hosts:
    - hostname: <IP>
      blocks: auto
```
it works properly

### Actual behaviour

This is what I got when `blocks: auto` was set at the top yaml level
```
$ dstack fleet
 FLEET INSTANCE BACKEND RESOURCES PRICE STATUS CREATED 
 tt-fleet 0 ssh (remote) cpu=32 mem=504GB disk=2804GB n300:24GB:4 $0 idle 3 mins ago

$ dstack apply -f pawelk-dev-tt.yaml 
The working_dir is not set — using legacy default "/workflow". Future versions will default to the image's working directory.

 Project main 
 User admin 
 Configuration pawelk-dev-tt.yaml 
 Type dev-environment 
 Resources cpu=2.. mem=8GB.. disk=100GB.. n300:1 
 Spot policy on-demand 
 Max price - 
 Retry policy - 
 Creation policy reuse-or-create 
 Idle duration 5m 
 Max duration - 
 Inactivity duration - 
 Reservation - 

No matching instance offers available. Possible reasons: https://dstack.ai/docs/guides/troubleshooting/#no-offers
```

### Expected behaviour

This is the desired output I got once I moved `blocks: auto` to the node level 
```
$ dstack apply -f pawelk-dev-tt.yaml 
The working_dir is not set — using legacy default "/workflow". Future versions will default to the image's working directory.

 Project              main                                  
 User                 admin                                 
 Configuration        pawelk-dev-tt.yaml                    
 Type                 dev-environment                       
 Resources            cpu=2.. mem=8GB.. disk=100GB.. n300:1 
 Spot policy          on-demand                             
[...]

PR #3695, opened 2026-03-25
fix: use fleet blocks config for string SSH hosts in create_fleet_ssh_instance_model

### LINKED ISSUE #3475: [Bug]: Instances created for jobs are not divided into blocks
### Steps to reproduce

1. Create an elastic fleet with `blocks: auto`.

```shell
> cat fleets/default.dstack.yml
type: fleet
name: default
nodes: ..
blocks: auto

> dstack apply -f fleets/default.dstack.yml
```

2. Run a job on the fleet, so that it provisions an instance that could be divided into more than one block.

```shell
> cat .dstack.yml 
type: dev-environment
ide: vscode

> dstack apply -b aws --fleet default
```

### Actual behaviour

The provisioned instance is not divided into blocks.

```shell
> dstack fleet
 FLEET    INSTANCE  BACKEND          RESOURCES                        PRICE    STATUS  CREATED    
 default  0         aws (eu-west-3)  cpu=2 mem=8GB disk=100GB (spot)  $0.0292  busy    5 mins ago 

> dstack fleet get default --json | jq '.instances[0].total_blocks'
1
```

### Expected behaviour

The provisioned instance is divided into blocks (in this example, 2 blocks).

### dstack version

0.20.3

### Server logs

```shell

```

### Additional information

_No response_

mattate:
## Summary

Fixes #3475

When SSH fleet hosts are specified as plain strings (e.g., `hosts: ["192.168.0.1"]`), the fleet-level `blocks` configuration is ignored and hardcoded to `1`. This prevents instances from being divided into multiple GPU blocks, even when `blocks: auto` or a specific number is set in the fleet YAML.

## Root Cause

In `create_fleet_ssh_instance_model()`, the `if isinstance(host, str)` branch hardcodes `blocks = 1`, while the `else` branch (for `SSHHostParams` objects) correctly reads `blocks` from `host.blocks`. The fleet-level `spec.configuration.blocks` is never used for string hosts.

## Fix

Replace `blocks = 1` with `blocks = spec.configuration.blocks` in the string host branch, consistent with how fleet-level configuration is propagated for other SSH parameters (`ssh_user`, `ssh_key`, `port`) in the same branch.

## Steps to reproduce

1. Create an SSH fleet with `blocks: 3` (or `blocks: auto`):
```yaml
type: fleet
name: my-fleet
blocks: 3
ssh_config:
  user: myuser
  identity_file: ~/.ssh/id_rsa
  hosts:
    - 192.168.0.1
```
2. Apply: `dstack apply -f fleet.dstack.yml`
3. Check blocks: `dstack fleet get my-fleet --json | jq ".instances[0].total_blocks"`
4. Observe `total_blocks: 1` instead of `3`

After this fix, the instance correctly reports `total_blocks: 3`.

## AI Assistance Disclosure

This PR was written with AI assistance (code agent).

### DIFF
--- a/src/dstack/_internal/server/services/fleets.py
+++ b/src/dstack/_internal/server/services/fleets.py
@@ -684,7 +684,7 @@ async def create_fleet_ssh_instance_model(
         port = ssh_params.port
         proxy_jump = ssh_params.proxy_jump
         internal_ip = None
-        blocks = 1
+        blocks = spec.configuration.blocks
     else:
         hostname = host.hostname
         ssh_user = host.user or ssh_params.user


un-def:
Thanks for the contribution. Actually, this fix is not related to https://github.com/dstackai/dstack/issues/3475 but addresses another known issue – https://github.com/dstackai/dstack/issues/3278

The fix is partial, though – it works when a host is specified as a string, but the [object form](https://dstack.ai/docs/reference/dstack.yml/fleet/#ssh_config-hosts) still defaults to `blocks: 1` if `ssh_config.hosts[*].blocks` is not specified. I think this makes the situation worse:

Now: 
```yaml
blocks: 4
ssh_config:
  hosts:
    - 192.168.100.100  # blocks = 1, top-level blocks is ignored
    - hostname: 192.168.100.200  # blocks = 1, top-level blocks is ignored
```

With the fix:
```yaml
blocks: 4
ssh_config:
  hosts:
    - 192.168.100.100  # blocks = 4, top-level blocks value is used as a fallback
    - hostname: 192.168.100.200  # blocks = 1, top-level blocks is still ignored
```

@peterschmidt85 I think we should reopen https://github.com/dstackai/dstack/issues/3278 and make a proper fix with the following fallback chain: `ssh_config.hosts[n].blocks` → top-level `blocks` → `1`

un-def:
Closed in favor of https://github.com/dstackai/dstack/pull/3700

PR #3702, opened 2026-03-26
Fix SELinux denials on SSH fleet provisioning

peterschmidt85:
## Summary
- On SELinux-enforcing hosts (RHEL, Rocky, CentOS), files moved from `/tmp` retain their original SELinux context (`user_tmp_t`/`unconfined_u`). systemd cannot read these files, causing the shim service to fail with "Permission denied".
- Add `chcon` after `mv` to set correct SELinux contexts: `systemd_unit_file_t` for the service file, `etc_t` for the env file. No-op on non-SELinux systems.
- Replace `mv` with `cp`+`rm` for the shim binary to ensure correct context in `/usr/local/bin/`.

## Test plan
- [x] Tested on RHEL 9.4 with SELinux Enforcing — fleet goes active
- [x] Tested on Ubuntu 24.04 (no SELinux) — fleet goes active
- [x] All Python tests pass (2357 passed)
- [x] Pre-commit hooks pass

### DIFF
--- a/src/dstack/_internal/core/backends/base/compute.py
+++ b/src/dstack/_internal/core/backends/base/compute.py
@@ -906,7 +906,7 @@ def get_shim_pre_start_commands(
         f"dlpath=$(sudo mktemp -t {DSTACK_SHIM_BINARY_NAME}.XXXXXXXXXX)",
         # -sS -- disable progress meter and warnings, but still show errors (unlike bare -s)
         f'sudo curl -sS --compressed --connect-timeout 60 --max-time 240 --retry 1 --output "$dlpath" "{url}"',
-        f'sudo mv "$dlpath" {dstack_shim_binary_path}',
+        f'sudo cp "$dlpath" {dstack_shim_binary_path} && sudo rm "$dlpath"',
         f"sudo chmod +x {dstack_shim_binary_path}",
         f"sudo mkdir {dstack_working_dir} -p",
     ]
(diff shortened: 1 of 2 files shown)


src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -906,7 +906,7 @@ def get_shim_pre_start_commands(
         f"dlpath=$(sudo mktemp -t {DSTACK_SHIM_BINARY_NAME}.XXXXXXXXXX)",
         # -sS -- disable progress meter and warnings, but still show errors (unlike bare -s)
         f'sudo curl -sS --compressed --connect-timeout 60 --max-time 240 --retry 1 --output "$dlpath" "{url}"',
-        f'sudo mv "$dlpath" {dstack_shim_binary_path}',
+        f'sudo cp "$dlpath" {dstack_shim_binary_path} && sudo rm "$dlpath"',
```
un-def, line 909: I think we should replace `mv` with `sudo cp && rm` (or add `chown 0:0` after `mv`) elsewhere to ensure `root:root` ownership. For example, `shim.env` is owned by SSH user (as it uploaded via SFTP): 

```
# ls -l /root/.dstack
total 33236
-rw-r--r--  1 root   root        278 Mar 26 14:24 host_info.json
drwxr-xr-x 11 root   root       4096 Mar 26 16:26 runners
-rw-rw-r--  1 debian debian      395 Mar 26 14:24 shim.env
-rw-r--r--  1 root   root   34015187 Mar 27 07:14 shim.log
```

un-def (approved):

PR #3716, opened 2026-03-30
Reject user interaction in `runner_ssh_tunnel`

### LINKED ISSUE #3715: [Bug]: `dstack server` may fall back to interactive auth methods when connecting to instances/jobs
### Steps to reproduce

1. Create an SSH fleet.
2. Make sure `password` and/or `keyboard-interactive` auth methods are enabled on the instance side.
3. Comment out the `dstack` key in `authorized_keys` 

### Actual behaviour

When the `publickey` method fails,  `runner_ssh_tunnel` falls back to the next method (e.g., `password`), asks for a password, waits for 15 seconds, and fails.

### Expected behaviour

`runner_ssh_tunnel` should immediately fail on any user interaction request.

### dstack version

0.20.15

### Server logs

```shell
[11:06:29] DEBUG    dstack._internal.server.background.pipeline_tasks.base:342 Processing instances item cc9e0e59-4b3c-459a-8277-fc060095a4cd
debian@192.168.122.75's password:
[...]
[11:06:44] DEBUG    dstack._internal.core.services.ssh.tunnel:180 SSH tunnel to debian@192.168.122.75 did not open in 15 seconds
           WARNING  dstack._internal.server.background.pipeline_tasks.instances.check:279 Instance vm-0 check: reachable=False health_status=HEALTHY message='SSH or tunnel error'
```

### Additional information

_No response_

un-def:
Fixes: https://github.com/dstackai/dstack/issues/3715

### DIFF
--- a/src/dstack/_internal/core/services/ssh/tunnel.py
+++ b/src/dstack/_internal/core/services/ssh/tunnel.py
@@ -70,6 +70,7 @@ class SSHTunnel:
         ssh_config_path: Union[PathLike, Literal["none"]] = "none",
         port: Optional[int] = None,
         ssh_proxies: Iterable[tuple[SSHConnectionParams, Optional[FilePathOrContent]]] = (),
+        batch_mode: bool = False,
     ):
         """
         :param forwarded_sockets: Connections to the specified local sockets will be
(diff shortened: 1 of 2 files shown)


src/dstack/_internal/core/services/ssh/tunnel.py
```diff
@@ -79,11 +80,22 @@ def __init__(
...
+            jump connection will ignore this option -- in that case, you should replace `ProxyJump`
+            with explicit `ProxyCommand=ssh [...] -o BatchMode=yes` in your config.
         """
         self.destination = destination
         self.forwarded_sockets = list(forwarded_sockets)
         self.reverse_forwarded_sockets = list(reverse_forwarded_sockets)
         self.options = options
+        # A copy of options with names normalized to lowercase. Used only internally, the actual
+        # ssh command is built from user-provided options as is.
+        self.options_normalized = {k.lower(): v for k, v in options.items()}
```
jvstme: (nit) Duplicating the attribute can backfire if `options` ever becomes mutable. Technically, it is already mutable, since (i) it is a public attribute, and (ii) it contains a reference to the caller-provided mutable dict.

Some alternatives:
- lowercasing on-demand (on line 163)
- lowercasing directly in `options`
- `requests.structures.CaseInsensitiveDict`

jvstme (approved):

src/dstack/_internal/core/services/ssh/tunnel.py
un-def: > lowercasing on-demand (on line 163)

Dropped `options_normalized` in favor of ad-hoc normalization in `open_command()`

PR #3721, opened 2026-03-31
Support router as replica with pipelines

Bihan:
Refer design document for this PR is [here.](https://docs.google.com/document/d/1AbBb6de81Ez-dmQCjiIuKLbnzAmZ4Meed2IRNWnpZKo/edit?usp=sharing)


src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,259 @@
...
+    def _fetcher(self) -> Fetcher[ServiceRouterWorkerSyncPipelineItem]:
+        return self.__fetcher
+
+    @property
+    def _workers(self) -> Sequence["ServiceRouterWorkerSyncWorker"]:
+        return self.__workers
+
+
+class ServiceRouterWorkerSyncFetcher(Fetcher[ServiceRouterWorkerSyncPipelineItem]):
+    @sentry_utils.instrument_named_task("pipeline_tasks.ServiceRouterWorkerSyncFetcher.fetch")
```
r4victor: I recently added `@sentry_utils.instrument_pipeline_task` – use it to avoid hardcoding `pipeline_tasks` prefix.

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,259 @@
...
+            )
+            sync_row = res.unique().scalar_one_or_none()
+            if sync_row is None:
+                log_lock_token_mismatch(logger, item)
+                return
+            run_model = sync_row.run
+            if run_model is None:
+                await session.delete(sync_row)
+                await session.commit()
+                return
```
r4victor: How can `run_model` be None here?

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,259 @@
...
+                select(RunModel)
+                .where(RunModel.id == item.run_id)
+                .options(
+                    selectinload(RunModel.project),
+                    selectinload(RunModel.jobs).selectinload(JobModel.project),
+                    selectinload(RunModel.jobs)
+                    .selectinload(JobModel.instance)
+                    .selectinload(InstanceModel.project),
+                )
+            )
```
r4victor, line 257: This is potentially a very inefficient select – a run can have thousands of job submissions. Select only the jobs that the processing needs, i.e. only the router replica job. Also every selectinload will be a separate query here – not sure if it's justified. joinedload may be a better suited for a one-to-one rel. Also, try to avoid loading all models's columns and use load_only to select only the necessary.

src/dstack/_internal/server/services/runs/router_worker_sync.py
```diff
@@ -0,0 +1,345 @@
...
+    group_name = router_group.name
+    assert group_name is not None, "Replica group name is set by validation"
+    router_jobs = [
+        j
+        for j in run_model.jobs
+        if job_belongs_to_group(j, group_name) and j.status == JobStatus.RUNNING
+    ]
+    if not router_jobs or not is_replica_registered(router_jobs):
+        return None
+    return router_jobs[0]
```
r4victor, line 114: Can there be multiple router jobs? If so, how does that work?

src/dstack/_internal/server/services/runs/__init__.py
```diff
@@ -93,6 +95,40 @@
...
+def run_spec_has_router_replica_group(run_spec: RunSpec) -> bool:
+    if run_spec.configuration.type != "service":
+        return False
+    cfg = run_spec.configuration
+    if not isinstance(cfg, ServiceConfiguration):
+        return False
+    return any(g.router is not None for g in cfg.replica_groups)
+
+
+async def ensure_service_router_worker_sync_row(
```
r4victor: Why put these router-speicfic functions in top of runs services.

src/dstack/_internal/server/services/runs/__init__.py
```diff
@@ -621,6 +657,7 @@ async def submit_run(
                             ],
                         )
                     global_replica_num += 1
+            await ensure_service_router_worker_sync_row(session, run_model, run_spec)
```
r4victor, line 624: I think in-place update supports `replicas`. What happens if a user adds a router replica in in-place update if ensure_service_router_worker_sync_row() gets called only on submit_run()?

src/dstack/_internal/server/services/runs/__init__.py
```diff
@@ -93,6 +95,40 @@
...
+) -> None:
+    if not run_spec_has_router_replica_group(run_spec):
+        return
+    res = await session.execute(
+        select(ServiceRouterWorkerSyncModel.id).where(
+            ServiceRouterWorkerSyncModel.run_id == run_model.id
+        )
+    )
+    if res.scalar_one_or_none() is not None:
+        return
```
r4victor: How can it be that ServiceRouterWorkerSyncModel already exists for a run if ensure_service_router_worker_sync_row is called only on run submit?

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,259 @@
...
+                )
+                .options(selectinload(ServiceRouterWorkerSyncModel.run))
+            )
+            sync_row = res.unique().scalar_one_or_none()
+            if sync_row is None:
+                log_lock_token_mismatch(logger, item)
+                return
+            run_model = sync_row.run
+            if run_model is None:
+                await session.delete(sync_row)
```
r4victor: We generally use soft deletes in dstack server easier debugging and historical data. Assuming there will be very few `ServiceRouterWorkerSyncModel` rows (one per service replica router), I'd also soft-delete it for consistency.

src/dstack/_internal/server/models.py
```diff
@@ -454,6 +458,31 @@ class RunModel(PipelineModelMixin, BaseModel):
     )
 
 
+class ServiceRouterWorkerSyncModel(PipelineModelMixin, BaseModel):
```
r4victor, line 471: Let's put it somewhere in the end of the file so that "core" models come first.

src/dstack/_internal/server/services/jobs/job_replica_http_client.py
```diff
@@ -0,0 +1,49 @@
+"""SSH-tunneled async HTTP client to a job's service port (same path as probes)."""
```
r4victor, line 1: put this file in jobs services?

src/dstack/_internal/server/services/runs/router_worker_sync.py
```diff
@@ -0,0 +1,345 @@
+"""Reconcile SGLang router /workers with dstack's registered worker replicas (async, SSH-tunneled)."""
```
r4victor, line 1: put this file in runs services

r4victor (changes requested):
Did a quick review of the pipeline code. Haven't looked into the worker sync logic.

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: I thought what if the run row can be hard-deleted, so sync_row.run becomes None. If this is not possible we can delete this block.

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: Please check if below proposed query addresses the concerns

1. Avoid loading thousands of job submissions: no longer load RunModel.jobs unconditionally. The selectinload(RunModel.jobs.and_(...)) restricts the loaded jobs to only RUNNING + registered replicas, which are the only ones sync_router_workers_for_run_model() can use (router job selection and worker list building both ignore non‑running / unregistered jobs).

2. selectinload is intentional: RunModel.jobs is a one‑to‑many collection; using joinedload would duplicate the RunModel row per job.

3. joinedload for one‑to‑one/many‑to‑one: RunModel.project, JobModel.project, JobModel.instance, InstanceModel.project are loaded with joinedload because these are scalar relationships from from run,job and instance.

4. Use load_only: This limits columns required by `sync_router_workers_for_run_model(run_for_sync)` and `_get_service_replica_client(job_model)`

```
res = await session.execute(
    select(RunModel)
    .where(RunModel.id == item.run_id)
    .options(
        load_only(RunModel.id, RunModel.run_spec),
        selectinload(
            RunModel.jobs.and_(
                JobModel.status == JobStatus.RUNNING,
                JobModel.registered == true(),
            )
        )
        .load_only(
            JobModel.id,
            JobModel.status,
            JobModel.registered,
            JobModel.job_spec_data,
            JobModel.job_provisioning_data,
            JobModel.job_runtime_data,
        )
        .options(
            joinedload(JobModel.project).load_only(ProjectModel.id, ProjectModel.ssh_private_key),
            joinedload(JobModel.instance)
            .load_only(InstanceModel.id, InstanceModel.remote_connection_info)
            .joinedload(InstanceModel.project)
            .load_only(ProjectModel.id, ProjectModel.ssh_private_key),
        ),
    )
)
```

src/dstack/_internal/server/services/runs/router_worker_sync.py
Bihan: For the first iteration, I suggest restricting the router replica group to count: 1 via configuration validation. The current sync logic effectively assumes a single active router job. We can extend this later to support multiple router replicas for HA.

src/dstack/_internal/server/services/runs/__init__.py
Bihan: I kept it there because they are used by run lifecycle. Should I shift them to  `src/dstack/_internal/server/services/router_worker_sync.py`?

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
r4victor: But you defined run_id as non-optional with ondelete="CASCADE" - how can it be possible?

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
r4victor: looks good, at least at a glance

src/dstack/_internal/server/services/runs/router_worker_sync.py
r4victor: it's worth a comment!

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: You are right. Maybe I delete this block.

src/dstack/_internal/server/services/runs/__init__.py
Bihan: Thanks for pointing out. I need to call `ensure_service_router_worker_sync_row` after [this ](https://github.com/Bihan/dstack/blob/e155d17a7edcf13cf2a930acd70cec52f1974311/src/dstack/_internal/server/services/runs/__init__.py#L528)

src/dstack/_internal/server/services/runs/__init__.py
r4victor: I mean at least they should not be at the top of the file.

Bihan: Got it.

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,275 @@
...
+            run_model = sync_row.run
+            if (
+                run_model.deleted
+                or run_model.status.is_finished()
+                or run_model.status != RunStatus.RUNNING
+                or not run_model_has_router_replica_group(run_model)
+            ):
+                await session.delete(sync_row)
+                await session.commit()
+                return
```
jvstme: Does this condition ever pass, and is really necessary, considering the fetcher fetches based on [`RunModel.status == RunStatus.RUNNING`](https://github.com/Bihan/dstack/blob/7b268cba676e850ff34c1f72b17a4d72ffe3d848/src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py#L121)?

src/dstack/_internal/server/services/job_replica_http_client.py
```diff
@@ -0,0 +1,49 @@
...
+from dstack._internal.server.models import JobModel
+from dstack._internal.server.services.jobs import get_job_spec
+from dstack._internal.server.services.ssh import container_ssh_tunnel
+from dstack._internal.utils.common import get_or_error
+
+SSH_CONNECT_TIMEOUT = timedelta(seconds=10)
+
+
+@asynccontextmanager
+async def _get_service_replica_client(
```
jvstme: (nit) This function is supposed to be imported in other modules, so it shouldn't be private (prefixed with `_`).

src/dstack/_internal/core/models/configurations.py
```diff
@@ -801,6 +801,12 @@ class ReplicaGroup(CoreModel):
         CommandsList,
         Field(description="The shell commands to run for replicas in this group"),
     ] = []
+    router: Annotated[
+        Optional[AnyServiceRouterConfig],
```
jvstme: `AnyServiceRouterConfig` has the `policy` and `pd_disaggregation` properties that are not applicable here. I think we might need a separate class for `ReplicaGroup.router`, without these options.

(see [this thread](https://docs.google.com/document/d/140bj2N9ZHl7dz1_QS7RZ-MX_9ANKxQufAjvoXI_PmDM/edit?disco=AAAB1lz1e7U))

src/dstack/_internal/server/services/runs/router_worker_sync.py
```diff
@@ -0,0 +1,348 @@
...
+        return
+    try:
+        async with _get_service_replica_client(router_job) as client:
+            await _update_workers_in_router_replica(client, target_workers)
+    except Exception as e:
+        logger.warning(
+            "%s: failed to sync workers with router: %r",
+            fmt(router_job),
+            e,
+        )
```
jvstme, line 346: `except Exception` may result in bugs being silenced (warnings are not reported to Sentry, so we generally don't see them).

If there are any specific exceptions that should be ignored with a warning, consider listing them explicitly in `except`. I assume this is the case for communication issues with the router (such as SSH or HTTP errors), because the router can become unreachable or can be misconfigured or broken by the user

src/dstack/_internal/server/services/runs/router_worker_sync.py
```diff
@@ -0,0 +1,348 @@
...
+    pass
+
+
+async def _stream_response_body_bytes(resp: Response, max_bytes: int) -> bytes:
+    buf = bytearray()
+    async for chunk in resp.aiter_bytes():
+        buf.extend(chunk)
+        if len(buf) > max_bytes:
+            raise _ResponseTooLargeError()
+    return bytes(buf)
```
jvstme, line 44: (nit) We have the `join_byte_stream_checked` function that appears to do the same thing

src/dstack/_internal/proxy/gateway/services/registry.py
```diff
@@ -291,6 +293,13 @@ async def apply_service(
             )
             for replica, conn in replica_conns.items()
         ]
+    router_replicas = [r for r in service.replicas if r.is_router_replica]
+    if router_replicas:
+        replica_configs_for_nginx = [c for c in replica_configs if c.id == router_replicas[0].id]
```
jvstme: What if the router replica is not yet registered, or temporarily unregistered (e.g., if it failed and is being restarted)? It seems that the gateway will then assume that the service doesn't have a router replica, and Nginx will direct incoming requests directly to worker replicas, which is not expected.

Also, do we actually need to register worker replicas on the gateway, considering the gateway should only communicate with the router replica? My [initial proposal](https://docs.google.com/document/d/140bj2N9ZHl7dz1_QS7RZ-MX_9ANKxQufAjvoXI_PmDM/edit?disco=AAAB1lz1fBY) was not to register them, and I think that would fix the problem above, and also optimize and simplify a few things (no need for extra network communication, no need to distinguish between router and worker replicas on the gateway, etc).

src/dstack/_internal/core/models/configurations.py
```diff
@@ -801,6 +801,12 @@ class ReplicaGroup(CoreModel):
         CommandsList,
         Field(description="The shell commands to run for replicas in this group"),
     ] = []
+    router: Annotated[
```
jvstme, line 804: Add excludes in `core/compatibility`, for client compatibility with older servers

src/dstack/_internal/proxy/gateway/services/registry.py
jvstme, line 1: Is there anything to prevent exposing the router replica's `/workers` API on Nginx?

See the second comment in [this thread](https://docs.google.com/document/d/140bj2N9ZHl7dz1_QS7RZ-MX_9ANKxQufAjvoXI_PmDM/edit?disco=AAAB1lz1fBY)

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
jvstme, line 1: If you have some time, consider covering the pipeline by unit tests. I think all (or at least most of) our existing pipelines and background tasks currently have good coverage

src/dstack/_internal/proxy/gateway/services/registry.py
Bihan: Yes. For SGLang, Nginx only allows the whitelisted paths from `get_nginx_service_config()`; everything else hits location / { return 403; }.

I have verified with curl on the service domain.

src/dstack/_internal/proxy/gateway/services/registry.py
Bihan: Done

src/dstack/_internal/server/services/job_replica_http_client.py
Bihan: Done

src/dstack/_internal/proxy/lib/services/service_connection.py
```diff
@@ -151,7 +151,9 @@ async def get_service_replica_client(
             timeout=HTTP_TIMEOUT,
         )
     # Nginx not available, forward directly to the tunnel
-    replica = random.choice(service.replicas)
+    router_replicas = [r for r in service.replicas if r.is_router_replica]
+    replicas_to_use = router_replicas if router_replicas else service.replicas
+    replica = random.choice(replicas_to_use)
```
jvstme: Same [issue](https://github.com/dstackai/dstack/pull/3721#discussion_r3061296546) as on the gateway — if the router replica is not yet registered, the in-server proxy will decide to forward requests directly to worker replicas.

The solution I can suggest is to update `ServerProxyRepo.get_service`, so that it only returns the replicas that the proxy should be aware of (router replicas if the service config defines a router, all replicas otherwise). This will be consistent with `GatewayProxyRepo.get_service`, which also doesn't return worker replicas (because they are not registered and not stored in the repo).

This will also allow to drop `Replica.is_router_replica` and `RegisterReplicaRequest.is_router_replica`.

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,286 @@
...
+        )
+        async with sync_lock:
+            async with get_session_ctx() as session:
+                now = get_current_datetime()
+                res = await session.execute(
+                    select(ServiceRouterWorkerSyncModel)
+                    .join(RunModel, RunModel.id == ServiceRouterWorkerSyncModel.run_id)
+                    .where(
+                        ServiceRouterWorkerSyncModel.deleted == false(),
+                        RunModel.status == RunStatus.RUNNING,
```
jvstme: The fetcher only fetches `running` runs, yet the worker has `ServiceRouterWorkerSyncModel` deletion logic for [non](https://github.com/Bihan/dstack/blob/3bc04dfce13de760d932babfea3ebd6162e7b470/src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py#L208)-`running` runs. Doesn't that mean that `ServiceRouterWorkerSyncModel` won't actually be deleted when the run stops?

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,286 @@
...
+            if (
+                run_model.deleted
+                or run_model.status.is_finished()
+                or run_model.status != RunStatus.RUNNING
+                or not run_model_has_router_replica_group(run_model)
+            ):
+                sync_row.deleted = True
+                sync_row.lock_expires_at = None
+                sync_row.lock_token = None
+                sync_row.lock_owner = None
```
jvstme: I think pipeline workers are [not supposed](https://github.com/dstackai/dstack/blob/da026f9ac0aac3b0f1ca6504e50f34dae7ea7de0/contributing/PIPELINES.md?plain=1#L33) to mutate ORM models and should use `update_map` instead, to handle locking correctly

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,286 @@
...
+                    .where(
+                        ServiceRouterWorkerSyncModel.id == item.id,
+                        ServiceRouterWorkerSyncModel.lock_token == item.lock_token,
+                    )
+                    .values(
+                        deleted=True,
+                        lock_expires_at=None,
+                        lock_token=None,
+                        lock_owner=None,
+                    )
```
jvstme: Shouldn't this also set `last_processed_at`? At least for debugging purposes, so that it's possible to tell when the entry was deleted.

I'd also suggest to refactor this piece so that it populates `update_map` instead of doing its own `update` query. Allow the `update` query at the end of this method to perform any updates, because it is already written correctly to set all the standard fields (`last_processed_at`, `lock_expires_at`, etc), log a warning if the update fails, etc.

src/dstack/_internal/server/services/runs/router_worker_sync.py
```diff
@@ -0,0 +1,345 @@
...
+        logger.debug("server_info response too large for worker %s", worker_url)
+    except Exception as e:
+        logger.debug("Could not fetch server_info for worker %s: %r", worker_url, e)
+    return {"status": "not_ready", "payload": None}
+
+
+async def _build_target_workers(
+    run_model: RunModel,
+    run_spec: RunSpec,
+    replica_groups: List,
```
jvstme: (nit) `list[ReplicaGroup]`

src/dstack/_internal/server/services/runs/router_worker_sync.py
```diff
@@ -0,0 +1,345 @@
...
+    worker_type: str
+    bootstrap_port: NotRequired[Optional[int]]
+
+
+def run_model_has_router_replica_group(run_model: RunModel) -> bool:
+    run_spec = RunSpec.__response__.parse_raw(run_model.run_spec)
+    return run_spec_has_router_replica_group(run_spec)
+
+
+def _get_router_job(run_model: RunModel, router_group) -> Optional[JobModel]:
```
jvstme: (nit) Annotate `router_group`

src/dstack/_internal/server/services/runs/router_worker_sync.py
```diff
@@ -0,0 +1,345 @@
...
+    try:
+        current = await _get_router_workers(client)
+        worker_id = None
+        for w in current:
+            u = w.get("url")
+            if u and isinstance(u, str) and u.rstrip("/") == worker_url.rstrip("/"):
+                wid = w.get("id")
+                if wid and isinstance(wid, str):
+                    worker_id = wid
+                    break
```
jvstme: Calling `GET /workers` for each worker just to get the worker ID is suboptimal — we already call `GET /workers` from `_update_workers_in_router_replica`, so we could build a mapping from worker URLs to IDs there instead

src/dstack/_internal/server/services/runs/router_worker_sync.py
```diff
@@ -0,0 +1,345 @@
...
+                        "payload": {"url": worker_url, "worker_type": "decode"},
+                    }
+                return {
+                    "status": "ready",
+                    "payload": {"url": worker_url, "worker_type": "regular"},
+                }
+    except _ResponseTooLargeError:
+        logger.debug("server_info response too large for worker %s", worker_url)
+    except Exception as e:
+        logger.debug("Could not fetch server_info for worker %s: %r", worker_url, e)
```
jvstme: In other similar places, you are using `warning` and `exception`, respectively, which I think are more appropriate than `debug`, because these messages are important for troubleshooting, and `debug` will be filtered out in production

src/dstack/_internal/proxy/gateway/services/registry.py
jvstme, line 1: Since it will be possible to use the SGLang router for services without a gateway (using the in-server proxy), I think we might need to implement a path whitelist mechanism in the [in-server proxy](https://github.com/Bihan/dstack/blob/3bc04dfce13de760d932babfea3ebd6162e7b470/src/dstack/_internal/server/services/proxy/services/service_proxy.py#L21) similar to how it's implemented here in the gateway

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -219,7 +224,6 @@ async def process(self, item: ServiceRouterWorkerSyncPipelineItem) -> None:
                     selectinload(
                         RunModel.jobs.and_(
                             JobModel.status == JobStatus.RUNNING,
-                            JobModel.registered == true(),
```
jvstme: Why this change? Without checking for `registered`, the pipeline now doesn't respect probes (`registered` is only set to `True` after probes pass), so it will register replicas with the SGLang router before their probes pass, which is unexpected.

Btw, the semantics of `registered` are now a bit blurred, since SGLang worker replicas will have `registered=True` before they are actually registered to receive requests. Maybe it's worth renaming the field, but I wouldn't suggest doing it in the same PR

src/dstack/_internal/server/services/runs/__init__.py
jvstme: > What happens if a user adds a router replica in in-place update

@Bihan, is this use case expected to work at all? I think it won't work with the current implementation, because adding a router replica means that only this replica should receive requests, which means that other existing replicas should be unregistered from the gateway, which doesn't seem to be implemented.

Similarly, due to the need to register or unregister existing replicas, I assume that the following use cases won't work as expected:

- Removing a router replica group.
- Adding the `router` property to an existing replica group.
- Removing the `router` property from an existing replica group.

If supporting these use cases requires additional effort, I can suggest to forbid them for now (see `_check_can_update_configuration`). And, in that case, only call `ensure_service_router_worker_sync_row` here and simplify its implementation

src/dstack/_internal/proxy/gateway/services/registry.py
jvstme: Could you explain how this works? From what I can see, the whitelist seems to only be applied when [`Service.router`](https://github.com/Bihan/dstack/blob/3bc04dfce13de760d932babfea3ebd6162e7b470/src/dstack/_internal/proxy/gateway/services/registry.py#L344) is populated, which is [based](https://github.com/Bihan/dstack/blob/3bc04dfce13de760d932babfea3ebd6162e7b470/src/dstack/_internal/server/services/services/__init__.py#L137) on the old-style service-level and gateway-level `router` config property, rather than the new replica-level `router` property. Or am I missing something?

Bihan: Done

src/dstack/_internal/server/services/runs/router_worker_sync.py
Bihan: Done

src/dstack/_internal/server/services/runs/router_worker_sync.py
Bihan: Done

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: Done

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: Done

src/dstack/_internal/proxy/lib/services/service_connection.py
Bihan: Done

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: Done

src/dstack/_internal/core/models/configurations.py
Bihan: Done

src/dstack/_internal/server/services/runs/router_worker_sync.py
Bihan: Done

src/dstack/_internal/core/models/configurations.py
Bihan: Done

src/dstack/_internal/server/services/job_replica_http_client.py
Bihan: Done

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: I have reverted. On the note about semantics: I agree that registered is acting more like “dstack-probe-ready” than “registered with the router.”

src/dstack/_internal/proxy/gateway/services/registry.py
Bihan: Yes you are right. I was hitting /workers without bearer token, which resulted in 403. I will fix this.

src/dstack/_internal/server/services/runs/router_worker_sync.py
Bihan: Done

src/dstack/_internal/server/services/runs/router_worker_sync.py
Bihan: Done

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: Thanks for pointing out. Done as it is supposed to.

src/dstack/_internal/proxy/gateway/services/registry.py
Bihan: Done

docs/blog/posts/pd-disaggregation.md
jvstme, line 1: (nit) In my view, blog posts should generally remain unchanged, as they are timestamped and serve a historical purpose. As a reader, I wouldn't expect their content to change significantly over time.

I would keep the blog post as is, but add a note at the top indicating that gateway routers are deprecated, along with a reference to the relevant replica-group routers docs or examples.

docs/docs/concepts/gateways.md
```diff
@@ -95,7 +95,11 @@ router:
 
 If you configure the `sglang` router, [services](../concepts/services.md) can run either [standard SGLang workers](../../examples/inference/sglang/index.md) or [Prefill-Decode workers](../../examples/inference/sglang/index.md#pd-disaggregation) (aka PD disaggregation).
 
-> Note, if you want to run services with PD disaggregation, the gateway must currently run in the same cluster as the service.
+!!! note "PD disaggregation"
+    To run services with PD disaggregation see [SGLang PD disaggregation](https://dstack.ai/examples/inference/sglang/#pd-disaggregation).
+
+!!! note "Deprecation"
+    Configuring the SGLang router in a gateway will be deprecated in a future release.
```
jvstme: (nit) I'd put this at the top of the Router section, so that users don't have to read all of it before they find out that it's irrelevant.

Or even remove the section.


Also:

> will be deprecated in a future release

More like *is* deprecated and will be disallowed in a future release?

examples/inference/sglang/README.md
```diff
@@ -156,53 +171,53 @@ replicas:
...
+    - 89.169.108.16   # CPU Host (router)
+    - 89.169.123.100  # GPU Host (prefill/decode workers)
+    - 89.169.110.65   # GPU Host (prefill/decode workers)
 ```
 
 </div>
 
-<!-- TODO: Gateway creation using fleets is coming to simplify this. -->
+!!! note "Gateway-based routing (deprecated)"
+    If you create a gateway with the [`sglang` router](https://dstack.ai/docs/concepts/gateways/#sglang), you can also run SGLang with PD disaggregation. This method will be deprecated in the future in favor of running the router as a replica.
```
jvstme: (nit)

> will be deprecated in the future

More like *is* deprecated and will be disallowed in the future?

examples/inference/sglang/README.md
```diff
@@ -156,53 +171,53 @@ replicas:
...
 
 Currently, auto-scaling only supports `rps` as the metric. TTFT and ITL metrics are coming soon.
 
-#### Gateway
-
-Note, running services with PD disaggregation currently requires the gateway to run in the same cluster as the service.
+#### SSH fleet
 
-For example, if you run services on the `kubernetes` backend, make sure to also create the gateway in the same backend:
+Create an [SSH fleet](https://dstack.ai/docs/concepts/fleets/#apply-a-configuration) that includes one CPU host for the router and one or more GPU hosts for the workers. Make sure the CPU and GPU hosts are in the same network.
```
jvstme: (nit) Does it have to be an SSH fleet specifically? I thought elastic (`nodes: 0..`) cloud and kubernetes fleets could work too — just don't specify any resource constraints in the fleet, and dstack will automatically provision the correct instances (both CPU and GPU, in the same fleet) based on the `resources` specified in `replicas` in the run configuration.

Only some backends won't work — like Nebius, which requires all instances in the cluster to be homogeneous.

The are more references to SSH fleets in the docs updated in this PR

examples/inference/sglang/README.md
```diff
@@ -156,53 +171,53 @@ replicas:
...
     resources:
       gpu: H200
 
 port: 8000
 model: zai-org/GLM-4.5-Air-FP8
+# SSH fleet containing both router (CPU) and workers (GPU).
+fleets: [pd-disagg]
 
-# Custom probe is required for PD disaggregation
+# Custom probe is required for PD disaggregation.
```
jvstme, line 183: (nit) By the way, is it still required? I thought `sync_router_workers_for_run_model` can gracefully handle the router or workers not being ready, and perform the registration eventually, once they become ready

src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -1075,7 +1075,8 @@ def get_dstack_gateway_wheel(build: str, router: Optional[AnyGatewayRouterConfig
     if build == "latest":
         build = _fetch_version(f"{base_url}/latest-version") or "latest"
         logger.debug("Found the latest gateway build: %s", build)
-    wheel = f"{base_url}/dstack_gateway-{build}-py3-none-any.whl"
+    # wheel = f"{base_url}/dstack_gateway-{build}-py3-none-any.whl"
+    wheel = "https://bihan-test-bucket.s3.eu-west-1.amazonaws.com/dstack_gateway-0.0.1-py3-none-any.whl"
```
jvstme: *reminder to revert before merging*

src/dstack/_internal/core/models/routers.py
```diff
@@ -31,6 +31,16 @@ class SGLangGatewayRouterConfig(CoreModel):
...
+    managed_by: Annotated[
+        Literal["gateway", "service"],
+        Field(
+            description=(
+                "Where the router process is managed. "
+                "`gateway`: the gateway runs and manages the router process. "
+                "`service`: the router runs inside the service (replica-group `router:`)."
+            )
+        ),
+    ] = "gateway"
```
jvstme: SGLangServiceRouterConfig is used for the router field in the run configuration, so now users can write something like that in the configuration, which doesn't make sense:

```yaml
router:
  type: sglang
  managed_by: service
```

To my understanding, [this](https://github.com/Bihan/dstack/blob/59d246bbcefebe399ea9d14c7236c254e48845c9/src/dstack/_internal/server/services/services/__init__.py#L280) is the motivation behind that field:

> When the gateway has no global SGLang config but a replica group does declare `router:`, we still return a default `SGLangServiceRouterConfig` so nginx/proxy code can treat the service as SGLang and apply path rules.

But I think think this shouldn't result in exposing an internal field to the user. Additionally, we'll drop `SGLangServiceRouterConfig` in a future release, yet the path rules will need to keep working.

So I would prefer to find another way to communicate the presence of a router replica to the gateway / proxy. For example, a boolean `RegisterServiceRequest.has_sglang_router_replica` and `Service.has_sglang_router_replica` could solve it.

Or we could pass `ReplicaGroupRouterConfig` in a field other than `RegisterServiceRequest.router` (but I'm not sure about that solution, because in the future `ReplicaGroupRouterConfig` can get server-only subproperties that are not relevant for the gateway).

src/dstack/_internal/core/models/configurations.py
```diff
@@ -1032,6 +1038,37 @@ def validate_replica_groups_have_commands_or_image(cls, values):
...
+        if router_groups:
+            router_group = router_groups[0]
+            if router_group.count.min != 1 or router_group.count.max != 1:
+                raise ValueError("For now replica group with `router` must have `count: 1`.")
+        return values
+
+    @root_validator()
+    def validate_replica_group_router_mutex(cls, values):
+        """
+        When a replica group sets `router:`, service-level `router` must be omitted.
```
jvstme, line 1058: 👍

src/dstack/_internal/server/migrations/versions/2026/03_29_1200_e7f4a91b2c3d_add_service_router_worker_sync.py
```diff
@@ -0,0 +1,66 @@
...
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_service_router_worker_sync")),
+    )
+    op.create_index(
+        op.f("ix_service_router_worker_sync_pipeline_fetch_q"),
+        "service_router_worker_sync",
+        [sa.literal_column("last_processed_at ASC")],
+        unique=False,
+    )
```
jvstme: Isn't this missing the `sqlite_where` and `postgresql_where` that are [present](https://github.com/Bihan/dstack/blob/59d246bbcefebe399ea9d14c7236c254e48845c9/src/dstack/_internal/server/models.py#L496) in `models.py`?

Not sure why they weren't added automatically. I'd suggest to try and re-generate the migration (using [Alembic](https://github.com/dstackai/dstack/blob/master/contributing/MIGRATIONS.md), in case that's not what you were using previously)

src/dstack/_internal/server/services/proxy/services/service_proxy.py
```diff
@@ -36,11 +37,16 @@ async def proxy(
...
 
-    client = await get_service_replica_client(service, repo, service_conn_pool)
-
     if not service.strip_prefix:
         path = concat_url_path(request.scope.get("root_path", "/"), request.url.path)
 
+    if service.router is not None and service.router.type == RouterType.SGLANG:
+        path_for_match = path if path.startswith("/") else f"/{path}"
+        if not _is_whitelisted_path(path_for_match, _SGLANG_WHITELISTED_PATHS):
+            raise ProxyError("Path is not allowed for this service", status.HTTP_404_NOT_FOUND)
```
jvstme: (nit) 403 Forbidden for consistency with the gateway and better semantics?

src/dstack/_internal/server/services/proxy/services/service_proxy.py
```diff
@@ -68,6 +74,23 @@ async def proxy(
     )
 
 
+_SGLANG_WHITELISTED_PATHS = (
+    "/generate",
+    "/v1/",
+    "/chat/completions",
+)
```
jvstme: (nit) Duplicates the [list from the gateway](https://github.com/Bihan/dstack/blob/59d246bbcefebe399ea9d14c7236c254e48845c9/src/dstack/_internal/proxy/gateway/services/registry.py#L343). Consider importing the constant from some common place, like `proxy/lib/const.py`

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -0,0 +1,305 @@
...
+                await session.execute(
+                    update(ServiceRouterWorkerSyncModel)
+                    .where(
+                        ServiceRouterWorkerSyncModel.id == item.id,
+                        ServiceRouterWorkerSyncModel.lock_token == item.lock_token,
+                    )
+                    .values(**early_cleanup_update_map)
+                )
+                await session.commit()
+                return
```
jvstme: (nit) Missing the `log_lock_token_changed_after_processing` call in case the update fails.

To avoid such discrepancies, consider refactoring, so that there is only one place that performs the update (currently there are three in the same method)

src/dstack/_internal/server/services/runs/__init__.py
Bihan: @jvstme 
The use case where we need in-place update is when user changes router start command.

Example:
From
`python -m sglang_router.launch_router --host 0.0.0.0 --port 8000 --pd-disaggregation --prefill-policy cache_aware`

To 
`python -m sglang_router.launch_router --host 0.0.0.0 --port 8000 --pd-disaggregation --prefill-policy round_robin`

I think we can forbid these cases in `_check_can_update_configuration`

- Removing a router replica group.

- Adding the router property to an existing replica group.

- Removing the router property from an existing replica group.

and only allow `commands` update case for router group.

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
Bihan: Done

src/dstack/_internal/server/services/proxy/services/service_proxy.py
Bihan: Done

src/dstack/_internal/server/services/proxy/services/service_proxy.py
Bihan: Done

src/dstack/_internal/server/migrations/versions/2026/03_29_1200_e7f4a91b2c3d_add_service_router_worker_sync.py
Bihan: Done

examples/inference/sglang/README.md
Bihan: Yes this is still required. Because probes queries `/v1/chat/completions` to register the job but router fails to serve `/v1/chat/completions` until workers are registered. Meanwhile, the router-worker sync pipeline only considers `RUNNING `jobs that are also `registered=True`.

jvstme: Oh, I see, so our [default probe](https://github.com/dstackai/dstack/blob/09a6a6a23d02cf9acce1ac48af6b7c008793f81c/src/dstack/_internal/server/services/jobs/configurators/base.py#L411) is the problem. But I assume it's possible to work around it by either setting `probes: []`, or not setting `model`. If that's the case, a custom probe is more of a recommendation, not a strict requirement.

Anyways, I think we were going to improve the UX here by introducing a different default probe for services with the SGLang router. Not in this PR, of course.

src/dstack/_internal/server/services/runs/__init__.py
jvstme: > The use case where we need in-place update is when user changes router start command.

Yes, and there are other use cases too, like changing `resources`. I only suggest to forbid the cases that are known to not work correctly (adding or removing a router replica group, adding or removing the `router` property to/from an existing replica group), and allow everything else.

jvstme: By the way, here's another example of how some in-place update cases could lead to unexpected results.

Imagine that a user has a regular service without a router, and they decide to "upgrade" it to a router, by defining new worker replica groups and a router replica group. This may even appear to work correctly, because old replicas will be stopped and new replicas will be registered as expected. But the service won't be re-registered, so `Service.router` and `Service.has_router_replica` on the gateway will look like there is no router, and the gateway won't apply the path whitelist, exposing the internal `/workers` API.

So this can also hurt security, not just UX, which is why I think it's important to fix it (by forbidding the in-place updates that don't work).

src/dstack/_internal/server/background/pipeline_tasks/service_router_worker_sync.py
```diff
@@ -219,15 +219,9 @@ async def process(self, item: ServiceRouterWorkerSyncPipelineItem) -> None:
...
                 now = get_current_datetime()
                 resolve_now_placeholders(early_cleanup_update_map, now=now)
-                await session.execute(
-                    update(ServiceRouterWorkerSyncModel)
-                    .where(
-                        ServiceRouterWorkerSyncModel.id == item.id,
-                        ServiceRouterWorkerSyncModel.lock_token == item.lock_token,
-                    )
-                    .values(**early_cleanup_update_map)
+                await _update_sync_row_or_log_lock_token_changed(
```
jvstme, line 222: (nit) Identical `set_processed_update_map_fields`, `set_unlock_update_map_fields`, and `resolve_now_placeholders` calls are also repeated in three places in this method. It's worth moving them inside `_update_sync_row_or_log_lock_token_changed`

src/dstack/_internal/server/services/services/__init__.py
```diff
@@ -266,12 +275,24 @@ def _build_service_router_config(
...
+    has_replica_group_router = any(
+        g.router is not None for g in service_configuration.replica_groups
+    )
     if not _gateway_has_sglang_router(gateway_configuration):
-        return None
+        if not has_replica_group_router:
+            return None
+        # In later releases we will deprecate service-level and gateway-level router
+        # configuration and return `ReplicaGroupRouterConfig` here instead.
+        return SGLangServiceRouterConfig()
```
jvstme: (nit) This may work too, but passing a fake `{"type": "sglang", "policy": "cache_aware", "pd_disaggregation": false}` object to the gateway just to trick it into enabling the SGLang whitelist doesn't look very straightforward to me.

I think we could just return `None` here. Leave the `Service.router` field exclusive to the deprecated gateway-based routers. And use both `Serivce.router` and `Service.has_router_replica` on the gateway / proxy to determine whether to enable the whitelist

src/dstack/_internal/server/services/proxy/services/service_proxy.py
```diff
@@ -36,11 +38,16 @@ async def proxy(
...
 
-    client = await get_service_replica_client(service, repo, service_conn_pool)
-
     if not service.strip_prefix:
         path = concat_url_path(request.scope.get("root_path", "/"), request.url.path)
 
+    if service.router is not None and service.router.type == RouterType.SGLANG:
+        path_for_match = path if path.startswith("/") else f"/{path}"
+        if not _is_whitelisted_path(path_for_match, SGLANG_WHITELISTED_PATHS):
+            raise ProxyError("Path is not allowed for this service", status.HTTP_403_FORBIDDEN)
```
jvstme: The whitelist doesn't seem to work. Testing the [example configuration](https://github.com/Bihan/dstack/blob/support_router_replica_with_pipelines/examples/inference/sglang/pd.dstack.yml), plus `gateway: false`, plus `auth: false`, minus `scaling`.

```shell
$ curl http://localhost:3000/proxy/services/main/prefill-decode/workers
{"workers":[],"total":0,"stats":{"prefill_count":0,"decode_count":0,"regular_count":0}}
```

Note that `service.router` is only [populated](https://github.com/Bihan/dstack/blob/0f8f1b6f1802e9c3c2f3b9f41c33579cc7d7d6dc/src/dstack/_internal/server/services/proxy/repo.py#L78) for gateway-based routers. It could be easier to rely on `service.has_router_replica`, although then I would recommend to [do the same thing](https://github.com/dstackai/dstack/pull/3721#discussion_r3086091943) on the gateway for consistency

jvstme (approved):
Looks good to me overall, but the following may require more attention:

- Forbid unsupported in-place updates ([thread](https://github.com/dstackai/dstack/pull/3721#discussion_r3070262165))
- Fix the path whitelist in the in-server proxy ([thread](https://github.com/dstackai/dstack/pull/3721#discussion_r3086429387))

These may be uncommon cases, but they are security-related, so I would prefer to
 address them before merging, or at least before the release

examples/inference/sglang/README.md
jvstme: For example, seems to work on AWS:

```yaml
type: fleet
name: aws-cluster
placement: cluster
backends: [aws]
nodes: 0..
```

```shell
$ dstack ps
 NAME                  BACKEND          GPU            PRICE           STATUS   SUBMITTED   
 prefill-decode                         -              -               running  10 mins ago 
    group=0 replica=0  aws (us-east-2)  -              $0.0383 (spot)  running  10 mins ago 
    group=1 replica=1  aws (us-east-2)  gpu=T4:16GB:1  $0.1372 (spot)  pulling  10 mins ago 
    group=2 replica=2  aws (us-east-2)  gpu=T4:16GB:1  $0.1372 (spot)  pulling  10 mins ago
```

```shell
$ dstack fleet
 aws-cluster      0..    -              auto  aws              -        active          27 mins ago 
    instance=0           -              spot  aws (us-east-2)  $0.0383  busy            21 mins ago 
    instance=1           gpu=T4:16GB:1  spot  aws (us-east-2)  $0.1372  busy (warning)  18 mins ago 
    instance=2           gpu=T4:16GB:1  spot  aws (us-east-2)  $0.1372  busy            18 mins ago
```

Bihan:
> Looks good to me overall, but the following may require more attention:
> 
> * Forbid unsupported in-place updates ([thread](https://github.com/dstackai/dstack/pull/3721#discussion_r3070262165))
> * Fix the path whitelist in the in-server proxy ([thread](https://github.com/dstackai/dstack/pull/3721#discussion_r3086429387))
> 
> These may be uncommon cases, but they are security-related, so I would prefer to address them before merging, or at least before the release

@jvstme Done

src/dstack/_internal/server/services/proxy/services/service_proxy.py
Bihan: Done

src/dstack/_internal/server/services/services/__init__.py
Bihan: Done

PR #3726, opened 2026-03-31
runpod: support on-demand CPU offers and provisioning

peterschmidt85:
## Summary
- Add RunPod on-demand CPU support in dstack.
- Enable `runpod-cpu` gpuhunt-flagged offers for selection.
- Use `deployCpuPod` for CPU jobs while keeping existing GPU flow unchanged.
- Apply CPU disk-cap filtering based on per-offer maximum disk.
- Add unit tests for CPU mutation generation and API dispatch.

## Testing
- pre-commit run --all-files
- pytest

Depends on https://github.com/dstackai/gpuhunt/pull/225

### DIFF
--- a/src/dstack/_internal/core/backends/runpod/compute.py
+++ b/src/dstack/_internal/core/backends/runpod/compute.py
@@ -88,7 +88,24 @@ class RunpodCompute(
         return offers
 
     def get_offers_modifiers(self, requirements: Requirements) -> Iterable[OfferModifier]:
-        return [get_offers_disk_modifier(CONFIGURABLE_DISK_SIZE, requirements)]
+        gpu_disk_modifier = get_offers_disk_modifier(CONFIGURABLE_DISK_SIZE, requirements)
+
+        def disk_modifier(
+            offer: InstanceOfferWithAvailability,
+        ) -> Optional[InstanceOfferWithAvailability]:
+            if len(offer.instance.resources.gpus) > 0:
+                return gpu_disk_modifier(offer)
+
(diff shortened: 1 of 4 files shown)


src/dstack/_internal/core/backends/runpod/compute.py
```diff
@@ -140,46 +157,61 @@ def run_job(
...
 
         instance_id = resp["id"]
 
-        # Call edit_pod to pass container_registry_auth_id.
+        # Keep the same post-create registry-auth flow for both GPU and CPU paths.
         # Expect a long time (~5m) for the pod to pick up the creds.
-        # TODO: remove editPod once createPod supports docker's username and password
-        # editPod is temporary solution to set container_registry_auth_id because createPod does not
-        # support it currently. This will be removed once createPod supports container_registry_auth_id
-        # or username and password
```
jvstme, line 182: (nit) Both of the deleted comments look useful to me.

- The `Call edit_pod to pass container_registry_auth_id` explains why we need to edit the pod.
- The `TODO` explains how we intend to fix the 5m wait and what input we need from the RunPod team.

jvstme (approved):

src/dstack/_internal/core/backends/runpod/compute.py
peterschmidt85: Restored the rationale/TODO in `compute.py` in commit 8651e9a4.\n\nWe now again document why `edit_pod` is used for registry creds and that this path should be removed once RunPod create-mutations support passing registry auth reliably.

jvstme (approved):

PR #3754, opened 2026-04-07
Show offers in dstack apply for elastic container fleets

peterschmidt85:
## Steps to reproduce
```yaml
type: fleet
name: my-fleet

# Elastic/template fleet (provision on demand)
nodes: 0..1

backends: [runpod]
spot_policy: auto
idle_duration: 10m
```

```bash
dstack apply -f fleet.dstack.yml
```

```text
Project        main
User           admin
Configuration  fleet.dstack.yml
Type           fleet
Fleet type     cloud
Nodes          min=0 target=0 max=1
Backends       runpod
Resources      cpu=2.. mem=8GB.. disk=100GB.. gpu=0..
Spot policy    auto

No matching instance offers available. Possible reasons:
https://dstack.ai/docs/guides/troubleshooting/#no-offers
```

## Expected
`dstack apply` should show matching offers for elastic container-based fleets, since runs are provisioned on demand from those offers.

This is important not only for offer visibility/convenience, but also for consistency: VM-based elastic fleets already show offers during `apply`, and container-based elastic fleets should behave the same way.

## Actual
No offers were shown for elastic container-based fleets during `apply`, even when offers existed.

## Solution
Refactored fleet offer semantics to avoid duplicating branches while preserving behavior:
- Added `get_fleet_offers(...)` as the shared offer lookup for fleet planning/provisioning.
- `get_fleet_offers(...)` defaults to filtering by `BACKENDS_WITH_CREATE_INSTANCE_SUPPORT`.
- Removed `get_create_instance_offers(...)`; create-instance paths now call `get_fleet_offers(...)` with `include_only_create_instance_supported_backends=True` explicitly.
- In `get_plan(...)`, when `nodes.target == 0`, we call `get_fleet_offers(...)` with create-instance filtering disabled, so elastic Runpod/container offers are shown.
- Removed `_is_elastic_cloud_fleet_spec(...)` helper and switched to direct `nodes.target` semantics.

This keeps non-elastic behavior unchanged while fixing elastic container fleet offer visibility in `dstack apply`.

## AI Assistance
This PR was prepared with AI assistance (Codex).

### DIFF
--- a/src/dstack/_internal/server/services/fleets.py
+++ b/src/dstack/_internal/server/services/fleets.py
@@ -443,24 +443,20 @@ async def get_plan(
     offers = []
     if effective_spec.configuration.ssh_config is None:
         requirements = get_fleet_requirements(effective_spec)
-        if _is_elastic_cloud_fleet_spec(effective_spec):
-            offers_with_backends = await offers_services.get_offers_by_requirements(
-                project=project,
-                profile=effective_spec.merged_profile,
-                requirements=requirements,
-                multinode=(
-                    effective_spec.configuration.placement == InstanceGroupPlacement.CLUSTER
-                ),
-                blocks=effective_spec.configuration.blocks,
-            )
-        else:
-            offers_with_backends = await get_create_instance_offers(
-                project=project,
-                profile=effective_spec.merged_profile,
-                requirements=requirements,
-                fleet_spec=effective_spec,
-                blocks=effective_spec.configuration.blocks,
-            )
+        nodes = effective_spec.configuration.nodes
+        include_only_create_instance_supported_backends = True
+        if nodes is not None:
+            include_only_create_instance_supported_backends = nodes.target != 0
+        offers_with_backends = await get_fleet_offers(
+            project=project,
+            profile=effective_spec.merged_profile,
+            requirements=requirements,
+            fleet_spec=effective_spec,
+            blocks=effective_spec.configuration.blocks,
+            include_only_create_instance_supported_backends=(
+                include_only_create_instance_supported_backends
+            ),
+        )
         offers = [offer for _, offer in offers_with_backends]
 
     _remove_fleet_spec_sensitive_info(effective_spec)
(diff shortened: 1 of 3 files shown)


src/dstack/_internal/server/services/fleets.py
```diff
@@ -442,13 +442,25 @@ async def get_plan(
...
+                blocks=effective_spec.configuration.blocks,
+            )
+        else:
+            offers_with_backends = await get_create_instance_offers(
+                project=project,
+                profile=effective_spec.merged_profile,
+                requirements=requirements,
+                fleet_spec=effective_spec,
+                blocks=effective_spec.configuration.blocks,
+            )
```
r4victor: Having two branches with get_offers_by_requirements and get_create_instance_offers seems too verbose here. I think it can be replaced with a get_create_instance_offers param that would enable/disable BACKENDS_WITH_CREATE_INSTANCE_SUPPORT filtering.

src/dstack/_internal/server/services/fleets.py
```diff
@@ -468,6 +480,16 @@ async def get_plan(
     return plan
 
 
+def _is_elastic_cloud_fleet_spec(fleet_spec: FleetSpec) -> bool:
```
r4victor: This function name is a misnomer since it does not really check that the fleet is elastic (nodes is a range) but that nodes.target == 0. And since target cannot be less than min, that's all we need to check for.

src/dstack/_internal/server/services/fleets.py
```diff
@@ -480,6 +488,49 @@ async def get_create_instance_offers(
     master_job_provisioning_data: Optional[JobProvisioningData] = None,
     infer_master_job_provisioning_data_from_fleet_instances: bool = True,
 ) -> List[Tuple[Backend, InstanceOfferWithAvailability]]:
+    """
```
r4victor: Do we still need get_create_instance_offers() or can we just call get_fleet_offers with include_only_create_instance_supported_backends=True?

PR #3774, opened 2026-04-14
Support --fleet in dstack offer

peterschmidt85:
### CLI
- Add `--fleet` support to `dstack offer`.
- Keep the current unscoped behavior when `--fleet` is not set.
- Scope offers to one or more selected fleets when `--fleet` is set.
- Deduplicate identical backend offers returned through multiple selected fleets while keeping matching existing instances separate.
- Support imported fleets in `dstack offer`.
- Keep `dstack apply` unchanged so it still selects a single best fleet.
- Remove the stale `--fleet` warning, add a plain-output hint, and document the behavior in `SKILL.md`.

## AI Assistance
- Assisted by Codex

### DIFF
--- a/src/dstack/_internal/server/services/runs/plan.py
+++ b/src/dstack/_internal/server/services/runs/plan.py
@@ -119,13 +119,23 @@ async def get_job_plans(
                 volumes=volumes,
                 exclude_not_available=False,
             )
-            if _should_force_non_fleet_offers(run_spec) or (
+            if _should_force_non_fleet_offers(run_spec):
+                instance_offers, backend_offers = await _get_dstack_offer_offers(
+                    session=session,
+                    project=project,
+                    profile=profile,
+                    run_spec=run_spec,
+                    job=jobs[0],
+                    volumes=volumes,
+                )
+            elif (
                 FeatureFlags.AUTOCREATED_FLEETS_ENABLED
                 and profile.fleets is None
                 and fleet_model is None
             ):
-                # Keep the old behavior returning all offers irrespective of fleets.
-                # Needed for supporting offers with autocreated fleets flow (and for `dstack offer`).
+                # Keep the old behavior returning all offers irrespective of fleets
+                # when no fleets are explicitly specified. Needed for supporting
+                # offers with autocreated fleets flow.
                 instance_offers, backend_offers = await _get_non_fleet_offers(
                     session=session,
                     project=project,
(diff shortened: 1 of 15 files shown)


src/tests/_internal/cli/utils/test_offer.py
```diff
@@ -0,0 +1,95 @@
...
+    run_spec = get_run_spec(repo_id="test-repo")
+    # Keep this helper's asyncio state isolated. `asyncio.run()` clears the current event loop
+    # on Python 3.9/3.10, which can break later tests that still expect one to exist.
+    loop = asyncio.new_event_loop()
+    try:
+        job = loop.run_until_complete(
+            get_jobs_from_run_spec(run_spec=run_spec, secrets={}, replica_num=0)
+        )[0]
+    finally:
+        loop.close()
```
r4victor, line 48: What's happening here?

src/dstack/_internal/server/services/runs/plan.py
```diff
@@ -672,6 +710,91 @@ async def _get_non_fleet_offers(
...
+            )
+        )
+    instance_offers.sort(key=lambda offer: offer[1].price or 0)
+    backend_offers = await get_backend_offers_in_run_candidate_fleets(
+        session=session,
+        project=project,
+        run_spec=run_spec,
+        job=job,
+        volumes=volumes,
+        max_offers_per_fleet=None,
```
r4victor, line 805: You need to specify non-None `max_offers_per_fleet` (currently 100 when selecting optimal fleet) – otherwise requesting offers across multiple fleets could hurt performance, especially memory usage.

src/dstack/_internal/server/services/runs/plan.py
```diff
@@ -672,6 +710,91 @@ async def _get_non_fleet_offers(
...
+    job: Job,
+    volumes: list[list[Volume]],
+) -> tuple[
+    list[tuple[InstanceModel, InstanceOfferWithAvailability]],
+    list[tuple[Backend, InstanceOfferWithAvailability]],
+]:
+    """
+    Returns existing-instance and backend offers across the run's candidate fleets.
+
+    This is used by `dstack offer` when fleets are explicitly selected. Unlike normal apply
```
r4victor: Why `dstack offer` returns existing instances?

src/tests/_internal/cli/utils/test_offer.py
peterschmidt85: 3.9 `get_event_loop()` has special policy-based behavior: it auto-creates a loop on the main thread only in a narrow case, and otherwise relies on current-loop state. ([docs.python.org](https://docs.python.org/3.9/library/asyncio-queue.html))

So the problematic sequence is:

* our test helper calls `asyncio.run(...)` that manipulates the thread’s current-loop state
* later, another 3.9 test does `asyncio.Queue()`
* `Queue()` still depends on the old current-loop machinery
* result: `RuntimeError: There is no current event loop in thread 'MainThread'`

https://github.com/dstackai/dstack/actions/runs/24390235360/job/71239917967

What changed in 3.10 is important:

* `asyncio.Queue` changed from `Queue(..., loop=None)` to just `Queue(maxsize=0)`, and the docs explicitly say: “Changed in version 3.10: Removed the loop parameter.” ([docs.python.org](https://docs.python.org/3.10/library/asyncio-queue.html))

src/dstack/_internal/server/services/runs/plan.py
peterschmidt85: Added a TODO comment here and kept this uncapped for now, since this path is expected to return exact `total_offers`. We already dedupe identical backend offers while merging, so I’d rather revisit capping only if this causes real perf issues.

PR #3781, opened 2026-04-15
Kubernetes: add support for volumes

### LINKED ISSUE #3757: [Feature]: Kubernetes: volumes support
### Problem

While Kubernetes supports a wide variety of both local and non-local (network) persistent volumes, `dstacks`'s Kubernetes backend lacks any volume support.

### Solution

https://docs.google.com/document/d/1RcY9CdeMjxmSF_w9QtzUGx0hqdFEbTaxKMR9IugasT0

### Workaround

_No response_

Yes

un-def:
Added support for `backend: kubernetes` volumes, which are mapped to `PersistentVolumeClaim`s. An existing PVC can be reused with `claim_name` field, otherwise a new PVC is created, similarly to `volume_id` field logic in other backends.

`VolumeConfiguration` is now a tagged union with backend-specific subclasses.

Kubernetes backend `DUMMY_REGION` "-" was changed to "" (an empty string), which was retrofitted to indicate "region is not applicable (for this backend)". CLI now hides corresponding fields if region = "".

Closes: https://github.com/dstackai/dstack/issues/3757

### DIFF
--- a/src/dstack/_internal/core/backends/kubernetes/compute.py
+++ b/src/dstack/_internal/core/backends/kubernetes/compute.py
@@ -3,6 +3,7 @@ import shlex
 import subprocess
 import tempfile
 import time
+from decimal import Decimal
 from enum import Enum
 from typing import List, Optional
 
@@ -17,11 +18,14 @@ from dstack._internal.core.backends.base.compute import (
     ComputeWithInstanceVolumesSupport,
     ComputeWithMultinodeSupport,
     ComputeWithPrivilegedSupport,
+    ComputeWithVolumeSupport,
     generate_unique_gateway_instance_name,
     generate_unique_instance_name_for_job,
     generate_unique_name,
+    generate_unique_volume_name,
     get_docker_commands,
     get_dstack_gateway_commands,
+    merge_tags,
 )
 from dstack._internal.core.backends.kubernetes.models import (
     KubernetesConfig,
@@ -32,13 +36,15 @@ from dstack._internal.core.backends.kubernetes.resources import (
     AMD_GPU_NAME_TO_DEVICE_IDS,
     AMD_GPU_NODE_TAINT,
     AMD_GPU_RESOURCE,
-    DUMMY_REGION,
     NVIDIA_GPU_NAME_TO_GPU_INFO,
     NVIDIA_GPU_NODE_TAINT,
     NVIDIA_GPU_PRODUCT_LABEL,
     NVIDIA_GPU_RESOURCE,
+    OBJECT_NAME_MAX_LENGTH,
     PodPhase,
     TaintEffect,
+    filter_invalid_labels,
+    format_dstack_label_key,
     format_memory,
     get_amd_gpu_from_node_labels,
     get_gpu_request_from_gpu_spec,
@@ -49,6 +55,7 @@ from dstack._internal.core.backends.kubernetes.resources import (
     get_nvidia_gpu_from_node_labels,
     is_hard_taint,
     is_taint_tolerated,
+    parse_quantity,
 )
 from dstack._internal.core.backends.kubernetes.utils import (
     call_api_method,
@@ -56,6 +63,7 @@ from dstack._internal.core.backends.kubernetes.utils import (
 )
 from dstack._internal.core.consts import DSTACK_RUNNER_SSH_PORT
 from dstack._internal.core.errors import ComputeError, ProvisioningError
+from dstack._internal.core.models.backends.base import BackendType
 from dstack._internal.core.models.common import CoreModel
 from dstack._internal.core.models.gateways import (
     GatewayComputeConfiguration,
@@ -70,7 +78,13 @@ from dstack._internal.core.models.placement import PlacementGroup
 from dstack._internal.core.models.resources import CPUSpec, GPUSpec
 from dstack._internal.core.models.routers import AnyGatewayRouterConfig
 from dstack._internal.core.models.runs import Job, JobProvisioningData, Requirements, Run
-from dstack._internal.core.models.volumes import InstanceMountPoint, Volume
+from dstack._internal.core.models.volumes import (
+    InstanceMountPoint,
+    KubernetesVolumeConfiguration,
+    Volume,
+    VolumeMountPoint,
+    VolumeProvisioningData,
+)
 from dstack._internal.utils.common import get_or_error
 from dstack._internal.utils.logging import get_logger
 
@@ -100,6 +114,7 @@ class KubernetesCompute(
     ComputeWithFilteredOffersCached,
     ComputeWithPrivilegedSupport,
     ComputeWithInstanceVolumesSupport,
+    ComputeWithVolumeSupport,
     ComputeWithGatewaySupport,
     ComputeWithMultinodeSupport,
     Compute,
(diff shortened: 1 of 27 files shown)


src/dstack/_internal/core/models/configurations.py
```diff
@@ -1114,23 +1120,47 @@ class ApplyConfigurationType(str, Enum):
...
-    VolumeConfiguration,
+    AnyVolumeConfiguration,
 ]
 
 
-class ApplyConfiguration(CoreModel):
+class BaseApplyConfiguration(CoreModel):
+    """
+    `BaseApplyConfiguration` parses the configuration based on the `type` discriminator field,
+    but futher dispatching (reparsing) may be required if there is another discriminator field,
```
r4victor:
```suggestion
    but further dispatching (reparsing) may be required if there is another discriminator field,
```

src/dstack/_internal/core/models/configurations.py
```diff
@@ -1114,23 +1120,47 @@ class ApplyConfigurationType(str, Enum):
...
+    """
+
     __root__: Annotated[
-        AnyApplyConfiguration,
+        Union[
+            # Final configurations
+            AnyRunConfiguration,
+            FleetConfiguration,
+            GatewayConfiguration,
+            # Base configurations (futher parsing required to get a concrete AnyApplyConfiguration)
```
r4victor:
```suggestion
            # Base configurations (further parsing required to get a concrete AnyApplyConfiguration)
```

src/dstack/_internal/core/models/configurations.py
```diff
@@ -1114,23 +1120,47 @@ class ApplyConfigurationType(str, Enum):
...
+        if not isinstance(conf, BaseVolumeConfiguration):
+            # If it's a final configuration (currently, any configuration other than
+            # BaseVolumeConfiguration), parse again rejecting extra fields
+            # for validation purposes only and return the final configuration
+            _ = BaseApplyConfiguration.parse_obj(data).__root__
+            return conf
     except ValidationError as e:
         raise ConfigurationError(e)
-    return conf
+    # Otherwise, delegate futher parsing to more specific parser
```
r4victor:
```suggestion
    # Otherwise, delegate further parsing to more specific parser
```

src/dstack/_internal/core/backends/kubernetes/compute.py
```diff
@@ -506,6 +553,102 @@ def terminate_gateway(
...
+        )
+        logger.debug("Created PVC %s for volume %s", pvc_name, volume.name)
+
+        return VolumeProvisioningData(
+            backend=BackendType.KUBERNETES,
+            volume_id=pvc_name,
+            size_gb=volume.configuration.size_gb,
+            attachable=False,
+            detachable=False,
+        )
```
r4victor, line 634: Since the volume is not attachable, VolumeAttachmentModel is not created for it and it's not tracked that the volume is in use by the instance. This means a volume can be deleted while in use. This is not a k8s problem but rather a problem introduced when implementing Runpod volumes and now k8s inherits it. Not sure if needs a fix now, but please give it a check.

Tested that the issue does exists for Runpod. Delete volume request succeeds but the deletion itself raises and error:

[{'message': 'You must remove this network volume from all pods before
                    deleting it.', 'path': ['deleteNetworkVolume'], 'extensions': {'code':
                    'RUNPOD'}}]

src/dstack/_internal/core/models/configurations.py
```diff
@@ -1114,23 +1120,47 @@ class ApplyConfigurationType(str, Enum):
...
+    return parse_volume_configuration(data)
 
 
 AnyDstackConfiguration = AnyApplyConfiguration
```
r4victor: DstackConfiguration now produces the following schema (note "volume": "#/definitions/KubernetesVolumeConfigurationRequest" in discriminator mapping). Is this intended? Any consequences?

```
{
  "title": "DstackConfigurationRequest",
  "discriminator": {
    "propertyName": "type",
    "mapping": {
      "dev-environment": "#/definitions/DevEnvironmentConfigurationRequest",
      "task": "#/definitions/TaskConfigurationRequest",
      "service": "#/definitions/ServiceConfigurationRequest",
      "fleet": "#/definitions/FleetConfigurationRequest",
      "gateway": "#/definitions/GatewayConfigurationRequest",
      "volume": "#/definitions/KubernetesVolumeConfigurationRequest"
    }
  },
...
```

un-def: Good catch! Fixed: https://github.com/dstackai/dstack/pull/3781/changes/d97a8d99e72b92d9486980f37ac780a46a3dbdcc

```json
{
  "title": "DstackConfigurationRequest",
  "discriminator": {
    "propertyName": "type",
    "mapping": {
      "dev-environment": "#/definitions/DevEnvironmentConfigurationRequest",
      "task": "#/definitions/TaskConfigurationRequest",
      "service": "#/definitions/ServiceConfigurationRequest",
      "fleet": "#/definitions/FleetConfigurationRequest",
      "gateway": "#/definitions/GatewayConfigurationRequest",
      "volume": "#/definitions/VolumeConfigurationRequest"
    }
  }
}
```

```json
{
  "title": "VolumeConfigurationRequest",
  "discriminator": {
    "propertyName": "backend",
    "mapping": {
      "aws": "#/definitions/AWSVolumeConfigurationRequest",
      "gcp": "#/definitions/GCPVolumeConfigurationRequest",
      "runpod": "#/definitions/RunpodVolumeConfigurationRequest",
      "kubernetes": "#/definitions/KubernetesVolumeConfigurationRequest"
    }
  }
}
```

src/dstack/_internal/core/backends/kubernetes/compute.py
un-def: Not sure how we can fix it since we don't track if the volume is used. Kubernetes allows to delete PVC in use, but it won't be deleted until it's released:

> If a user deletes a PVC in active use by a Pod, the PVC is not removed immediately. PVC removal is postponed until the PVC is no longer actively used by any Pods. Also, if an admin deletes a PV that is bound to a PVC, the PV is not removed immediately. PV removal is postponed until the PV is no longer bound to a PVC.

As a `dstack` user, I can delete a volume which is used by a run, but the run won't be interrupted (which is probably good), and the PVC/PV won't be deleted until the job pod is terminated, with possible unexpected charges for storage (which is bad).

It seems that the only (kubernetes-specific) solution is to iterate over all active pods checking if any of them uses the PVC in question.

un-def: Here is the patch, but it (arguably) makes things even worse:

```patch
diff --git a/src/dstack/_internal/core/backends/kubernetes/compute.py b/src/dstack/_internal/core/backends/kubernetes/compute.py
index 4ea833bd0..9ab903c78 100644
--- a/src/dstack/_internal/core/backends/kubernetes/compute.py
+++ b/src/dstack/_internal/core/backends/kubernetes/compute.py
@@ -638,6 +638,25 @@ class KubernetesCompute(
         pvc_name = volume.volume_id
         assert pvc_name is not None

+        pvc_in_use = False
+        for pod in self.api.list_namespaced_pod(self.config.namespace).items:
+            try:
+                pod_phase = PodPhase(get_or_error(get_or_error(pod.status).phase))
+                pod_volumes = get_or_error(pod.spec).volumes or []
+            except ValueError:
+                continue
+            if pod_phase.is_finished():
+                continue
+            if any(
+                v.persistent_volume_claim is not None
+                and v.persistent_volume_claim.claim_name == pvc_name
+                for v in pod_volumes
+            ):
+                pvc_in_use = True
+                break
+        if pvc_in_use:
+            raise ComputeError(f"Failed to delete volume in use: {volume.name}")
+
         pvc = call_api_method(
             self.api.delete_namespaced_persistent_volume_claim,
             expected=404,
```

From the user's point of view, the volume is still successfully removed, only the server admin knows that something went wrong (if they checks the logs):

```
DEBUG    dstack._internal.server.background.pipeline_tasks.base:357 Processing volumes item 667396e3-6d13-4e09-87a0-3c8f36d0c85e
ERROR    dstack._internal.server.background.pipeline_tasks.volumes:408 Got exception when deleting volume volume-kubernetes. Please
                    terminate it manually to avoid unexpected charges.
<...>
                    dstack._internal.core.errors.ComputeError: Failed to delete volume in use: volume-kubernetes
```

Without the patch, the PVC (and PV if reclaimPolicy = Delete) will  be deleted eventually by Kubernetes itself.

r4victor (approved):

src/dstack/_internal/core/backends/kubernetes/compute.py
un-def: https://github.com/dstackai/dstack/issues/3789

PR #3798, opened 2026-04-20
[Docs] Update examples

peterschmidt85:
- [x] Update `Services`, `SGLang`, `vLLM`, `NIM`, and `TensorRT-LLM` examples (migrate to the new models)
- [x] Update `TensorRT-LLM` example (replace the legacy DeepSeek/Triton flow with an inline Qwen3 example)
- [x] Remove `Models` section with `Wan2.2` (logic here is its' better to have up-to-date `Inference` examples than out-dated and incomplete `Models` section)
- [x] Remove deprecated example, incl. `Llama 3.1`, `Llama 3.2`, `Deepseek`, and other examples that we previously unlisted
- [x] Rely on "inline" examples only (logic here is to keep the minimum of maintainable examples, and do not aim for full reproducibility - this requires too much effort to maintain and gets outdated rather quickly) - **WIP**
- [ ] Get rid out of "examples generation"; move all examples from `/examples` to `docs` and remove the corresponding hook (logic here is to do not complicate examples - see above)
- [ ] **MAJOR** Update `TRL` examples under `Single-node training`
- [ ] **MAJOR** Update `AMD` example
- [ ] **MAJOR** Update `TRL` examples under `Distributed training`
- [ ] Update `TPU` example


peterschmidt85:
@Bihan please review the complete changes; if all is correct, I'll merge it and move the rest of the mentioned items above into a separate issue/PR. Plus share any feed back on the not-complete items above if you agree/disagree.

examples/inference/sglang/README.md
```diff
@@ -1,190 +1,226 @@
...
 
     ```yaml
     type: service
-    name: deepseek-r1
+    name: qwen397
 
-    image: lmsysorg/sglang:latest
-    env:
-      - MODEL_ID=deepseek-ai/DeepSeek-R1-Distill-Llama-8B
+    image: lmsysorg/sglang:dev
```
Bihan: Why `lmsysorg/sglang:dev` instead of `lmsysorg/sglang:latest`?

examples/inference/vllm/README.md
```diff
@@ -1,122 +1,104 @@
...
+    image: vllm/vllm-openai
 
-<div class="termy">
+    commands:
+      - |
+        uv pip install --system -U vllm --torch-backend=auto \
+          --extra-index-url https://wheels.vllm.ai/nightly
+        uv pip install --system -U --pre flashinfer-jit-cache \
+          --index-url https://flashinfer.ai/whl/nightly/cu129
+        export VLLM_DEEP_GEMM_WARMUP=relax
```
Bihan: Why does this example install `nightly vLLM` and `flashinfer` on top of `vllm/vllm-openai`, when the vLLM guide serves the same model with just the image? Is that required for the FP8 variant

https://docs.vllm.ai/projects/recipes/en/latest/Qwen/Qwen3.5.html#nvidia_1

PR #3804, opened 2026-04-21
[Internal] Improve warning message in `ServerConfigManager.apply_config()`

un-def:
* Include project name
* Include action -- create or update

### DIFF
--- a/src/dstack/_internal/server/services/config.py
+++ b/src/dstack/_internal/server/services/config.py
@@ -17,7 +17,7 @@ from dstack._internal.core.errors import (
     ServerClientError,
 )
 from dstack._internal.core.models.backends.base import BackendType
-from dstack._internal.core.models.common import CoreModel
+from dstack._internal.core.models.common import ApplyAction, CoreModel
 from dstack._internal.server import settings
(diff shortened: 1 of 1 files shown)


src/dstack/_internal/server/services/config.py
```diff
@@ -164,20 +164,25 @@ async def _apply_project_config(
                 comparable_backend_config = current_source_backend_config or current_backend_config
                 if backend_config == comparable_backend_config:
                     continue
+            # current_backend_config may be None if backend exists
+            # but it's config is invalid (e.g. cannot be decrypted).
+            # Update backend in this case.
+            if current_backend_config is None and not backend_exists:
+                apply_action = ApplyAction.CREATE
```
r4victor: `ApplyAction` is used in the context of `dstack apply`. Server apply config is a different thing so I'd not bring it here although it'll work – using strings would be better.

un-def: https://github.com/dstackai/dstack/pull/3804/changes/fcf9303df6217cccf85c91e2c89cc19cc2760d7f

r4victor (approved):

PR #3845, opened 2026-05-04
Gateway exports

jvstme:
Allow exporting gateways to other projects.

```
$ dstack export --project main
 NAME       FLEETS   GATEWAYS      IMPORTERS 
 my-export  -        main-gateway  team      

$ dstack import --project team
 NAME            FLEETS   GATEWAYS     
 main/my-export  -        main-gateway 

$ dstack gateway --project team
 NAME               BACKEND          HOSTNAME        DOMAIN                 DEFAULT  STATUS  
 main/main-gateway  aws (eu-west-1)  108.131.126.35  gtw.mycompany.example           running
```

Imported gateways can be used by specifying them in the `gateway` property in service configurations.

```yaml
type: service
port: 80
image: nginx
gateway: main/main-gateway
 ```
 
 Limitations:
 - Setting an imported gateway as the default gateway is not yet supported.
 - Evicting services from a gateway that is no longer imported is not yet supported.
 
 #3851

### CODE REVIEW

**Comment by jvstme** [comment]
@r4victor, sorry for the huge PR. It's mostly boilerplate, but at least `server/services` and `server/security` are worth reviewing

File: src/dstack/_internal/server/routers/gateways.py [hunk 1]
```diff
@@ -28,22 +31,32 @@
...
+            include_imported=body.include_imported,
+        )
     )
 
 
 @router.post("/get", response_model=models.Gateway)
 async def get_gateway(
     body: schemas.GetGatewayRequest,
     session: AsyncSession = Depends(get_session),
-    user_project: Tuple[UserModel, ProjectModel] = Depends(ProjectMemberOrPublicAccess()),
```
r4victor, line 44: It seems `ProjectMemberOrPublicAccess()` was used so that everyone could get gateways in public projects. Is the tightening intentional? Does the UI for public projects continue to work without problems?

r4victor (approved):

src/dstack/_internal/server/routers/gateways.py
jvstme: Wasn't intentional, thanks

PR #3848, opened 2026-05-04
Add server-side validation for fleet configuration subtypes

### LINKED ISSUE #2223: Forbid mixing cloud and SSH fleet configuration properties
### Problem

It is not always clear which fleet configuration properties are relevant for cloud fleets and which are relevant for SSH fleets.
This leads users to compose configurations with both cloud and SSH properties.

```yaml
type: fleet

# SSH properties
ssh_config:
  user: "user"
  identity_file: /home/user/.ssh/dstack
  hosts:
    - 127.0.0.1
  port: 2261

# Cloud properties
resources:
  cpu: 7
  memory: 10GB
  disk: 15GB
```
In that case `dstack` provisions an SSH fleet and ignores the cloud properties silently, which is highly likely to cause misunderstandings. [Example](https://github.com/dstackai/dstack/issues/1994#issuecomment-2514670334):
> it completely ignored declared resource request with CPU 7, 10GB RAM and 15GB Volume

### Solution

Validate fleet configurations and show errors if cloud and SSH properties are mixed.

### Implementation note

This can be implemented by adding validators on top of the existing configuration structure or by rethinking the configuration structure, e.g. moving cloud and SSH properties to different mappings or even introducing different configuration (sub)types for cloud and SSH fleets.

un-def:
Forbids mixing cloud and SSH fleet configuration properties. Default values (mostly None) are ignored for compatibility.

Closes: https://github.com/dstackai/dstack/issues/2223


un-def:
image

---

image

jvstme (approved):

PR #3856, opened 2026-05-06
Add dstack-dev skill for using dev environments

peterschmidt85:
## Summary
- Adds a new `dstack-dev` skill covering the workflow for using dstack dev environments to experiment with, tune, and debug workloads before promoting them to tasks or services.
- Clarifies the existing `dstack` skill's description so the two skills have non-overlapping scopes.


peterschmidt85:
Watch #4012

PR #3868, opened 2026-05-08
Support nvidia dynamo

Bihan:
Service Configuration example 

```
type: service
name: dynamo-pd


env:
  - HF_TOKEN

replicas: 
  - name: router
    count: 1
    docker: true
    router:
      type: dynamo
    commands:
      # DIND ships docker but not pip — set up a venv for ai-dynamo.
      - apt-get update
      - apt-get install -y python3-dev python3-venv
      - python3 -m venv ~/dyn-venv
      - source ~/dyn-venv/bin/activate
      - pip install -U pip
      - pip install --pre "ai-dynamo[sglang]"
      # Pull Dynamo and start the supporting compose stack (NATS, etcd, ...).
      - git clone https://github.com/ai-dynamo/dynamo.git
      - docker compose -f dynamo/deploy/docker-compose.yml up -d
      # Run the Dynamo frontend (was the commented dev command).
      - |
        python3 -m dynamo.frontend \
          --http-host 0.0.0.0 --http-port 8000 \
          --discovery-backend etcd --router-mode kv \
          --kv-cache-block-size 64
    resources:
      cpu: 4

  # ── prefill worker ─────────────────────────────────────────────────────
  # dstackai/base + Python 3.12 + NVCC (needed for CUDA kernels in sglang).
  - name: prefill
    count: 1..2
    scaling:
      metric: rps
      target: 4 
    python: "3.12"
    nvcc: true
    commands:
      # dstack injected DSTACK_ROUTER_INTERNAL_IP after the router replica
      # was provisioned. Compose the etcd/NATS endpoints from it.
      - export ETCD_ENDPOINTS="http://$DSTACK_ROUTER_INTERNAL_IP:2379"
      - export NATS_SERVER="nats://$DSTACK_ROUTER_INTERNAL_IP:4222"
      - export DYN_SYSTEM_HOST="0.0.0.0"
      - export DYN_SYSTEM_PORT="8000"
      # Wait until the router's etcd and NATS ports are actually accepting
      # connections — having the IP isn't the same as having the services up.
      - |
        until (echo > /dev/tcp/$DSTACK_ROUTER_INTERNAL_IP/2379) 2>/dev/null \
           && (echo > /dev/tcp/$DSTACK_ROUTER_INTERNAL_IP/4222) 2>/dev/null; do
          echo "waiting for etcd/NATS on $DSTACK_ROUTER_INTERNAL_IP..."; sleep 3
        done
      - pip install --pre "ai-dynamo[sglang]"
      - |
        python3 -m dynamo.sglang \
          --model-path $MODEL_ID --served-model-name $MODEL_ID \
          --discovery-backend etcd --host 0.0.0.0 \
          --page-size 64 \
          --disaggregation-mode prefill --disaggregation-transfer-backend nixl
    resources:
      gpu: L4

  # ── decode worker ──────────────────────────────────────────────────────
  - name: decode
    count: 1
    python: "3.12"
    nvcc: true
    commands:
      - export ETCD_ENDPOINTS="http://$DSTACK_ROUTER_INTERNAL_IP:2379"
      - export NATS_SERVER="nats://$DSTACK_ROUTER_INTERNAL_IP:4222"
      - export DYN_SYSTEM_HOST="0.0.0.0"
      - export DYN_SYSTEM_PORT="8000"
      - |
        until (echo > /dev/tcp/$DSTACK_ROUTER_INTERNAL_IP/2379) 2>/dev/null \
           && (echo > /dev/tcp/$DSTACK_ROUTER_INTERNAL_IP/4222) 2>/dev/null; do
          echo "waiting for etcd/NATS on $DSTACK_ROUTER_INTERNAL_IP..."; sleep 3
        done
      - pip install --pre "ai-dynamo[sglang]"
      - |
        python3 -m dynamo.sglang \
          --model-path $MODEL_ID --served-model-name $MODEL_ID \
          --discovery-backend etcd --host 0.0.0.0 \
          --page-size 64 \
          --disaggregation-mode decode --disaggregation-transfer-backend nixl
    resources:
      gpu: L4

port: 8000
model: meta-llama/Llama-3.2-3B-Instruct

probes:
 - type: http
   url: /health
   interval: 15s
```


src/dstack/_internal/server/services/runs/spec.py
jvstme, line 1: These in-place update cases are allowed, but will result in a broken service:
- Changing top-level service properties, resulting in a redeployment of the router replica (`image`, `env`, etc).
- Changing `router.type` from `sglang` to `dynamo`.

src/dstack/_internal/server/services/runs/replicas.py
```diff
@@ -1,14 +1,33 @@
...
+#
+#   ROUTER_FAILED          — router job has reached a terminal state
+#                            (TERMINATING/TERMINATED/FAILED/ABORTED/DONE).
+#                            The condition is permanent; the caller should
+#                            stop deferring and terminate this worker with a
+#                            clear reason — waiting longer cannot recover the
+#                            run because the router will not come back with a
+#                            fresh internal_ip.
+ROUTER_NOT_PROVISIONED: Dict[str, str] = {}
+ROUTER_FAILED: Dict[str, str] = {}
```
jvstme: (nit) Using empty dicts as sentinel objects looks a bit error-prone — if the caller uses `==` instead of `is`, the code will appear to be correct and will pass type checks, but won't work as expected.

I'd suggest to define an enum with these two variants instead.

src/dstack/_internal/server/services/runs/replicas.py
```diff
@@ -124,3 +143,95 @@ def has_out_of_date_replicas(run: RunModel, group_filter: Optional[str] = None)
...
+    cfg = run_spec.configuration
+    if not isinstance(cfg, ServiceConfiguration):
+        return None
+    global_replica_num = 0
+    for group in cfg.replica_groups:
+        if group.router is not None:
+            return global_replica_num
+        assert group.count.min is not None
+        global_replica_num += group.count.min
+    return None
```
jvstme: I believe this function will start to return incorrect values after an in-place place. For example, if an in-place update changes `group.count.min` of a worker replica group that precedes the router replica group.

Some alternative ways to determine which jobs to load in `_fetch_run_model`:
- If the run spec includes a dynamo router, load all active jobs (but not terminated, because there can be too many). And if the router job isn't loaded by that query, then I think we can assume that it is terminated (or that the group is scaled to zero, once that is allowed).
- Duplicate the replica group name as a `JobModel` field and use it as a filter in the database query.

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
```diff
@@ -477,6 +491,54 @@ async def _prepare_startup_context(
...
+                ),
+            )
+            return None
+        logger.debug(
+            "%s: waiting for router replica to be provisioned",
+            fmt(context.job_model),
+        )
+        return None
+    if router_env:
+        context.job.job_spec.env.update(router_env)
```
jvstme: (nit) Mutating `job_spec.env` looks a bit hacky. Now `job_spec.env` contains just the user-provided environment in some cases, and one extra system variable in other cases, which makes the code a bit less straightforward.

I would suggest to return `router_env` from this function as a `_StartupContext` field, then pass it through the function calls until it reaches `RunnerClient.submit_job`, where it can be merged with other env variables.

(`RunnerClient.submit_job` uses [the same hack](https://github.com/dstackai/dstack/blob/565b9a625b9408e5e1c9c1b4c394b1102eb98d9a/src/dstack/_internal/server/services/runner/client.py#L108) when merging variables, but it creates a copy of the job spec, so the spec mutation doesn't leak to the caller)

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
```diff
@@ -560,14 +625,39 @@ async def _fetch_run_model(
         .options(joinedload(RunModel.fleet).load_only(FleetModel.id, FleetModel.name))
     )
     if replica_num is not None:
+        # Pre-fetch the bare run_spec to discover whether the run has a
+        # router replica group, and if so at which replica_num. The query
+        # below then includes both this replica AND the router replica
+        # For runs without a router group (services without one, plus
+        # all tasks and dev-environments), the helper returns None and we
+        # fall through to the original single-replica behavior.
+        spec_res = await session.execute(select(RunModel.run_spec).where(RunModel.id == run_id))
```
jvstme: (nit) It would be great to avoid this extra query. For example, by loading the run spec as a join in `_refetch_locked_job_model`

src/dstack/_internal/server/services/runs/replicas.py
```diff
@@ -124,3 +143,95 @@ def has_out_of_date_replicas(run: RunModel, group_filter: Optional[str] = None)
...
+    if router_group.router is None or router_group.router.type != RouterType.DYNAMO:
+        return None
+    if job_belongs_to_group(job_model, router_group.name):
+        # Router replica itself doesn't need to be told its own IP.
+        return None
+
+    router_job = find_router_job(run_model, router_group.name)
+    if router_job is None:
+        # No router job yet — the run was just submitted and jobs haven't
+        # been materialized. Treat as "not provisioned" so the caller defers.
```
jvstme: (nit) Is this possible? I think at run submission we create the jobs in the same database transaction as the run, see [`submit_run()`](https://github.com/dstackai/dstack/blob/565b9a625b9408e5e1c9c1b4c394b1102eb98d9a/src/dstack/_internal/server/services/runs/__init__.py#L609)

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
jvstme, line 1: A few unit tests for these changes could be useful

src/dstack/_internal/server/services/runs/spec.py
Bihan: Done

src/dstack/_internal/server/services/runs/replicas.py
Bihan: Done

src/dstack/_internal/server/services/runs/replicas.py
Bihan: Done

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
Bihan: Done

src/dstack/_internal/server/services/runs/replicas.py
Bihan: Thanks for pointing out. The scenario cannot happen. I have updated the comment and return status as below.
```
if router_job is None:
        # The router's latest submission is in a terminal state and was
        # filtered out by _fetch_run_model's not-terminated predicate.
        return RouterEnvStatus.FAILED
```

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
Bihan: Done

src/dstack/_internal/server/services/runs/replicas.py
```diff
@@ -39,6 +23,31 @@ class GroupRolloutState:
     registered_non_terminating_replica_count: int
 
 
+class RouterEnvStatus(str, Enum):
+    """Outcomes returned from get_router_env_for_job() when no env dict is
+    appropriate. Each value carries a distinct caller-side action.
+
+    Using an enum (rather than empty-dict sentinels) means callers can rely
+    on either `is` or `==` to compare — both yield correct, unambiguous
+    results — and stray dicts from elsewhere can never accidentally match.
```
jvstme, line 32: (nit) I think this comparison between enums and empty dicts won't make much sense to a reader who hasn't seen the previous implementation with empty dicts.

Overall, many comments and docstrings in the PR look a bit too verbose and redundant to me. I'd only keep minimal comments that add important context which is not otherwise clear from the naming or the implementation

src/tests/_internal/core/models/test_run_spec_validators.py
jvstme, line 1: (nit) We prefer the `src/tests` directory structure to mirror `src/dstack`. I'd put these tests in `test_runs.py`

src/dstack/_internal/server/services/runs/spec.py
jvstme: > Changing top-level service properties, resulting in a redeployment of the router replica (`image`, `env`, etc).

Also relevant to [spec fields](https://github.com/Bihan/dstack/blob/3bfd45caf163225a6c2591ee6de90e7048e821ff/src/dstack/_internal/server/services/runs/spec.py#L26). For example, changing the contents of a file referenced in `files` will still trigger a rolling deployment

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
jvstme: Sorry, I was referring to the changes in this file, the pipeline, that aren't yet covered. Pipelines are important to cover, because a broken pipeline can potentially affect many jobs on the server

jvstme (approved):
Looks good overall.

Some comments (especially [1](https://github.com/dstackai/dstack/pull/3868#discussion_r3219854207) and [2](https://github.com/dstackai/dstack/pull/3868#discussion_r3222354359)) may still be worth addressing, although there doesn't seem to be anything critical.

jvstme (approved):

PR #3875, opened 2026-05-12
Add JarvisLabs backend

peterschmidt85:
Adds JarvisLabs as a dstack backend.

Implementation notes:
- Adds backend registration, config models, configurator, API client, compute implementation, docs, and backend tests.
- Uses the JarvisLabs provider from gpuhunt for offer selection. This branch depends on https://github.com/dstackai/gpuhunt/pull/231 until the provider is released.
- Supports JarvisLabs VM workloads only. GPU VMs and CPU VMs use separate JarvisLabs create/destroy APIs.
- Supports GPU spot by passing the selected offer's spot flag to JarvisLabs GPU VM creation. CPU spot is not emitted by gpuhunt and is not supported.
- Does not select a JarvisLabs template or custom image; provisioning uses the provider default VM image.
- Validates configured regions against gpuhunt's JarvisLabs supported-region map and fails closed if an unsupported region reaches a regional API call.
- Registers the project SSH key in JarvisLabs before creating an instance.
- Starts the dstack shim over SSH and persists hostname only after shim startup succeeds, so provisioning can retry after a server restart.
- Maps immediate and delayed JarvisLabs create capacity failures to `NoCapacityError` and destroys any failed machine id returned by JarvisLabs before retrying another offer. Non-capacity failed create status raises `ProvisioningError`. After a VM is running, interruption/unreachability is handled by the generic VM health path, as with other VM backends.
- Wraps JarvisLabs request failures and malformed success responses as `BackendError` instead of leaking raw transport/JSON exceptions.

E2E validation:
- CPU on-demand task provisioned and completed on JarvisLabs.
- L4 GPU on-demand task provisioned and completed CUDA tensor matmul on the GPU.
- H100 GPU spot task provisioned with JarvisLabs `is_spot: true` and completed CUDA tensor matmul on the GPU.
- Requested 120GB/200GB disks were visible inside containers in the live disk checks.
- Server restart was tested while JarvisLabs runs were active; provisioning resumed instead of losing the run.
- L4 spot no-capacity was observed from JarvisLabs and handled as a capacity failure.

Added tests cover config validation, API payloads, API error normalization, spot flag propagation, region failure behavior, capacity-failure mapping and cleanup, CPU/GPU provisioning data, disk sizing, SSH username parsing, termination routing, and restart-safe hostname persistence.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "python-multipart>=0.0.16",
     "filelock",
     "psutil",
-    "gpuhunt @ git+https://github.com/dstackai/gpuhunt.git@jarvislabs",
+    "gpuhunt==0.1.22",
     "argcomplete>=3.5.0",
     "ignore-python>=0.2.0",
     "orjson",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,9 +67,6 @@ artifacts = [
     "src/dstack/_internal/server/statics/**",
 ]
 
-[tool.hatch.metadata]
-allow-direct-references = true
-
 [tool.hatch.metadata.hooks.fancy-pypi-readme]
 content-type = "text/markdown"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "python-multipart>=0.0.16",
     "filelock",
     "psutil",
-    "gpuhunt==0.1.21",
+    "gpuhunt @ git+https://github.com/dstackai/gpuhunt.git@jarvislabs",
     "argcomplete>=3.5.0",
     "ignore-python>=0.2.0",
     "orjson",
@@ -66,6 +66,9 @@ artifacts = [
 artifacts = [
     "src/dstack/_internal/server/statics/**",
 ]
+
+[tool.hatch.metadata]
+allow-direct-references = true
 
 [tool.hatch.metadata.hooks.fancy-pypi-readme]
 content-type = "text/markdown"
--- a/src/dstack/_internal/core/backends/jarvislabs/compute.py
+++ b/src/dstack/_internal/core/backends/jarvislabs/compute.py
@@ -1,0 +1,364 @@
+import shlex
+import subprocess
+import tempfile
+import time
+from typing import List, Optional
+
+import gpuhunt
+
+from dstack._internal.core.backends.base.backend import Compute
+from dstack._internal.core.backends.base.compute import (
+    ComputeWithCreateInstanceSupport,
+    ComputeWithFilteredOffersCached,
+    ComputeWithInstanceVolumesSupport,
+    ComputeWithPrivilegedSupport,
+    generate_unique_instance_name,
+    get_shim_commands,
+)
+from dstack._internal.core.backends.base.offers import get_catalog_offers
+from dstack._internal.core.backends.jarvislabs.api_client import JarvisLabsAPIClient
+from dstack._internal.core.backends.jarvislabs.models import JarvisLabsConfig
+from dstack._internal.core.errors import BackendError, NoCapacityError, ProvisioningError
+from dstack._internal.core.models.backends.base import BackendType
+from dstack._internal.core.models.instances import (
+    InstanceAvailability,
+    InstanceConfiguration,
+    InstanceOfferWithAvailability,
+)
+from dstack._internal.core.models.placement import PlacementGroup
+from dstack._internal.core.models.runs import JobProvisioningData, Requirements
+from dstack._internal.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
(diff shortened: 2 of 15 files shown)


pyproject.toml
jvstme, line 1: (reminder) Revert and upgrade gpuhunt before merging

src/dstack/_internal/core/backends/jarvislabs/compute.py
```diff
@@ -0,0 +1,364 @@
...
+            try:
+                from gpuhunt.providers.jarvislabs import JarvisLabsProvider
+            except ImportError as e:
+                raise BackendError(
+                    "JarvisLabs backend requires gpuhunt with JarvisLabs provider support"
+                ) from e
+            catalog = gpuhunt.Catalog(balance_resources=False, auto_reload=False)
+            catalog.add_provider(JarvisLabsProvider(api_key=self.config.creds.api_key))
+            self._catalog = catalog
+        return self._catalog
```
jvstme: (nit)

gpuhunt is guaranteed to support Jarvis Labs, because:
- The correct gpuhunt version is pinned in `pyproject.toml`.
- This PR has other imports from `gpuhunt.providers.jarvislabs` that are placed at the top of the module and performed at import time.

So it's safe to remove this method and import `JarvisLabsProvider` at the top level.

src/dstack/_internal/core/backends/jarvislabs/compute.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def _get_jarvislabs_gpu_type(instance_offer: InstanceOfferWithAvailability) -> str:
+    gpu = instance_offer.instance.resources.gpus[0]
+    memory_gb = round(gpu.memory_mib / 1024)
+    if gpu.name == "A100" and memory_gb == 80:
+        return "A100-80GB"
+    return gpu.name
```
jvstme, line 183: To my understanding, this function is supposed to revert the effect of [`_gpu_name_and_memory`](https://github.com/dstackai/gpuhunt/blob/d00bc6431410276de6baf1ca2929f5962ed1881e/src/gpuhunt/providers/jarvislabs.py#L243) from gpuhunt, but it doesn't, because `_gpu_name_and_memory` also cuts spaces.

This will result in unexpected behavior if Jarvis Labs ever adds GPUs with a space in the name (e.g., `H100 NVL`). Such GPUs will be shown in dstack offers, but will always fail to provision (because we'll be passing an incorrect GPU type to the API, e.g., `H100NVL`).

Some solutions I can suggest:
- Do not cut spaces in gpuhunt. E.g., skip GPUs that have spaces.
- Introduce a whitelist of supported GPUs either in gpuhunt or in dstack.
- Pass the initial GPU type from gpuhunt to dstack through `RawCatalogItem.provider_data`.

src/dstack/_internal/core/backends/jarvislabs/compute.py
```diff
@@ -0,0 +1,364 @@
...
+            return
+        status_value = str(status.get("status")).lower()
+        if status_value == "failed":
+            message = _format_failed_status(status)
+            if _looks_like_no_capacity(message):
+                raise NoCapacityError(message)
+            raise ProvisioningError(message)
+        if status_value == "running":
+            return
+        time.sleep(CREATE_FAILURE_POLL_INTERVAL_SECONDS)
```
jvstme: This 30-second wait is happening before we commit the instance ID to the database, so if the server is interrupted here, the instance will become orphaned.

Have you confirmed that this wait is necessary? Does Jarvis Labs actually interrupt instances within 30 seconds, and does this happen often?

src/dstack/_internal/core/backends/jarvislabs/compute.py
```diff
@@ -0,0 +1,364 @@
...
+DEFAULT_USERNAME = "ubuntu"
+SSH_CONNECT_TIMEOUT_SECONDS = 10
+SSH_SETUP_TIMEOUT_SECONDS = 240
+SSH_LAUNCH_TIMEOUT_SECONDS = 60
+CREATE_FAILURE_POLL_INTERVAL_SECONDS = 5
+CREATE_FAILURE_POLL_TIMEOUT_SECONDS = 30
+
+
+class JarvisLabsCompute(
+    ComputeWithFilteredOffersCached,
```
jvstme: It should be possible to use `ComputeWithAllOffersCached` instead, which will have more cache hits and potentially a smaller memory footprint (`ComputeWithFilteredOffersCached` stores redundant copies of the same offer for different requirements).

The prerequisite to using `ComputeWithAllOffersCached` is addressing [this comment](https://github.com/dstackai/gpuhunt/pull/231#discussion_r3266049342) from the gpuhunt PR, so that gpuhunt returns the same offers regardless of the query filter

jvstme (approved):

jvstme (approved):

PR #3885, opened 2026-05-17
Reject negative retry durations

pragnyanramtha:
## Summary
- Reject negative `ProfileRetry.duration` values during configuration parsing.
- Reuse the same validation for the `--retry-duration` CLI override.
- Add regression coverage for model parsing and CLI argument handling.

## Why
`retry.duration` represents the maximum retry window. Previously, values like `-1` and `"-1"` parsed successfully, creating nonsensical retry specs instead of a configuration error.

## Validation
- `uv run pytest src/tests/_internal/core/models/test_profiles.py src/tests/_internal/cli/services/configurators/test_profile.py` — 14 passed
- `uv run pytest src/tests/_internal/core/models/test_profiles.py src/tests/_internal/core/models/test_runs.py src/tests/_internal/core/models/test_fleets.py` — 29 passed
- `uv run ruff check .` — All checks passed
- `uv run ruff format --check src/dstack/_internal/core/models/profiles.py src/dstack/_internal/cli/services/profile.py src/tests/_internal/core/models/test_profiles.py src/tests/_internal/cli/services/configurators/test_profile.py` — 4 files already formatted

## AI Assistance
This PR was written with AI assistance from OpenAI Codex.

### DIFF
--- a/src/dstack/_internal/server/services/runs/spec.py
+++ b/src/dstack/_internal/server/services/runs/spec.py
@@ -4,6 +4,7 @@ from dstack._internal.core.models.configurations import (
     SERVICE_HTTPS_DEFAULT,
     ServiceConfiguration,
 )
+from dstack._internal.core.models.profiles import ProfileRetry
 from dstack._internal.core.models.repos.virtual import DEFAULT_VIRTUAL_REPO_ID, VirtualRunRepoData
 from dstack._internal.core.models.routers import RouterType
 from dstack._internal.core.models.runs import LEGACY_REPO_DIR, AnyRunConfiguration, RunSpec
@@ -74,6 +75,7 @@ def validate_run_spec_and_set_defaults(
     # the defaults depend on the server version, not the client version.
     if run_spec.run_name is not None:
         validate_dstack_resource_name(run_spec.run_name)
+    _validate_retry_duration(run_spec)
     for mount_point in run_spec.configuration.volumes:
         if not is_valid_docker_volume_target(mount_point.path):
             raise ServerClientError(f"Invalid volume mount path: {mount_point.path}")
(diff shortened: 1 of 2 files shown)


src/dstack/_internal/core/models/profiles.py
jvstme, line 1: Hi @pragnyanramtha, thanks for contributing.

We use the same Pydantic models when loading data from the database, so making the Pydantic validation stricter is not an option — it would cause the dstack server to fail when loading existing runs that may already have a negative retry duration set. You can confirm this by starting a run with a negative retry duration on a master or release version of the dstack server, upgrading the server to the version from this PR (without stopping the run), and then following the server logs.

As an alternative, I’d suggest adding this validation only on the server side for newly submitted runs. See [`validate_run_spec_and_set_defaults`](https://github.com/dstackai/dstack/blob/117195d7194555f3556920d3a0b38f8cb76041ce/src/dstack/_internal/server/services/runs/spec.py#L65).

jvstme (changes requested):

src/dstack/_internal/core/models/profiles.py
pragnyanramtha: Thanks for the pointer. Updated in 29fb4119: removed the stricter ProfileRetry/Pydantic validation and moved the negative retry.duration check into validate_run_spec_and_set_defaults, so existing DB-loaded runs can still parse. I also added tests covering backward-compatible ProfileRetry(duration=-1) parsing plus server rejection for newly submitted run specs.

Validated locally:
- uv run pytest src/tests/_internal/server/services/runs/test_spec.py src/tests/_internal/cli/services/configurators/test_profile.py -q
- uv run ruff check src/dstack/_internal/server/services/runs/spec.py src/tests/_internal/server/services/runs/test_spec.py src/dstack/_internal/cli/services/profile.py src/tests/_internal/cli/services/configurators/test_profile.py src/dstack/_internal/core/models/profiles.py
- uv run ruff format --check src/dstack/_internal/server/services/runs/spec.py src/tests/_internal/server/services/runs/test_spec.py src/dstack/_internal/cli/services/profile.py src/tests/_internal/cli/services/configurators/test_profile.py src/dstack/_internal/core/models/profiles.py
- git diff --check

jvstme (approved):
Looks good to me.

There seems to be an unrelated Pyright issue in CI right now. We’ll fix it in master shortly, and then I’ll merge this PR

PR #3913, opened 2026-05-28
[Docs]: Tenant isolation guide

jvstme:
Add a guide with instructions on hardening the
isolation between tenants and hosts (where
"tenants" can refer to users or jobs, not
necessarily organizations; couldn't come up with a
more concise yet descriptive title).


un-def (approved):

mkdocs/docs/guides/tenant-isolation.md
```diff
@@ -0,0 +1,96 @@
...
+
+By default, most `dstack` jobs run in host networking mode, sharing the network namespace of the instance. This is significant because the instance also runs internal `dstack` services — including APIs used to manage containers and SSH authorized keys. A container on the host network can reach these APIs.
+
+The `DSTACK_SERVER_JOB_NETWORK_MODE` environment variable controls which jobs get host vs. bridge networking:
+
+| Value | Name | Behavior |
+|---|---|---|
+| `1` | `HOST_FOR_MULTINODE_ONLY` | Host for distributed tasks, bridge otherwise |
+| `2` | `HOST_WHEN_POSSIBLE` | Host whenever the job occupies a full instance (default) |
+| `3` | `FORCED_BRIDGE` | Always bridge, including distributed tasks |
```
peterschmidt85, line 63: Must be explicit that this prevents both distributed tasks and PD disaggregation to work

mkdocs/docs/guides/tenant-isolation.md
```diff
@@ -0,0 +1,96 @@
...
+
+While attached to a run, users can SSH directly into the host machine — not just the container — using:
+
+```shell
+ssh <run-name>-host
+```
+
+This gives unrestricted access to the underlying instance, bypassing container boundaries.
+
+If desired, host SSH access can be disabled server-wide by configuring the [SSH proxy](server-deployment.md#ssh-proxy) and setting the following environment variable when starting the `dstack` server:
```
peterschmidt85, line 24: DOn't we want to include a guide on how to set up SSH proxy here?

jvstme: There's a reference to the deployment guide, I wouldn't duplicate the guide here

mkdocs/docs/guides/tenant-isolation.md
jvstme: Technically, distributed tasks can still work with `FORCED_BRIDGE`, e.g., if they use some external coordinator and don't need inter-job communication.

I've updated the section to describe the difference between host and bridge networking, and mentioned that host networking facilitates distributed tasks and routers. This should help users make a more informed decision

mkdocs/docs/guides/tenant-isolation.md
```diff
@@ -0,0 +1,98 @@
...
+title: Tenant isolation
+description: Restricting access to hosts managed by dstack
+---
+
+# Tenant isolation
+
+`dstack` assumes mutual trust between users of the same project. While users' jobs run in Docker containers, users and their containers may have broad access to the underlying hosts. This guide explains how to restrict access to the host when stronger boundaries are required.
+
+!!! info "Disclaimer"
+    Even with all precautions, complete isolation on shared hardware is hardly achievable — container escape vulnerabilities are common. The best way to provide true isolation between users is to place them in different `dstack` projects and not share hardware between them.
```
peterschmidt85: Is there a way to make it sound less harsh? IMO its too much focusing around "hardly achievable" i don't think the main question of the reader is whether its archivable or not. Maybe we can make the text here more neutral? Or a the very least don't frame it as "Disclaimer".

mkdocs/docs/guides/tenant-isolation.md
```diff
@@ -0,0 +1,98 @@
...
+| Property | Applies to |
+|---|---|
+| `privileged: true` | Tasks, dev environments, services |
+| `docker: true` | Tasks, dev environments, services |
+| `replicas[i].privileged: true` | Services with replica groups |
+| `replicas[i].docker: true` | Services with replica groups |
+
+To block runs that request privileged mode, write a [REST plugin](../reference/plugins/rest/index.md) or a [Python plugin](../reference/plugins/python/index.md) with an apply policy.
+
+<div editor-title="src/isolation_plugin/__init__.py">
```
peterschmidt85: IMO plugins are still experimental and I'm somewhat not super confident of pushing everyone in this direction just yet. Maybe we could make it a collapsible section? Or even better add this example to the Plugins documentation, and only use a link from here - as we do with SSH proxy?

peterschmidt85 (approved):

mkdocs/docs/guides/tenant-isolation.md
jvstme: Updated the wording

mkdocs/docs/guides/tenant-isolation.md
jvstme: The plugins documentation already has a similar example actually. I updated it and removed the inline example from this guide

jvstme:
@peterschmidt85, I'm merging, but if you feel like adjusting anything else, please do

PR #3917, opened 2026-05-29
[shim] Pass proxy variables to the container

### LINKED ISSUE #3906: [Bug]: dstack-shim doesn't respect proxy-related environment variables
## Problem

  When `dstack-shim` starts a job container, the entrypoint runs an `install_pkg openssh-server` step (plus, in the non-shim flow, `install_pkg curl` and a `curl` download of the runner binary). On hosts that
  only have outbound internet access through an HTTP proxy, this step fails with a network/DNS error even though the shim itself was started with `http_proxy`/`https_proxy` exported and Docker pulls succeed (the
   Docker daemon reads its own proxy config).
  
  The root cause is that **no proxy env vars are forwarded into the container**, so `apt-get`/`yum`/`apk` inside the entrypoint shell never see them.

  ## Repro
  
  1. Run `dstack-shim` on a host that requires `http_proxy`/`https_proxy` for outbound traffic.
  2. Submit any job that uses an image without `sshd` preinstalled (e.g., a plain `ubuntu:22.04`).
  3. Container entrypoint exits non-zero in `apt-get update` / `apt-get install -y openssh-server`.

  ## Where the gap is

  The entrypoint chain is built in `getSSHShellCommands()` and runs `install_pkg openssh-server`:

  https://github.com/dstackai/dstack/blob/master/runner/internal/shim/docker.go#L984-L1000

  The container is created in `createContainer` with `Env` populated only from `PJRT_DEVICE` (plus GPU/HABANA vars later); nothing forwards proxy vars from the shim's own environment:

  https://github.com/dstackai/dstack/blob/master/runner/internal/shim/docker.go#L862-L901
  
  `TaskConfig` has no env field at all, so the server can't pass them through either:

  https://github.com/dstackai/dstack/blob/master/runner/internal/shim/models.go#L83-L105

  The same install snippet exists on the Python side (non-shim flow) and has the same gap:

  https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/core/backends/base/compute.py#L982-L1027

  Docker does not auto-propagate the daemon's environment into containers, so even if proxy vars are set on the host (e.g., via `get_shim_env` or cloud-init), they don't reach the entrypoint.

  ## Suggested fix

  Minimal change in `createContainer` (`runner/internal/shim/docker.go`): forward proxy vars from the shim's own environment into the container `Env`:

  ```go
  for _, name := range []string{
      "http_proxy", "https_proxy", "no_proxy",
      "HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY",
  } {
      if v, ok := os.LookupEnv(name); ok {
          envVars = append(envVars, name+"="+v)
      }
  }
  ```

  A more flexible follow-up would be to add an `Env` field to `TaskConfig` so the server can pass per-task env vars to the shim.

  ## Environment

  - dstack version: `0.20.21`
  - Backend: `ssh`
  - Host OS: `Ubuntu 26.04 LTS`
  - Container image: `nvidia/cuda:12.8.0-devel-ubuntu24.04`

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### dstack version

0.20.21

### Server logs

```shell
Exited (none)
W: Some index files failed to download. They have been ignored, or old ones used instead.
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package openssh-server
```

### Additional information

_No response_

un-def:
Fixes: https://github.com/dstackai/dstack/issues/3906

### DIFF
--- a/runner/cmd/shim/main.go
+++ b/runner/cmd/shim/main.go
@@ -27,6 +27,13 @@ import (
 // Version is a build-time variable. The value is overridden by ldflags.
 var Version string
 
+// https://everything.curl.dev/usingcurl/proxies/env.html
+// https://cs.opensource.google/go/x/net/+/657eb1317b5dd33038d683297c6be9cae05fa97d:http/httpproxy/proxy.go
+// We accept HTTP_PROXY in upper case without additional checks as it's unlikely that
+// the shim is running in the CGI context
+// The lower case form should be used as some applications ignore the upper case form, e.g., curl, apt
+const defaultPassEnv = "http_proxy,https_proxy,no_proxy,HTTP_PROXY,HTTPS_PROXY,NO_PROXY"
+
 func main() {
 	os.Exit(mainInner())
 }
@@ -147,6 +154,13 @@ func mainInner() int {
 				Sources:     cli.EnvVars("DSTACK_DCGM_ADDRESS"),
 			},
 			/* Docker Parameters */
+			&cli.StringFlag{
+				Name:        "pass-env",
+				Usage:       "Environment variables to pass on to the container, a comma-separated list of names",
+				Value:       defaultPassEnv,
+				Destination: &args.Docker.PassEnv,
+				Sources:     cli.EnvVars("DSTACK_DOCKER_PASS_ENV"),
+			},
 			&cli.BoolFlag{
 				Name:        "privileged",
 				Usage:       "Give extended privileges to the container",
--- a/runner/internal/shim/docker.go
+++ b/runner/internal/shim/docker.go
@@ -135,6 +135,7 @@ type DockerRunner struct {
 	client       *docker.Client
 	dockerParams DockerParameters
 	dockerInfo   dockersystem.Info
+	baseEnv      []string
 	gpus         []host.GpuInfo
 	gpuVendor    gpu.GpuVendor
 	gpuLock      *GpuLock
@@ -149,6 +150,15 @@ func NewDockerRunner(ctx context.Context, dockerParams DockerParameters) (*Docke
 	dockerInfo, err := client.Info(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("get docker info: %w", err)
+	}
+
+	// Copy variables once rather than on a per-task basis
+	// We don't expect variables to change during the shim's lifetime
+	baseEnv := []string{}
+	for _, name := range dockerParams.DockerPassEnv() {
+		if value, ok := os.LookupEnv(name); ok {
+			baseEnv = append(baseEnv, fmt.Sprintf("%s=%s", name, value))
+		}
 	}
 
 	var gpuVendor gpu.GpuVendor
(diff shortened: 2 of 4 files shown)


un-def:
## Test setup

### Prepare a VM with HTTP proxy:

#### Install tinyproxy
```shell
sudo apt update
sudo apt install -y tinyproxy
sudo systemctl enable --now tinyproxy
```

#### Block egress except for tinyproxy
```shell
PROXY_UID=$(id -u tinyproxy)
# let the tinyproxy process talk to the world
sudo iptables -A OUTPUT -m owner --uid-owner $PROXY_UID -j ACCEPT
# let anyone use loopback (so your app can reach the proxy on 127.0.0.1:8888)
sudo iptables -A OUTPUT -o lo -j ACCEPT
# keep already-open connections alive
sudo iptables -A OUTPUT -m conntrack --ctstate ESTABLISHED,RELATED -j ACCEPT
# reject everyone else's outbound HTTP/HTTPS
sudo iptables -A OUTPUT -p tcp -m multiport --dports 80,443 -j REJECT
```

#### Configure HTTP proxy for users
It's required since `dstack` logs in via SSH and calls `curl` to download the shim binary. Without `*_proxy` variables, provisioning would fail.

Note, we use the lower case form – both `curl` and `apt` ignore `HTTP_PROXY` due to CGI: https://everything.curl.dev/usingcurl/proxies/env.html#http_proxy-in-lower-case-only

Add these lines to `/etc/environment` (see [`pam_env(8)`](https://man7.org/linux/man-pages/man8/pam_env.8.html)):

```
http_proxy="http://127.0.0.1:8888"
https_proxy="http://127.0.0.1:8888"
no_proxy="localhost,127.0.0.1,::1"
```

#### Configure HTTP proxy for services
At least for Docker and `dstack-shim`. The easiest way to set proxy variables for all services via [`systemd.conf`](https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html) drop-in. Create `/etc/systemd/system.conf.d/proxy.conf`:

```
[Manager]
DefaultEnvironment="http_proxy=http://127.0.0.1:8888"
DefaultEnvironment="https_proxy=http://127.0.0.1:8888"
DefaultEnvironment="no_proxy=localhost,127.0.0.1,::1"
```

#### Reload systemd and services
```shell
sudo systemctl daemon-reexec
sudo systemctl restart docker.service
```

#### Check the setup
```shell
$ ssh debian@192.168.122.58 -i ~/.ssh/vm_debian
```

```shell
$ curl https://api.github.com/zen
Avoid administrative distraction.
$ unset http_proxy https_proxy
$ curl https://api.github.com/zen
curl: (7) Failed to connect to api.github.com port 443 after 0 ms: Could not connect to server
```

#### Deploy an SSH fleet
```yaml
type: fleet
name: vm
ssh_config:
  hosts:
    - hostname: 192.168.122.58
      user: debian
      identity_file: ~/.ssh/vm_debian
```

#### Run a job
```yaml
type: dev-environment
image: ubuntu  # or alpine or fedora, all three package managers should work
```

#### Disable env passing and check again

Redeploy the fleet with additional `env` section
```yaml
type: fleet

... 

env:
  - DSTACK_DOCKER_PASS_ENV=
```

Run the same run configuration
```
Exited (none)
Package openssh-server is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'openssh-server' has no installation candidate
```

jvstme (approved):

Issue #1679, opened 2024-09-10
[Bug]: Some locally committed files are empty in runs


jvstme:
### Steps to reproduce

Clone a git repo and commit any binary file or a large (e.g. 1GB) text file. Do not push to remote.
```shell
> git clone https://github.com/dstackai/gpuhunt.git
Cloning into 'gpuhunt'...
remote: Enumerating objects: 1264, done.
remote: Counting objects: 100% (623/623), done.
remote: Compressing objects: 100% (294/294), done.
remote: Total 1264 (delta 417), reused 458 (delta 321), pack-reused 641 (from 1)
Receiving objects: 100% (1264/1264), 272.24 KiB | 1010.00 KiB/s, done.
Resolving deltas: 100% (686/686), done.

> cd gpuhunt/

> dd if=/dev/zero of=test-file bs=65536 count=1
1+0 records in
1+0 records out
65536 bytes (66 kB, 64 KiB) copied, 0.000250103 s, 262 MB/s

> git add test-file

> git commit -m test
```
Run a `dstack` configuration in this repo.
```shell
> printf "type: dev-environment\nide: vscode\nname: test-run" > .dstack.yml
> dstack init
OK
> dstack apply -y
```
Wait until the run is running.

### Actual behaviour

The file is empty in the run.
```shell
> ssh test-run ls -lh /workflow/test-file
-rw-r--r-- 1 root root 0 Sep 10 08:09 /workflow/test-file
```

### Expected behaviour

The file is not empty.
```shell
> ssh test-run ls -lh /workflow/test-file
-rw-r--r-- 1 root root 64K Sep 10 08:09 /workflow/test-file
```

### Additional information

[Here](https://github.com/dstackai/dstack/blob/9c8b7578f82795e5973ee2505210dc876e5ef9e4/src/dstack/_internal/core/models/repos/remote.py#L324) `git diff` only returns file names for binary or large files, not the actual diff.

PR #3918, opened 2026-05-29
Handle repo patch with non-UTF8 sequences

### LINKED ISSUE #3880: [Bug]: dstack apply fails with UnicodeDecodeError when a modified tracked file contains non-UTF-8 bytes
### dstack version

0.20.19

### Python version

3.13.7

### Host OS

Linux 5.15.0-135-generic (Ubuntu)

### Host Arch

x86_64

### What happened?

`dstack apply -f <run config>` aborts during repo-diff packaging when the working
tree contains a modified tracked file whose content has any sequence the UTF-8
decoder rejects (e.g., a stray Latin-1 byte, BOM artifact, or otherwise
malformed UTF-8). In my case the offending file was a LaTeX paper
(`*.tex`) unrelated to the run.

Workaround: `git stash` the offending file before `dstack apply`, then
`git stash pop` afterwards. This obviously doesn't scale if the file has to be
in the diff.

### Steps to reproduce

1. In a git repo, modify a tracked text file so it contains at least one
   non-UTF-8 byte sequence (e.g., a Latin-1-encoded character not valid in
   UTF-8, or a malformed multi-byte sequence).
2. Without committing, run `dstack apply -f <any run config>` from that repo.
3. The CLI aborts with the trace below before any plan is shown.

### Relevant log output

```
File ".../dstack/_internal/cli/services/configurators/run.py", line 567, in get_repo
    repo = get_repo_from_dir(local_path)
File ".../dstack/_internal/core/models/repos/remote.py", line 372, in _repo_diff_verbose
    _interactive_git_proc(repo.git.diff(repo_hash, as_process=True), collector)
File ".../dstack/_internal/core/models/repos/remote.py", line 363, in _interactive_git_proc
    collector.write(stdout)
File ".../dstack/_internal/core/models/repos/remote.py", line 259, in write
    self.buffer.write(v.decode())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 28317: invalid continuation byte
```

### Root cause

`_DiffCollector.write` at `remote.py:259` calls `v.decode()` (i.e.,
`v.decode(\"utf-8\")` with default strict errors) on the raw bytes returned by
`git diff`. Any non-UTF-8 byte sequence in a tracked file makes this raise,
which propagates out of `_repo_diff_verbose` and kills `dstack apply` before
the user can see the plan.

### Suggested fix

In `dstack/_internal/core/models/repos/remote.py`, change:

\`\`\`python
def write(self, v: bytes):
    self.buffer.write(v.decode())
\`\`\`

to either:

\`\`\`python
def write(self, v: bytes):
    self.buffer.write(v.decode(\"utf-8\", errors=\"replace\"))
\`\`\`

(simplest) or use \`codecs.getincrementaldecoder(\"utf-8\")(errors=\"replace\")\` if
chunked decoding across multiple \`write\` calls is expected. Same fix likely
wanted in the \`untracked_files\` loop in \`_repo_diff_verbose\` for parity with
the spirit of #390.

### Related

#390 (closed) — same exception class but for untracked binary files. The fix
there handled the untracked-binary path; tracked-text-with-non-UTF-8 still
falls through to the strict decoder.

un-def:
Git text diff is not actually text in the usual sense -- it may contain any nonprintable bytes.

But since we send patches as files (multipart/form-data), not as JSON strings, we don't need to decode them at all.

In addition, the missing --binary flag was added.
Without this flag, modified binary files were effectively excluded from the patch with "Binary files a and b differ".

Fixes: https://github.com/dstackai/dstack/issues/3880

### DIFF
--- a/src/dstack/_internal/core/models/repos/remote.py
+++ b/src/dstack/_internal/core/models/repos/remote.py
@@ -58,7 +58,7 @@ class RemoteRepoInfo(
 class RemoteRunRepoData(RemoteRepoInfo):
     repo_branch: Optional[str] = None
     repo_hash: Optional[str] = None
-    repo_diff: Annotated[Optional[str], Field(exclude=True)] = None
+    repo_diff: Annotated[Optional[bytes], Field(exclude=True)] = None
     repo_config_name: Optional[str] = None
     repo_config_email: Optional[str] = None
 
@@ -183,13 +183,15 @@ class RemoteRepo(Repo):
     def has_code_to_write(self) -> bool:
         # repo_diff is:
         # * None for RemoteRepo.from_url()
-        # * an empty string for RemoteRepo.from_dir() if there are no changes ("clean" state)
-        # * a non-empty string for RemoteRepo.from_dir() if there are changes ("dirty" state)
+        # * empty bytes for RemoteRepo.from_dir() if there are no changes ("clean" state)
+        #   and untracked files
+        # * non-empty bytes for RemoteRepo.from_dir() if there are changes ("dirty" state)
+        #   and/or untracked files
         return bool(self.run_repo_data.repo_diff)
 
     def write_code_file(self, fp: BinaryIO) -> str:
         if self.run_repo_data.repo_diff is not None:
-            fp.write(self.run_repo_data.repo_diff.encode())
+            fp.write(self.run_repo_data.repo_diff)
         return get_sha256(fp)
 
     def get_repo_info(self) -> RemoteRepoInfo:
@@ -238,7 +240,7 @@ class _DiffCollector:
         self.delay = delay
         self.warned = False
         self.start_time = time.monotonic()
-        self.buffer = io.StringIO()
+        self.buffer = io.BytesIO()
 
     def timeout(self):
         now = time.monotonic()
@@ -256,9 +258,9 @@ class _DiffCollector:
         )
 
     def write(self, v: bytes):
-        self.buffer.write(v.decode())
+        self.buffer.write(v)
 
-    def get(self) -> str:
+    def get(self) -> bytes:
         if self.warned:
             print()
         return self.buffer.getvalue()
@@ -366,10 +368,10 @@ def _interactive_git_proc(
             continue
 
 
-def _repo_diff_verbose(repo: git.Repo, repo_hash: str, warning_time: float = 5) -> str:
+def _repo_diff_verbose(repo: git.Repo, repo_hash: str, warning_time: float = 5) -> bytes:
     collector = _DiffCollector(warning_time)
     try:
-        _interactive_git_proc(repo.git.diff(repo_hash, as_process=True), collector)
+        _interactive_git_proc(repo.git.diff(repo_hash, binary=True, as_process=True), collector)
         for filename in repo.untracked_files:
             _interactive_git_proc(
                 repo.git.diff("/dev/null", filename, no_index=True, binary=True, as_process=True),


src/dstack/_internal/core/models/repos/remote.py
```diff
@@ -58,7 +58,7 @@ class RemoteRepoInfo(
 class RemoteRunRepoData(RemoteRepoInfo):
     repo_branch: Optional[str] = None
     repo_hash: Optional[str] = None
-    repo_diff: Annotated[Optional[str], Field(exclude=True)] = None
+    repo_diff: Annotated[Optional[bytes], Field(exclude=True)] = None
```
jvstme, line 61: (nit) Technically a breaking change for SDK users, since `RemoteRunRepoData` is stored in the public but undocumented `run_repo_data` attribute of the public `RemoteRepo` class

src/dstack/_internal/core/models/repos/remote.py
jvstme, line 1: It looks like the PR also partially fixes and partially changes the behavior (from one unexpected to another unexpected) described #1679.
- Locally committed binary files are now delivered as expected.
- Locally committed large text files now cause `dstack apply` to fail (previously, such files would be delivered with no contents).

  ```shell
  $ dstack apply
  GitCommandError: Cmd('git') failed due to: exit code(128)
    cmdline: git diff --binary 623de5df3d3d13007661c7ee1a13a8d5197e734a
    stderr: 'fatal: unable to generate diff for test-file
  '
  $ ls -lh test-file 
  -rw-r--r--. 1 root root 1.1G Jun  1 14:08 test-file
  ```

jvstme (approved):

src/dstack/_internal/core/models/repos/remote.py
un-def: Good point, but I'd consider all undocumented fields as private

src/dstack/_internal/core/models/repos/remote.py
un-def: It seems that large ASCII files (sometimes) are detected as binary files.

Note that both produce `Binary files ... differ` without `--binary` but with `--binary` a "truly" binary file is handled correctly despite being larger (2GB vs 1.4GB)  while with a base64-encoded file Git fails with `fatal: unable to generate diff `. Weird.

```shell
dd if=/dev/urandom bs=1M count=1000 | base64 > blob.asc
dd if=/dev/urandom bs=1M count=2000 > blob.bin

ls -l blob.*
-rw-rw-r-- 1 def def 1.4G Jun  2 09:33 blob.asc
-rw-rw-r-- 1 def def 2.0G Jun  2 09:10 blob.bin

file blob.asc
blob.asc: ASCII text

git --no-pager diff --no-index -- /dev/null blob.asc
diff --git a/blob.asc b/blob.asc
new file mode 100644
index 0000000..f2d7c9e
Binary files /dev/null and b/blob.asc differ

git --no-pager diff --no-index --binary -- /dev/null blob.asc
diff --git a/blob.asc b/blob.asc
new file mode 100644
index 0000000..f2d7c9e
fatal: unable to generate diff for /dev/null

file blob.bin
blob.bin: data

git --no-pager diff --no-index -- /dev/null blob.bin
diff --git a/blob.bin b/blob.bin
new file mode 100644
index 0000000..f71bad7
Binary files /dev/null and b/blob.bin differ

git --no-pager diff --no-index --binary -- /dev/null blob.bin
diff --git a/blob.bin b/blob.bin
new file mode 100644
index 0000000000000000000000000000000000000000..f71bad7e725b891b8c0260707b8c34481f3da331
GIT binary patch
<ommitted>

literal 0
HcmV?d00001
```

Another unexpected behavior still present:

* Uncommitted (untracked) large text files are empty (`git diff` fails but its exit status is ignored due to `ignore_status=True`):

   ```shell
  dd if=/dev/random bs=1M count=1000 | base64 > blob.asc
  1000+0 records in
  1000+0 records out
  1048576000 bytes (1.0 GB, 1000 MiB) copied, 4.00187 s, 262 MB/s

  git --no-pager diff --no-index --binary -- /dev/null blob.asc
  diff --git a/blob.asc b/blob.asc
  new file mode 100644
  index 0000000..ce045ae
  fatal: unable to generate diff for /dev/null
  ```

PR #3932, opened 2026-06-05
Replica group `spot_policy` and `reservation`

### LINKED ISSUE #3929: [Feature]: Support distributing service replicas across reserved and spot capacity
### Problem

When running a service with multiple replicas, it's common to want a **guaranteed baseline** (reserved/on-demand instances) with **spot overflow** for burst capacity. The natural approach is to create two fleets (one on-demand, one spot) and reference both in the service's `fleets` parameter.

However, dstack currently selects a **single fleet for all replicas** of a service. This makes it impossible to distribute replicas across fleets with different purchasing options.

### Observed behavior

**Setup:**
- Fleet A: `nodes: 1`, `spot_policy: on-demand` (with capacity reservation)
- Fleet B: `nodes: 0..1`, `spot_policy: spot`
- Service: `replicas: 2`, `fleets: [fleet-a, fleet-b]`, `spot_policy: auto`

**Result:** `failed_to_start_due_to_no_capacity` — "Failed to use specified fleets"

**When Fleet B is changed to `nodes: 0..2`:**
All 2 replicas are provisioned on Fleet B (spot), Fleet A (reserved) is completely ignored.

### Root cause (code analysis)

1. **Single fleet per run** — `_should_wait_for_run_fleet_assignment` in `jobs_submitted.py` ensures only `replica_num=0` selects the fleet. Once `run_model.fleet_id` is set, all subsequent replicas are constrained to the same fleet.

2. **Capacity check excludes undersized fleets** — `_run_can_fit_into_fleet` in `plan.py` requires:
   ```python
   fleet_available_capacity = fleet.nodes.max - occupied_instances
   if fleet_available_capacity < nodes_required_num:  # total min replicas
       return False  # fleet excluded entirely
   ```
   A fleet is excluded if its `nodes.max` can't accommodate ALL replicas.

3. **Price-based sort with no fleet priority** — `find_optimal_fleet_with_offers` sorts by:
   ```python
   sort_key=(
       not candidate.has_pool_capacity,      # idle instances first
       candidate.min_instance_offer_price,   # cheapest price
       min_backend_offer_price,
   )
   ```
   Spot always wins over on-demand/reserved on price.

### Desired behavior

Allow replicas of a single service to be distributed across multiple fleets. Example use case:

```yaml
type: service
name: my-service
replicas: 3..10

fleets:
  - name: gpu-reserved
    min_replicas: 1  # always keep at least 1 replica here
  - name: gpu-spot
    # overflow capacity

spot_policy: auto
```

### Possible solutions

1. **Per-fleet replica pinning** — allow specifying min/max replicas per fleet in the `fleets` list
2. **Fleet priority/weight** — ordered preference with fallback (try fleet A first, overflow to fleet B)
3. **Remove the single-fleet constraint** — allow `run_model.fleet_id` to be per-job instead of per-run, and change `_run_can_fit_into_fleet` to check aggregate capacity across all candidate fleets

### Workaround

Currently the only workaround is running two separate services (one per fleet) behind an external load balancer, which adds operational complexity and breaks the single-endpoint abstraction.

Yes

jvstme:
Allow to define `spot_policy` and `reservation`
per replica group in service configurations.

```yaml
type: service
image: my-image
port: 80

replicas:
- name: baseline
  reservation: my-reservation
  count: 1
- name: overflow
  spot_policy: auto
  count: 0..3
  scaling:
    metric: rps
    target: 1
```

Closes #3929

### DIFF
--- a/src/dstack/_internal/core/backends/base/compute.py
+++ b/src/dstack/_internal/core/backends/base/compute.py
@@ -318,7 +318,7 @@ class ComputeWithCreateInstanceSupport(ABC):
             user=run.user,
             ssh_keys=[SSHKey(public=project_ssh_public_key.strip())],
             volumes=volumes,
-            reservation=run.run_spec.configuration.reservation,
+            reservation=job.job_spec.requirements.reservation,
             tags=run.run_spec.merged_profile.tags,
         )
         instance_offer = instance_offer.copy()
--- a/src/dstack/_internal/core/compatibility/runs.py
+++ b/src/dstack/_internal/core/compatibility/runs.py
@@ -122,6 +122,10 @@ def get_run_spec_excludes(run_spec: RunSpec) -> IncludeExcludeDictType:
                 replica_group_excludes["nvcc"] = True
             if all(g.privileged is None for g in replicas):
                 replica_group_excludes["privileged"] = True
+            if all(g.spot_policy is None for g in replicas):
+                replica_group_excludes["spot_policy"] = True
+            if all(g.reservation is None for g in replicas):
+                replica_group_excludes["reservation"] = True
             if replica_group_excludes:
                 configuration_excludes["replicas"] = {"__all__": replica_group_excludes}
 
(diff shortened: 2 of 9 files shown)


Bihan (approved):
In this [check](https://github.com/dstackai/dstack/blob/d0e95cb718d1c19898a4c41995eba0b7e54a074c/src/dstack/_internal/server/services/offers.py#L71), when reservation is set at the replica-group level, it is resolved into
job.job_spec.requirements.reservation, while profile.reservation remains None.

Rest is ready to merge.

Bihan (approved):

PR #3938, opened 2026-06-05
Use paginated Vast.ai v1 instances endpoint

7Z0nE:
## Summary

- Move `get_instances()` and `get_instance(id)` in the Vast.ai backend off `GET /api/v0/instances/` and onto `GET /api/v1/instances/`. Vast.ai's [April 2026 product update](https://vast.ai/article/april-2026-product-update) announces that the v0 list endpoint will start returning paginated results (max 25 per response), so callers must handle the keyset cursor. The v1 endpoint already enforces that contract via `after_token` / `next_token`.
- `get_instances()` walks pages until `next_token` is empty so the existing 3 s cache fills with the full list.
- `get_instance(id)` issues a targeted v1 query with `select_filters={"id": {"eq": id}}` and `limit=1`, so a single-row lookup stays one HTTP request instead of paginating through every owned instance.
- Other Vast.ai endpoints used by the backend — `PUT /asks/{id}/` (create), `DELETE /instances/{id}/` (destroy), `POST /bundles/` (search offers), `PUT /instances/request_logs/{id}/` — have no v1 replacement in the published OpenAPI yet, so they remain on v0.

## Test plan

- [x] `uv run pytest src/tests/_internal/core/backends/vastai/ -v` — 15 passed (4 new `test_api_client.py` cases cover pagination, single-instance filter, 404 handling, and that destroy still hits v0).
- [x] `uv run pytest src/tests/_internal/core/backends/ -q` — 244 passed (no regressions in sibling backends).
- [x] `uv run pre-commit run --files <changed>` — ruff/format clean.
- [ ] Smoke test against a real Vast.ai account (would appreciate a maintainer doing this — I do not have credentials).

### DIFF
--- a/src/tests/_internal/core/backends/vastai/test_api_client.py
+++ b/src/tests/_internal/core/backends/vastai/test_api_client.py
@@ -1,0 +1,108 @@
+import json
+from urllib.parse import parse_qs, urlparse
+
+import pytest
+
+from dstack._internal.core.backends.vastai.api_client import VastAIAPIClient
+
+
+class _FakeResponse:
+    def __init__(self, payload, status_code=200):
+        self._payload = payload
+        self.status_code = status_code
+        self.text = json.dumps(payload)
+
+    def json(self):
+        return self._payload
+
+    def raise_for_status(self):
+        if self.status_code >= 400:
+            raise AssertionError(f"unexpected status {self.status_code}")
+
+
+@pytest.fixture
+def client():
+    return VastAIAPIClient(api_key="test-key")
+
+
+def _parse_call(call):
+    """Return (path, query_dict) for a recorded get() call."""
+    url = call.args[0] if call.args else call.kwargs["url"]
+    parsed = urlparse(url)
+    query = {k: v[0] for k, v in parse_qs(parsed.query).items()}
+    if "params" in call.kwargs and call.kwargs["params"]:
+        for k, v in call.kwargs["params"].items():
+            query[k] = str(v)
+    return parsed.path, query
+
+
+def test_get_instances_uses_v1_endpoint_and_paginates(client, monkeypatch):
+    pages = [
+        {"instances": [{"id": 1}, {"id": 2}], "next_token": "tok-1"},
+        {"instances": [{"id": 3}], "next_token": None},
+    ]
+    calls = []
+
+    def fake_get(url, params=None):
+        calls.append((url, dict(params or {})))
+        return _FakeResponse(pages[len(calls) - 1])
+
+    monkeypatch.setattr(client.s, "get", fake_get)
+
+    instances = client.get_instances(cache_ttl=0)
+
+    assert [i["id"] for i in instances] == [1, 2, 3]
+    assert len(calls) == 2
+    # First call hits v1 and includes select_filters={} and a limit.
+    first_url, first_params = calls[0]
+    assert "/api/v1/instances/" in first_url
+    assert first_params["select_filters"] == "{}"
+    assert first_params["limit"] == 25
+    assert "after_token" not in first_params
+    # Second call carries the next_token from the prior response.
+    _, second_params = calls[1]
+    assert second_params["after_token"] == "tok-1"
+
+
+def test_get_instance_uses_v1_select_filters(client, monkeypatch):
+    calls = []
+
+    def fake_get(url, params=None):
+        calls.append((url, params))
+        return _FakeResponse(
+            {"instances": [{"id": 42, "actual_status": "running"}], "next_token": None}
+        )
+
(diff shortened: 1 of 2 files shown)


r4victor:
@7Z0nE, thanks for the PR. The change looks solid and can be merged. Although I'm not sure if we need get_instances() and thus _list_instances_v1() and the cache at all since it's used only for get_instance(). Apparently, it was needed when there was no [show-instance API](https://docs.vast.ai/api-reference/instances/show-instance). I'm happy with either of these:

* Drop get_instances() and  _list_instances_v1() and the cache in favor of show-instance.
* Implement get_instance() via _list_instances_v1() as you did but drop get_instances().
* Keep the PR as is in case get_instances() is useful in future.

WDYT?

src/tests/_internal/core/backends/vastai/test_api_client.py
r4victor, line 1: Let's drop this new test file completely – the tests don't seem very useful.

7Z0nE: will do

7Z0nE:
> I'm not sure if we need get_instances() and thus _list_instances_v1() and the cache at all since it's used only for get_instance(). Apparently, it was needed when there was no [show-instance API](https://docs.vast.ai/api-reference/instances/show-instance)

The show-instance API (GET /api/v0/instances/{id}) still has a "v0" in the path. Requests to this v0 endpoint will not work with API keys generted from the new v1 endpoints. Thus we cannot use show-instance and need to keep the workaround with the v1/instances endpoint (https://docs.vast.ai/api-reference/instances/show-instances).

I will test this again to clarify. If it works, then I would do a simplification.

> Drop get_instances() and _list_instances_v1() and the cache in favor of show-instance.
I assumed that get_instances is called by users of the `VastAiAPIClient` and therefore needs to be supported.
A possibility I see would be to migrate all usages to list_instances with a pagination interface.

will post more updates when I have time to go deeper.

askobara:
Hello there!

I've faced the same issue, so found this PR.

> Requests to this v0 endpoint will not work with API keys generted from the new v1 endpoints

@7Z0nE can you please clarify what do you mean? API keys works for both /v0 and /v1 endpoints if use them in headers `self.s.headers["Authorization"] = f"Bearer {api_key}"`, I'm not seeing any problems with this.

Also, a bit about PR itself, I think that it's a bad design decision to make _url_v0 and _url_v1 methods and related properties, it's quickly become unmaintainable. From my point of view, I would suggest simple creating `API_URL = "https://console.vast.ai/api"`, and use it in inside each endpoint's method in a way `f"{API_URL}/v0/bundles/"`, `f"{API_URL}/v1/instances/"` and so on.

trashhalo:
Confirming this against a real Vast.ai account — the v0 break is live and this migration is the right fix.

```
GET https://console.vast.ai/api/v0/instances/  -> HTTP 410  {"success": false, "error": "deprecated_endpoint", ...}
GET https://console.vast.ai/api/v1/instances/  -> HTTP 200  keys: success, total_instances, label_counts, instances_found, instances, next_token
```

This isn't cosmetic — it breaks backend registration entirely: `auth_test()` calls `get_instances()`, the 410 raises, and dstack surfaces it to the operator as "Invalid credentials" even though the key is fine. Run monitoring hits the same dead endpoint.

I've been running the equivalent v1 route in production on a self-hosted dstack broker (via a local patch) and can confirm end-to-end: backend registration succeeds, offers list, and `dstack apply` provisions and monitors a Vast instance cleanly through the v1 path. My patch is just the naive `/api/v0/` -> `/api/v1/` URL swap, which is why I want to flag that **your PR is the more correct version** — v1 caps at 25 instances/page (per the April 2026 update), so the pagination walk in `_list_instances_v1()` is necessary; a plain swap silently truncates past 25 owned instances.

The v1 response I see preserves the `success`/`instances` schema and just adds the pagination fields (`next_token`, `total_instances`, ...), matching what the PR assumes. Would be great to get this merged — happy to run any additional smoke test against my account if that helps unblock it.

jvstme:
> ```
> GET https://console.vast.ai/api/v0/instances/  -> HTTP 410  {"success": false, "error": "deprecated_endpoint", ...}
> ```

Interestingly, I cannot reproduce this error. Maybe Vast.ai returns it selectively to some users but not others.

Anyways, based on feedback from this PR, I've implemented a slightly simpler solution in #4031 that uses `GET /api/v0/instances/{id}` instead of `GET /api/vX/instances/`. I believe this should solve the issue.

Thank you everyone for participating in this discussion and contributing. The fix is available in `master` and we'll appreciate you testing it. It should be released later this week.

Issue #3953, opened 2026-06-10
Stuck runner/shim install loops re-download the full binaries every check (no backoff, no cap, no content check)


peterschmidt85:
## Summary

A server-driven runner/shim **re-install loop** with no backoff, no attempt cap, and no content check. If an instance's *reported* version never equals the server's *expected* version, the server re-issues an install on every instance check (~7–14s) and the shim **re-downloads the full ~37 MB binaries every time, indefinitely** → large, sustained egress from the public `dstack-runner-downloads` bucket (paid by the bucket owner). Jobs keep running, so it's silent.

**Status:** reproduced live on AWS (below). PR #3954 bounds the **egress**; the **loop itself** is not fixed yet — see *Why the loop isn't fully fixed yet*.

## Steps to reproduce

1. Run **two dstack servers on different versions** (e.g. `0.20.18` and `0.20.23`) against the **same database / backend**.
2. Launch any instance (a plain fleet node is enough — no job required).
3. → the instance re-downloads `dstack-runner` + `dstack-shim` from `dstack-runner-downloads` **continuously** — the full binaries every ~30s, forever.

Two servers is just the simplest trigger. The general condition is **any** persistent mismatch between the instance's reported version and the server's expected version — e.g. a flapping `DSTACK_*_VERSION_URL`, or a download that succeeds but then fails `chmod`/`rename` so the new binary never lands.

### Reproduced live

Two servers (`0.20.18` + `0.20.23`) on one Postgres + AWS backend, one `eu-west-1` instance. The servers interleave, each forcing *its* version; the instance ping-pongs and re-downloads each cycle. Server log (~30s apart):

```
Instance <name>: installing runner 0.20.23 -> 0.20.18 from https://dstack-runner-downloads.s3.eu-west-1.amazonaws.com/0.20.18/binaries/dstack-runner-linux-amd64
Instance <name>: installing runner 0.20.18 -> 0.20.23 from https://dstack-runner-downloads.s3.eu-west-1.amazonaws.com/0.20.23/binaries/dstack-runner-linux-amd64
```

(The original incident was expensive because it was **cross-region** — `eu-west-1` bucket, readers in `us-east-2`/`us-west-2`; the loop itself is region-independent.)

## What's happening

Per instance, every ~7–14s (`instances/__init__.py` — 7s for transitional states, 14s for `IDLE`/`BUSY`) the server:

[...]

PR #3954, opened 2026-06-10
shim: avoid re-downloading runner/shim binaries on forced re-install

peterschmidt85:
Bounds the wasted egress from the runner/shim re-install loop (#3953): the shim re-downloaded the **full** binaries on every server-issued install (`force=true`), and a transient HTTP error during bootstrap could brick the instance. Two changes, both in the download path.

### `downloadFile` — cache + revalidate instead of always re-downloading
`runner/internal/shim/components/utils.go`. Installing `url` into `path` now:

1. `force=false` and `path` exists → return (unchanged).
2. Resolve a **per-URL cache file** next to the binary: `<path>.<sha256(url)[:16]>.cache` (+ a `.etag` sidecar). Keying by URL means different versions get separate caches instead of overwriting one slot.
3. `ensureCached` GETs `url` with `If-None-Match: <stored etag>`:
   - `304 Not Modified` → cache is current, **no body transferred**;
   - `200` → stream into the cache (temp file + atomic rename) and store the response `ETag`.
4. Install the cached bytes into `path`: copy → `chmod` → atomic rename. Skipped if `path` already matches the cache (same size). If `chmod`/rename fails, the next call retries from the cache — **no re-download**.
5. `pruneCaches` keeps the newest `maxCachedVersions` (5) and deletes older ones, so disk stays bounded.

Net: a repeated/forced install of an unchanged version is a `304` with no transfer; two servers alternating between versions download each version **once**, then `304`.

### Bootstrap `curl` — fail instead of saving an error page
`src/dstack/_internal/core/backends/base/compute.py`. The shim-bootstrap `curl` gains `-f` and chains `mv`, so a transient `403`/`5xx` is no longer written to `dstack-shim` and then executed (`/usr/local/bin/dstack-shim: 2: Syntax error: newline unexpected`); it fails cleanly.

### Tests
`runner/internal/shim/components/utils_test.go` (pass under `-race`):
- unchanged artifact under `force` → body served **once**;
- changed `ETag` → re-downloaded and updated;
- `force=false` + existing file → **no** request;
- two alternating versions → each downloaded **exactly once** (guards the per-URL-cache regression).

### Validated end-to-end (AWS)
Two servers on one DB + backend, one real `eu-west-1` instance, custom-versioned builds on the staging bucket, measured via S3 request metrics:
- **Both shims carry this fix** (`0.20.30 ↔ 0.20.31`): 22 conditional GETs over 25 min of continuous flip-flop → **1** real transfer (~17 MB warmup), the rest `304`. Steady-state egress ≈ **0** (unfixed: ~430 MB and growing).
- **Mixed** (unfixed `0.20.29` shim ↔ fixed `0.20.31`): **~454 MB in 11 min, climbing** — see scope below.

### Scope — what this does NOT fix (tracked in #3953)
This PR bounds the **egress**, not the **loop**:
- With two servers expecting different versions the version never converges, so the server keeps issuing installs and the shim still **restarts every ~30s** — now at **zero bytes**, but the churn remains.
- Because the fix lives in the **shim binary**, a **mixed fleet whose in-control shim predates this PR still re-downloads** — the older binary has no cache (the ~454 MB case above).

Fully stopping the loop needs a **server-side** guard (no-downgrade on comparable versions + an attempt cap/backoff). That's **deferred**: dstack's version space (real semver vs CI build-numbers like `27305` vs dev/`None`) has no reliable ordering yet, so the server can't safely decide which version wins. See #3953.

### DIFF
--- a/runner/internal/shim/components/utils.go
+++ b/runner/internal/shim/components/utils.go
@@ -2,6 +2,8 @@ package components
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
@@ -9,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
@@ -20,17 +23,16 @@ const (
 	downloadTimeout = 10 * time.Minute
 	cacheSuffix     = ".cache"
 	etagSuffix      = ".etag"
+	// Max per-version caches kept next to a binary (bounds disk use).
+	maxCachedVersions = 5
 )
 
-// downloadFile ensures that path contains the artifact served at url.
+// downloadFile ensures path holds the artifact at url.
 //
-// To avoid re-transferring bytes that were already fetched (e.g. when the caller
-// repeatedly forces installs of the same artifact), the downloaded bytes are cached
-// next to path and revalidated with the object's ETag via a conditional request: an
-// unchanged artifact is answered with 304 Not Modified and no body is transferred.
-// The cached bytes are then installed to path via chmod+rename; if that finalization
-// fails, a later call retries it from the cache without re-downloading. Without
-// force, an already-installed path is left untouched (preserving prior behavior).
+// Bytes are cached next to path (one cache per URL, validated by ETag), so a repeated
+// or forced install of an unchanged version returns 304 and transfers nothing. Cached
+// bytes are then chmod+renamed into place; a failed rename retries from cache without
+// re-downloading. With force=false an existing path is left as-is.
 func downloadFile(ctx context.Context, url string, path string, mode os.FileMode, force bool) error {
 	if !force {
 		if _, err := os.Stat(path); err == nil {
@@ -41,17 +43,22 @@ func downloadFile(ctx context.Context, url string, path string, mode os.FileMode
 		}
 	}
 
-	cachePath := path + cacheSuffix
-	etagPath := path + etagSuffix
+	// One cache file per URL so several versions can coexist. With a single shared
+	// cache, a request for a different version would overwrite it and force a
+	// re-download every time the requested version changes.
+	key := urlKey(url)
+	cachePath := fmt.Sprintf("%s.%s%s", path, key, cacheSuffix)
+	etagPath := fmt.Sprintf("%s.%s%s", path, key, etagSuffix)
 
 	downloaded, err := ensureCached(ctx, url, cachePath, etagPath)
 	if err != nil {
 		return err
 	}
+	if downloaded {
+		pruneCaches(ctx, path, maxCachedVersions)
+	}
 
-	// If the artifact has not changed and path already matches the cache, there is
-	// nothing to do. Otherwise (first install, changed artifact, or a previous
-	// finalization that failed) install the cached bytes -- without re-downloading.
+	// Install the cached bytes; skip if path already matches the cache.
 	if !downloaded {
 		installed, err := sameSize(path, cachePath)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


runner/internal/shim/components/utils.go
```diff
@@ -2,93 +2,261 @@ package components
...
-	defer func() {
-		if err := tempFile.Close(); err != nil {
-			log.Error(ctx, "close temp file", "err", err)
+	if downloaded {
+		pruneCaches(ctx, path, maxCachedVersions)
+	}
+
+	// Install the cached bytes; skip if path already matches the cache.
+	if !downloaded {
+		installed, err := sameSize(path, cachePath)
```
r4victor, line 63: Same-size adjacent Go builds are unlikely but not impossible.

Suggested fix: after a successful rename in installFile, write the cache's urlKey to a `<path>.installed` marker and replace sameSize with a marker comparison.

runner/internal/shim/components/utils.go
```diff
@@ -2,93 +2,261 @@ package components
...
+	downloadTimeout = 10 * time.Minute
+	cacheSuffix     = ".cache"
+	etagSuffix      = ".etag"
+	// Max per-version caches kept next to a binary (bounds disk use).
+	maxCachedVersions = 5
+)
 
+// downloadFile ensures path holds the artifact at url.
+//
+// Bytes are cached next to path (one cache per URL, validated by ETag), so a repeated
```
r4victor, line 32: I don't quite like putting these helper-files in /usr/local/bin. How about using a shimHomeDir subdir?

runner/internal/shim/components/utils_test.go
```diff
@@ -0,0 +1,171 @@
...
+// every swap.
+func TestDownloadFileFlipFlopDownloadsEachVersionOnce(t *testing.T) {
+	var served23, served18 atomic.Int64
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var body, etag string
+		switch r.URL.Path {
+		case "/0.20.23/runner":
+			body, etag = "RUNNER-0.20.23", `"e23"`
+		case "/0.20.18/runner":
+			body, etag = "RUNNER-0.20.18-x", `"e18"` // intentionally different length
```
r4victor, line 129: See tests only work with different lengths!

src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -926,8 +926,11 @@ def get_shim_pre_start_commands(
     return [
         f"dlpath=$(sudo mktemp -t {DSTACK_SHIM_BINARY_NAME}.XXXXXXXXXX)",
         # -sS -- disable progress meter and warnings, but still show errors (unlike bare -s)
-        f'sudo curl -sS --compressed --connect-timeout 60 --max-time 240 --retry 1 --output "$dlpath" "{url}"',
-        f'sudo mv "$dlpath" {dstack_shim_binary_path}',
+        # -f  -- fail (non-zero exit, no error body written) on HTTP errors, so a transient
+        #        403/5xx is never saved as the shim binary; chain `mv` so a failed download
+        #        is never installed (otherwise it would run as a script -> "Syntax error").
+        f'sudo curl -fsS --compressed --connect-timeout 60 --max-time 240 --retry 1 --output "$dlpath" "{url}"'
+        f' && sudo mv "$dlpath" {dstack_shim_binary_path}',
```
r4victor, line 933: Not sure if ` && ` is needed since commands are chained with  ` && `. Adding `-f` should be sufficient

r4victor (changes requested):
Nice overall

PR #3958, opened 2026-06-11
Support targeting specific instances

peterschmidt85:
## Summary

Adds a strict `instances` schema for runs so users can target existing fleet instances explicitly:

- short form: `instances: [my-fleet-1]`
- by name: `instances: [{name: my-fleet-1}]`
- by hostname: `instances: [{hostname: worker-1}]`
- by fleet instance number: `instances: [{fleet: my-fleet, instance: 1}]`, including `project/fleet`

Compared to https://github.com/dstackai/dstack/pull/3925, this PR keeps the normal fleet selection path mostly unchanged. The new selector resolution and targeted reusable-instance assignment path is used only when `instances` is set; missing targets fail without provisioning new capacity.

## Testing

Automated:

- `uv run --no-sync pytest src/tests/_internal/core/models/test_profiles.py src/tests/_internal/server/services/test_instances.py src/tests/_internal/server/services/runs/test_plan.py src/tests/_internal/server/background/pipeline_tasks/test_submitted_jobs.py -q` (`100 passed, 77 skipped`)
- `uv run --no-sync ruff check ...`
- `uv run --no-sync pyright -p .`
- `git diff --check`

E2E:

- Backend AWS fleet with two `t2.small` instances:
  - normal run without `instances`
  - short-form targeted instance
  - missing target leaves the fleet at two idle instances and does not provision
- SSH fleet backed by two EC2 instances in one VPC/security group using the dstack AMI:
  - normal run
  - target by `name`
  - target by `hostname` using a service
  - target by `{fleet, instance}`
  - missing target fails without occupying instances
  - multinode negative: one selected physical instance is not enough
  - multinode positive: two selected physical instances both run one node

All E2E AWS resources were cleaned up after verification.

AI Assistance: implemented and tested with Codex.


src/dstack/_internal/core/models/profiles.py
```diff
@@ -234,6 +234,58 @@ def crons(self) -> List[str]:
...
+class InstanceHostnameSelector(CoreModel):
+    hostname: Annotated[
+        str, Field(description="The fleet instance hostname or IP address", min_length=1)
+    ]
+
+
+def _parse_fleet_instance_selector_fleet(v: Any) -> Any:
+    if isinstance(v, str):
+        return EntityReference.parse(v)
+    return v
```
jvstme, line 250: (nit) Unnecessary, pydantic is already doing the parsing, since `fleet` is annotated as `EntityReference`

src/dstack/_internal/core/models/profiles.py
```diff
@@ -416,11 +484,23 @@ class ProfileParams(CoreModel):
...
     )
 
+    def dict(self, *args, **kwargs) -> Dict:
+        # super() does not work with pydantic-duality
+        res = CoreModel.dict(self, *args, **kwargs)
+        # Older servers reject unknown fields in profile/request models. Since `instances`
+        # did not exist before, do not serialize it when unset.
+        if self.instances is None:
+            res.pop("instances", None)
+        return res
```
jvstme: For backwards compatibility, please use excludes instead of overriding `dict()`

src/dstack/_internal/server/services/runs/plan.py
```diff
@@ -138,7 +141,16 @@ async def get_job_plans(
             replica_group_name=replica_group_name,
         )
         if candidate_fleet_models is None:  # `dstack offer` path
```
jvstme: (nit) Consider removing the comment, since now this path is also used for regular runs, not just `dstack offer`

src/dstack/_internal/server/services/instances.py
```diff
@@ -375,6 +380,176 @@ def get_instance_ssh_private_keys(instance_model: InstanceModel) -> tuple[str, O
...
+    selector_conditions = _get_instance_selector_conditions(project, selectors)
+    if selector_conditions:
+        filters.append(or_(*selector_conditions))
+
+    stmt = (
+        select(InstanceModel)
+        .join(InstanceModel.fleet)
+        .join(FleetModel.project)
+        .where(*filters)
+        .options(contains_eager(InstanceModel.fleet).contains_eager(FleetModel.project))
```
jvstme: (nit) `.load_only()`?

src/dstack/_internal/server/services/instances.py
```diff
@@ -375,6 +380,176 @@ def get_instance_ssh_private_keys(instance_model: InstanceModel) -> tuple[str, O
...
+def _get_hostname_selector_condition(selector: InstanceHostnameSelector):
+    pattern = f"%{_escape_like(selector.hostname)}%"
+    return or_(
+        InstanceModel.job_provisioning_data.ilike(pattern, escape="\\"),
+        InstanceModel.remote_connection_info.ilike(pattern, escape="\\"),
+    )
+
+
+def _escape_like(value: str) -> str:
+    return value.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
```
jvstme: (nit) Consider using `Model.column.icontains(string, autoescape=True)` instead, which does the escaping automatically

src/dstack/_internal/server/services/instances.py
```diff
@@ -375,6 +380,176 @@ def get_instance_ssh_private_keys(instance_model: InstanceModel) -> tuple[str, O
...
+        ProjectModel.name == ref.project,
+    )
+
+
+def _get_hostname_selector_condition(selector: InstanceHostnameSelector):
+    pattern = f"%{_escape_like(selector.hostname)}%"
+    return or_(
+        InstanceModel.job_provisioning_data.ilike(pattern, escape="\\"),
+        InstanceModel.remote_connection_info.ilike(pattern, escape="\\"),
+    )
```
jvstme, line 552: (nit) Consider adding a comment that this is only an optimization, and the actual filtering is done in-memory in `instance_matches_selector`. Filtering a JSON column with `LIKE` just looks too wrong

src/dstack/_internal/server/services/instances.py
```diff
@@ -528,8 +703,9 @@ def get_shared_instances_with_offers(
             continue
         total_blocks = common_utils.get_or_error(instance.total_blocks)
         idle_blocks = total_blocks - instance.busy_blocks
-        min_blocks = total_blocks if multinode else 1
-        for blocks in range(min_blocks, total_blocks + 1):
+        # Multinode still needs one physical instance per node. The offer itself must be
+        # sized to the node requirements, so start from one block rather than the full host.
+        for blocks in range(1, total_blocks + 1):
```
jvstme: This breaks distributed tasks. `dstack` now runs distributes tasks on individual blocks with bridge network, preventing inter-node communication.

```shell
$ dstack ps -v
 NAME               BACKEND           RESOURCES                    INSTANCE TYPE    PRICE           STATUS   PROBES  SUBMITTED   ERROR 
 massive-fish-1                       -                                             -               running          3 mins ago        
    job=0           aws (eu-north-1)  cpu=1 mem=4GB disk=100GB     m7i.large (1/2)  $0.0151 (spot)  running          3 mins ago        
    job=1           aws (eu-north-1)  cpu=1 mem=4GB disk=100GB     m7i.large (1/2)  $0.0151 (spot)  running          3 mins ago
```

src/dstack/_internal/server/services/runs/plan.py
```diff
@@ -522,6 +552,109 @@ def get_instance_offers_in_fleet(
...
+    volumes: Optional[list[list[Volume]]] = None,
+    exclude_not_available: bool = False,
+) -> list[tuple[InstanceModel, InstanceOfferWithAvailability]]:
+    candidates: list[
+        tuple[
+            bool,
+            float,
+            float,
+            list[tuple[InstanceModel, InstanceOfferWithAvailability]],
+        ]
```
jvstme: (nit) Difficult to read. Consider using a dataclass or some other data structure with named fields

src/dstack/_internal/core/models/profiles.py
peterschmidt85: I checked this locally and `EntityReference` is not parsed from a string automatically here: without the pre-validator, `fleet: my-fleet` fails with `value is not a valid dict`. My understanding is that this helper is still needed for the string short form. Could you confirm if you had a different parsing path in mind?

jvstme: Yes, you are right

jvstme (approved):
Should be OK once the remaining comments ([1](https://github.com/dstackai/dstack/pull/3958#discussion_r3403381344), [2](https://github.com/dstackai/dstack/pull/3958#discussion_r3401753586)) are addressed

src/dstack/_internal/core/models/profiles.py
peterschmidt85: Addressed by removing the `ProfileParams.dict()` override and relying on the existing compatibility exclude path instead. Added coverage that unset `instances` is excluded via `get_profile_excludes()`, while configured `instances` is preserved.

jvstme (approved):

Issue #3959, opened 2026-06-12
[Feature]: Replicated Gateways


jvstme:
### Problem

A gateway can have only one stateful replica, with no option to transfer or share its state with other replicas.

Supporting multiple replicas would be beneficial for:

* High availability.
* Migrating the gateway between instances.

### Solution

Support replicated gateways.

Implementation plan:
- [x] Basic replication (`replicas` property, provisioning, API, CLI, UI)
- [x] Manual scaling
  - [x] Independent replica provisioning and termination (gateway replica pipeline)
  - [x] State sync mechanism
  - [x] Gateway apply plan
  - [x] `replicas` in-place update
- [ ] Fault tolerance
  - [x] Service/replica registration does not fail if some gateway replicas are unavailable
  - [ ] Service autoscaling works if some gateway replicas are unavailable
- [x] Replicated gateways with ACM certificates
- [x] Gateway replica events
- [ ] Debt
  - [x] (internal) Naming cleanup
  - [x] Fix registration latency
  - [ ] Test and permit more than 3 replicas per gateway
  - [ ] (?) Do not provision service before registration succeeds
  - [ ] (internal) Drop `GatewayModel.gateway_replica_id`

### Would you like to help us implement this feature by sending a PR?

Yes

PR #3960, opened 2026-06-12
Support gateways with multiple replicas

jvstme:
A gateway can now have multiple replicas for
improved availability.

```yaml
type: gateway
name: example-gateway

backend: aws
region: eu-west-1

domain: example.com

certificate: null
replicas: 2
```

To balance requests between gateway replicas, add
DNS records for each replica or set up a load
balancer outside of `dstack`. Replica hostnames
are displayed in `dstack` CLI and UI.

```shell
$ dstack gateway list
 NAME             BACKEND          HOSTNAME        DOMAIN       DEFAULT  STATUS
 example-gateway                                   example.com  ✓        running
    replica=0     aws (eu-west-1)  34.244.128.46
    replica=1     aws (eu-west-1)  18.201.201.174
```

Limitations:
- Changing the number of replicas or redeploying
  replicas is not supported.
- HTTPS is not supported. Use an external load
  balancer for TLS termination.
- An unavailable gateway replica prevents any new
  services or service replicas from being added.
- All replicas are bound to the same backend and
  region.
- At most 3 replicas are allowed per gateway.

Implementation notes:
- `GatewayComputeModel` now represents a gateway
  replica.
- In this version, the terms "compute" and
  "replica" are used interchangeably. The plan is
  to switch to using exclusively "replica" later.
- In this version, replica provisioning and
  termination are still done in the gateway
  pipeline, for all replicas at once. The plan is
  to introduce gateway replica pipelines later to
  allow for independent replica processing.
  
#3959


src/dstack/_internal/server/background/pipeline_tasks/gateways.py
```diff
@@ -303,16 +302,28 @@ async def _process_submitted_gateway(gateway_model: GatewayModel) -> _SubmittedR
...
-            backend_compute=backend.compute(),
-            project_name=gateway_model.project.name,
-            configuration=configuration,
-            backend_id=backend_model.id,
-        )
+        for replica_num in range(replicas):
+            logger.debug(
+                "%s replica %d: creating gateway compute", fmt(gateway_model), replica_num
+            )
+            gateway_compute_model = await gateways_services.create_gateway_compute(
```
r4victor, line 314: What if one replica fails but others are provisioned – there needs to be a clean up of successfully provisioned replicas.

src/dstack/_internal/core/models/gateways.py
```diff
@@ -90,6 +92,13 @@ class GatewayConfiguration(CoreModel):
             " Set to `null` to disable. Defaults to `type: lets-encrypt`"
         ),
     ] = LetsEncryptGatewayCertificate()
+    replicas: Annotated[
```
r4victor, line 95: Let's put a technical upper bound on the number of provisioned replicas, e.g. 20 (to avoid provisioning 1000 replicas in a loop).

src/dstack/_internal/server/services/gateways/__init__.py
```diff
@@ -571,30 +583,71 @@ async def generate_gateway_name(session: AsyncSession, project: ProjectModel) ->
...
-        logger.warning(
-            "Failed to connect to gateway %s: %s", gateway.gateway_compute.ip_address, e
+        _, connections = await get_or_add_gateway_connections(session, gateway_id)
+    except GatewayError:
+        return None
+    per_replica: list[PerWindowStats] = []
+    for conn in connections:
+        stats = await conn.get_stats(project_name, run_name)
+        if stats is None:  # Stats not fetched yet
+            return None
```
r4victor, line 638: If any one replica goes unavailable, it breaks autoscaling. I expect this needs to be fixed for HA so worth adding a TODO/FIXME.

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
```diff
@@ -1197,35 +1197,40 @@ async def _register_service_replica(
...
-        raise GatewayError(repr(e))
-    except GatewayError as e:
-        if "already exists in service" in e.msg:
-            logger.warning(
-                (
-                    "%s: could not register replica in gateway: %s."
-                    " NOTE: if you just updated dstack from pre-0.19.25 to 0.19.25+,"
-                    " expect to see this warning once for every running service replica"
-                ),
+    for conn in connections:
```
r4victor, line 1200: What happens if some gateway replicas are registered and some are not – it seems the job won't be considered registered and won't be unregistered from the succeeded gateway replicas.  Same applies to service registration.

Not sure about the consequences.

src/dstack/_internal/server/background/pipeline_tasks/gateways.py
jvstme: The gateway will enter the `failed` status, but any successfully provisioned replicas will remain in the database and can be deleted with `dstack gateway delete`.

This is consistent with the existing handling of single-replica gateway provisioning failures. If dstack creates an instance for a gateway but later fails to connect to it, the instance is not cleaned up automatically and can only be removed along with the `failed` gateway using `dstack gateway delete`.

While this behavior may be counterintuitive and worth revisiting, I think it will be easier to address after gateway replica statuses and pipelines are introduced, which I plan to implement in the next iteration.

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
jvstme: Dangling service replicas are possible, but they shouldn't have any visible effect. Such replicas will be terminated by dstack (since registration failed on one of the gateway replicas), so successful gateway replicas won't be able to connect to them, and Nginx won't forward any requests to them. The dangling replicas will be removed from the gateway when the service is unregistered.

As for dangling services, they were already possible before if a service unregistration request failed. They are [cleaned up](https://github.com/dstackai/dstack/blob/ccfec5611f4cbff00cddeaa63c970ed9bbfc68da/src/dstack/_internal/server/services/services/__init__.py#L208) when a service with the same name is registered again.

This is worth addressing, but I think it can wait until the state synchronization mechanism is introduced, so we don't have to implement distributed transaction orchestration that will soon become obsolete

src/dstack/_internal/core/models/gateways.py
jvstme: Added a very conservative limit that has been tested and has a higher chance of working well. Added on the server side so that it can be revisited as the implementation improves

r4victor (approved):

Issue #3957, opened 2026-06-11
[Bug]: dstack server fails to start with `smg_grpc_proto` `RuntimeError` with grpcio<1.81


r4victor:
### Steps to reproduce

1. Install dstack in a fresh environment with any grpcio version <1.81:

```
uv pip install 'grpcio<1.81'
uv pip install 'dstack[all]==0.20.24' --refresh
```

2. `dstack server` fails with `RuntimeError: The grpc package installed is at version 1.80.0, but the generated code in mlx_engine_pb2_grpc.py depends on grpcio>=1.81.0.`

Apparently  `smg_grpc_proto` works only with the latest grpcio==1.81.0 version. 

### Expected behaviour

Figure out if older `grpcio` versions can be supported. If not, specify `grpcio>=1.81.0`.

### dstack version

master

### Server logs

```shell
File "/Users/r4victor/Projects/dstack/venvt/lib/python3.12/site-packages/dstack/_internal/server/services/runs/router_worker_sync.py", line 17, in <module>
    from smg_grpc_proto import (
  File "/Users/r4victor/Projects/dstack/venvt/lib/python3.12/site-packages/smg_grpc_proto/__init__.py", line 10, in <module>
    from smg_grpc_proto.generated import (
  File "/Users/r4victor/Projects/dstack/venvt/lib/python3.12/site-packages/smg_grpc_proto/generated/mlx_engine_pb2_grpc.py", line 21, in <module>
    raise RuntimeError(
RuntimeError: The grpc package installed is at version 1.80.0, but the generated code in mlx_engine_pb2_grpc.py depends on grpcio>=1.81.0. Please upgrade your grpc module to grpcio>=1.81.0 or downgrade your generated code using grpcio-tools<=1.80.0.
```

Bihan:
#3971  fixes the issue

PR #3971, opened 2026-06-16
Fix runtime error with grpcio

Bihan:
Fixes issue #3957 
Supporting `grpcio<1.81` would require pinning `smg-grpc-proto<=0.4.8`, since 0.4.9+ stubs enforce grpcio>=1.81 at import time. Pinning smg-grpc-proto<=0.4.8 keeps us on older pre-built stubs and blocks newer fixes, which is risky long-term. Therefore this PR prefers pinning `grpcio>=1.81.0` instead.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -200,7 +200,7 @@ server = [
     "asyncpg",
     "python-json-logger>=3.1.0",
     "prometheus-client",
-    "grpcio>=1.50",
+    "grpcio>=1.81.0",
     "protobuf>=6.33.5",
     "smg-grpc-proto>=0.4.7",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -202,7 +202,7 @@ server = [
     "prometheus-client",
     "grpcio>=1.81.0",
     "protobuf>=6.33.5",
-    "smg-grpc-proto>=0.4.7",
+    "smg-grpc-proto==0.4.9", # Pin to avoid grpcio version mismatch when smg-grpc-proto bumps grpcio dependency https://github.com/dstackai/dstack/pull/3971
 ]
 aws = [
     "boto3>=1.38.13",


r4victor (approved):

jvstme:
The latest `smg-grpc-proto==0.4.10` now implicitly depends on `grpcio>=1.81.1`, so the steps to reproduce from the issue are still relevant, just with `grpcio<1.81.1`.

```shell
$ rm -rf .venv
$ git checkout Bihan/fix_issue_3957_runtime_error_with_grpcio
$ uv venv
$ uv pip install "grpcio<1.81.1"
$ uv pip install .[all] --refresh
$ ./.venv/bin/dstack server
  File "/dstack/.venv/lib/python3.13/site-packages/smg_grpc_proto/generated/mlx_engine_pb2_grpc.py", line 21, in <module>
    raise RuntimeError(
    ...<5 lines>...
    )
RuntimeError: The grpc package installed is at version 1.81.0, but the generated code in mlx_engine_pb2_grpc.py depends on grpcio>=1.81.1. Please upgrade your grpc module to grpcio>=1.81.1 or downgrade your generated code using grpcio-tools<=1.81.0.
```

Looking at some of the previous `smg-grpc-proto` releases, it seems that they bump the implicit `grpcio` dependency quite often.

```shell
$ pip install smg-grpc-proto==0.4.10
$ sed -n 8p /tmp/venv/lib64/python3.14/site-packages/smg_grpc_proto/generated/common_pb2_grpc.py
GRPC_GENERATED_VERSION = '1.81.1'
$ pip install smg-grpc-proto==0.4.9
$ sed -n 8p /tmp/venv/lib64/python3.14/site-packages/smg_grpc_proto/generated/common_pb2_grpc.py
GRPC_GENERATED_VERSION = '1.81.0'
$ pip install smg-grpc-proto==0.4.8
$ sed -n 8p /tmp/venv/lib64/python3.14/site-packages/smg_grpc_proto/generated/common_pb2_grpc.py
GRPC_GENERATED_VERSION = '1.80.0'
$ pip install smg-grpc-proto==0.4.5
$ sed -n 8p /tmp/venv/lib64/python3.14/site-packages/smg_grpc_proto/generated/common_pb2_grpc.py
GRPC_GENERATED_VERSION = '1.78.0'
```

I assume this means that we will keep seeing the `RuntimeError` issue in the future, unless we pin `smg-grpc-proto`, or unless `smg-grpc-proto` can specify what `grpcio` version it needs in its package dependencies.

r4victor:
Seems restrictive to bump grpcio to the latest version constantly. How about making smg-grpc-proto an optional dep to limit the blast radius?

jvstme:
> making smg-grpc-proto an optional dep

I'm not sure, it will add an extra setup step for our users to get the SMG integration running, and users that need the integration will still have to deal with a potential version mismatch. If we add it to the `all` extra, then users that install `dstack[all]` will have to deal with it too.

I think I would prefer to restrict the `smg-grpc-proto` version for now.

And optionally ask its maintainers to address the issue on their side. I don't know if there are any established best practices for shipping gRPC SDKs as Python packages, but I assume the package could at least:
- Specify the required gRPC version in its dependencies.
- Be built against an older gRPC version, so as not to require all users to have the latest version.

> best practices for shipping gRPC SDKs

By the way, google packages don't seem to have that runtime version check.

Bihan:
So for now I will update the PR with

```
smg-grpc-proto==0.4.9
grpcio>=1.81.0
```

jvstme (approved):

PR #3986, opened 2026-06-24
Optimize /api/runs/list job loading

### LINKED ISSUE #3983: [Bug]: /api/runs/list loads all job submissions before applying job_submissions_limit
### Steps to reproduce

1. On Postgres, create a project with 20 recent runs where each run has many historical job submissions, e.g. 3000 `jobs` rows per run for the same `(run_id, replica_num, job_num)` with increasing `submission_num`. Optionally add probes.
2. Call the Runs UI-shaped endpoint:

```json
POST /api/runs/list
{
  "limit": 20,
  "job_submissions_limit": 1,
  "only_active": false
}
```

The CLI/public API has the same issue, usually with `project_name`, `limit: 100`, and `job_submissions_limit: 1`.

### Actual behaviour

The response contains only the latest submission, e.g. `20 runs / 20 jobs / 20 submissions`, but the backend first eager-loads all historical job rows via `selectinload(RunModel.jobs).joinedload(JobModel.probes)` and only then applies `job_submissions_limit` in Python.

Local synthetic repro with 20 runs and 3000 historical submissions per run:

- Normal UI-shaped request: avg `~2.9s`, max `~3.3s`
- Admin UI-shaped request: avg `~3.0s`, max `~3.2s`
- `include_jobs=false` still takes `~2.6-2.8s`
- Service breakdown: SQL execute `~0.36s`; ORM load/materialization `~2.1-2.8s`; conversion `~0.11s`; JSON serialization `~0.002s`

There is also an admin scalability issue: admin requests load all visible projects and then query runs with `project_id IN (<all project ids>)`. With 30k extra projects this added `~0.7s` locally; with 50k projects it failed with asyncpg: `the number of query arguments cannot exceed 32767`.

### Expected behaviour

- `job_submissions_limit` should limit job rows loaded from the database, not only submissions returned in the response.
- `include_jobs=false` should not load jobs.
- Admin run listing should not build a giant `project_id IN (...)`; if `project_name` is provided, query that project directly; if admin and no `project_name`, avoid the project ID filter; for non-admin users use a join/subquery through `members`.

### dstack version

master

### Server logs

```shell

```

### Additional information

Relevant code path: `src/dstack/_internal/server/services/runs/__init__.py`.

peterschmidt85:
Fixes #3983

## Problem

`/api/runs/list` powers both the Runs UI and `dstack ps`. UI/CLI list views normally need only the latest job submission, but the server still eager-loaded the full `RunModel.jobs` history and trimmed it later during response serialization.

On runs with many retries/submissions, one page could materialize tens of thousands of `JobModel` rows. Admin/all-project requests also had extra overhead because the service loaded accessible projects into Python and filtered runs with a large `project_id IN (...)` list.

## Fix

- Stop eager-loading `RunModel.jobs` in the runs-list path.
- After selecting the requested page of runs, load only the job rows needed for the response.
- When `job_submissions_limit` is set, load the latest N submissions per `(run_id, replica_num, job_num)` with a SQL window query.
- Also load the latest terminated submission with a termination reason per logical job, so status messages such as `retrying` stay correct without loading full history.
- Replace Python-built project ID lists with direct project lookup/subqueries for single-project, admin all-project, and member all-project paths.
- Ensure lightweight UI/CLI list callers pass `job_submissions_limit=1`.

### DIFF
--- a/src/dstack/_internal/server/services/runs/__init__.py
+++ b/src/dstack/_internal/server/services/runs/__init__.py
@@ -52,6 +52,7 @@ from dstack._internal.server.models import (
     UserModel,
 )
 from dstack._internal.server.services import events, services
+from dstack._internal.server.services import projects as projects_services
 from dstack._internal.server.services import repos as repos_services
 from dstack._internal.server.services.jobs import (
     check_can_attach_job_volumes,
@@ -178,11 +179,13 @@ async def list_user_runs(
     repo = None
     project = None
     if project_name is not None:
-        project = await _get_project_model(
+        projects = await projects_services.list_user_project_models(
             session=session,
             user=user,
-            project_name=project_name,
+            only_names=True,
+            project_names=[project_name],
         )
+        project = next(iter(projects), None)
         if project is None:
             return []
         if repo_id is not None:
@@ -229,31 +232,6 @@ async def list_user_runs(
     if len(run_models) > len(runs):
         logger.debug("Can't load %s runs", len(run_models) - len(runs))
     return runs
-
-
-async def _get_project_model(
-    session: AsyncSession,
-    user: UserModel,
-    project_name: str,
-) -> Optional[ProjectModel]:
-    """
-    Resolve project_name and check user access for the runs list.
-
-    This avoids loading project relations that are not used by runs list responses.
-    """
-    filters = [
-        ProjectModel.name == project_name,
-        ProjectModel.deleted == False,
-    ]
-    if user.global_role != GlobalRole.ADMIN:
-        filters.extend(
-            [
-                MemberModel.project_id == ProjectModel.id,
-                MemberModel.user_id == user.id,
-            ]
-        )
-    res = await session.execute(select(ProjectModel).where(*filters))
-    return res.scalar()
 
 
 async def list_projects_run_models(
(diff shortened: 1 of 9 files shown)


r4victor (approved):

src/dstack/_internal/server/services/runs/__init__.py
```diff
@@ -220,9 +231,35 @@ async def list_user_runs(
...
+    ]
+    if user.global_role != GlobalRole.ADMIN:
+        filters.extend(
+            [
+                MemberModel.project_id == ProjectModel.id,
+                MemberModel.user_id == user.id,
+            ]
+        )
+    res = await session.execute(select(ProjectModel).where(*filters))
+    return res.scalar()
```
jvstme: (nit) `dstack._internal.server.services.projects` would be a more suitable place for this function.

I also think I would prefer to add the `project_names` filter to the existing `list_user_project_models` function instead of adding this new function and duplicating user authorization logic in its implementation.

src/dstack/_internal/server/services/runs/__init__.py
```diff
@@ -271,16 +328,179 @@ async def list_projects_run_models(
...
+    run_ids: List[uuid.UUID],
+) -> List[JobModel]:
+    """List jobs with termination reasons used to compute run.status_message."""
+    return await _list_latest_job_models_per_job(
+        session=session,
+        run_ids=run_ids,
+        limit_per_job=1,
+        include_probes=False,
+        only_with_termination_reason=True,
+    )
```
jvstme: (nit) These two functions look somewhat redundant to me

jvstme (approved):

jvstme (approved):

PR #4002, opened 2026-07-03
Fix/ebs device resolve retry

### LINKED ISSUE #4001: [Bug]: shim occasionally fails to detect AWS EBS Volumes due to race condition
### Steps to reproduce

- start a dev environment in AWS with a volume attachment
- it may fail right away, if not, detach without terminating and reapply the run
- do this several times until failure

### Actual behaviour

Runs fail with "prepareVolumes error: format and mount volume: get real device name: volume <volume-id> not found among block devices"

### Expected behaviour

runs are successful without failure

### dstack version

0.20.26

### Server logs

```shell

```

### Additional information

Caused by a race condition. shim is trying to get real device name before volume is ready

james-boydell:
Fixes #4001 
`GetRealDeviceName` ran `lsblk` once and hard-failed with "volume not found among block devices" if the device wasn't there yet. But `AttachVolume` returning success doesn't mean the guest kernel has enumerated the NVMe device, so resolving right after attach intermittently failed. This polls for the device (30s timeout, 500ms interval) instead of failing on the first miss. Threads `ctx` through the backend interface so retries log to `shim.log`.

### DIFF
--- a/runner/internal/shim/backends/aws.go
+++ b/runner/internal/shim/backends/aws.go
@@ -8,6 +8,8 @@ import (
 	"os/exec"
 	"strings"
 	"time"
+
+	"github.com/dstackai/dstack/runner/internal/common/log"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


r4victor (approved):
Tested it fixes #4002 on AWS. Theoretically, the same bug is possible on other backends so ideally the retry logic should be outside of GetRealDeviceName, not inside. But currently we support only GCP shim-attached volumes besides AWS, and it's not reproducible on GCP. Otherwise the PR is good so merging it as is for now.

r4victor:
@james-boydell, thanks for the PR

PR #4006, opened 2026-07-04
Retry NVIDIA containers without display capability

### LINKED ISSUE #4004: [Bug]: NVIDIA tasks fail on headless hosts without /dev/nvidia-modeset
### Steps to reproduce

Run an NVIDIA task on a headless host where CUDA works but `/dev/nvidia-modeset` is not present.

Example CloudRift host where this reproduced:

```text
region: ap-east-tw-kn-1
instance type: rtx49-10c-kn.1
GPU: NVIDIA GeForce RTX 4090
driver: 575.57.08
```

The host has the usual NVIDIA compute devices:

```shell
ls /dev/nvidia0 /dev/nvidiactl /dev/nvidia-uvm /dev/nvidia-uvm-tools
ls /dev/nvidia-caps
```

It does not have `/dev/nvidia-modeset`:

```shell
ls /dev/nvidia-modeset
```

Plain Docker CUDA works:

```shell
docker run --rm --gpus all nvidia/cuda:12.4.1-base-ubuntu22.04 nvidia-smi -L
```

The same host fails when Docker is asked for the NVIDIA capabilities currently requested by `dstack-shim`:

```shell
docker run --rm \
  --gpus 'all,"capabilities=gpu,utility,compute,graphics,video,display,compat32"' \
  nvidia/cuda:12.4.1-base-ubuntu22.04 \
  nvidia-smi -L
```

### Actual behaviour

The container fails before the user command starts:

```text
nvidia-container-cli: mount error: stat failed: /dev/nvidia-modeset: no such file or directory
```

This prevents normal CUDA workloads from running, even though the host can run `nvidia-smi` and CUDA containers when `display` is not requested.

### Expected behaviour

CUDA/PyTorch workloads should run on NVIDIA compute hosts that do not expose display devices.

`display` should not be required for ordinary compute workloads.

### dstack version

master

### Server logs

```shell
nvidia-container-cli: mount error: stat failed: /dev/nvidia-modeset: no such file or directory
```

### Additional information

`dstack-shim` currently requests this Docker `container.DeviceRequest.Capabilities` set for NVIDIA GPUs:

```text
gpu,utility,compute,graphics,video,display,compat32
```

NVIDIA documents `display` as X11 display support. The documented default driver capabilities are `utility,compute`: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/docker-specialized.html

A safe fix is to keep the current capability set on the first attempt, then retry without only `display` if NVIDIA Container Runtime fails specifically because `/dev/nvidia-modeset` is missing. This should not apply to non-NVIDIA vendors or explicit `GPUDevices` handling.

peterschmidt85:
Fixes #4004.

Currently, `dstack-shim` asks Docker for this NVIDIA `container.DeviceRequest.Capabilities` set:

```text
gpu,utility,compute,graphics,video,display,compat32
```

This is wrong on headless NVIDIA hosts where CUDA works but `/dev/nvidia-modeset` is absent. Requesting `display` makes NVIDIA Container Runtime fail before the user command starts:

```text
nvidia-container-cli: mount error: stat failed: /dev/nvidia-modeset: no such file or directory
```

The fix keeps the current full capability set on the first start attempt. If an NVIDIA container fails specifically because `/dev/nvidia-modeset` is missing, the shim removes the failed container and retries without only `display`:

```text
gpu,utility,compute,graphics,video,compat32
```

NVIDIA documents `display` as X11 display support. The documented default driver capabilities are `utility,compute`: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/docker-specialized.html

The fallback applies only to the normal NVIDIA Docker `DeviceRequest` path. AMD, Tenstorrent, Intel, and explicit `GPUDevices` handling are unchanged.

AI Assistance: This PR was prepared with AI assistance.

### DIFF
--- a/runner/internal/shim/docker.go
+++ b/runner/internal/shim/docker.go
@@ -46,7 +46,13 @@ const (
 	LabelKeyIsTask = LabelKeyPrefix + "is-task"
 	LabelKeyTaskID = LabelKeyPrefix + "task-id"
 	LabelValueTrue = "true"
+
+	nvidiaModesetDevicePath = "/dev/nvidia-modeset"
 )
+
+type createContainerOptions struct {
+	disableNvidiaDisplayCapability bool
+}
 
 // dockerd reports pulling progress as a stream of JSON Lines. The format of records is not documented in the API documentation,
 // although it's occasionally mentioned, e.g., https://docs.docker.com/reference/api/engine/version-history/#v148-api-changes
@@ -465,7 +471,7 @@ func (d *DockerRunner) Run(ctx context.Context, taskID string) error {
 	if err := d.tasks.Update(task); err != nil {
 		return fmt.Errorf("%w: failed to update task %s: %w", ErrInternal, task.ID, err)
 	}
-	if err := d.createContainer(ctx, &task); err != nil {
+	if err := d.createContainer(ctx, &task, createContainerOptions{}); err != nil {
 		errMessage := fmt.Sprintf("createContainer error: %s", err.Error())
 		log.Error(ctx, errMessage)
 		task.SetStatusTerminated(string(types.TerminationReasonCreatingContainerError), errMessage)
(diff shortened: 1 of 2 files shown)


peterschmidt85:
I prefer retrying after the NVIDIA runtime error rather than checking `/dev/nvidia-modeset` before container creation.

The real question is not just whether one host device file exists. The real question is whether NVIDIA Container Runtime can start this container with the requested capabilities on this host.

So the first attempt keeps the current capability set unchanged. If it works, behavior is unchanged. If NVIDIA Container Runtime fails specifically because `/dev/nvidia-modeset` is missing, then we know `display` caused the failure and retry without only `display`.

This makes the condition for changing behavior the actual runtime failure, not our pre-check guess about host devices.

un-def (approved):

PR #4014, opened 2026-07-08
Add in-place update for gateway `replicas`

jvstme:
Support scaling gateways via in-place update of
the `replicas` configuration property. For now,
this feature is only intended for internal testing
and is hidden behind the `GATEWAY_SCALING` flag,
since it has limited applicability at the current
stage — without a state sync mechanism, newly
started gateway replicas are not aware of existing
services.

#3959


src/dstack/_internal/server/background/pipeline_tasks/gateways.py
```diff
@@ -243,8 +269,7 @@ async def _process_submitted_item(item: GatewayPipelineItem):
     set_processed_update_map_fields(update_map)
     set_unlock_update_map_fields(update_map)
     async with get_session_ctx() as session:
-        for gateway_compute_model in result.gateway_compute_models:
-            session.add(gateway_compute_model)
+        await _apply_replica_scaling(session, result.scale_result)
```
r4victor: I think there is an issue with `_apply_replica_scaling()` being called before lock_token guard. It can lead to a stale worker committing new replicas without updating the gateway. E.g. a stale submitted gateway can create replicas. Then, it will be re-processed with duplicate replicas committed again. All related db changes / side effects should be done after the lock_token guard.

src/dstack/_internal/server/background/pipeline_tasks/gateways.py
r4victor, line 1: If a gateway replica never provisions (e.g. there is some cloud issue, backend availability, etc), the reconcile loop never terminates, re-creating replicas forever. We need some backoff logic and a limit. For fleets, we have `_MAX_CONSOLIDATION_ATTEMPTS` and `_CONSOLIDATION_RETRY_DELAYS`.

src/dstack/_internal/utils/common.py
```diff
@@ -309,6 +309,13 @@ def batched(seq: Iterable[T], n: int) -> Iterable[List[T]]:
     return iter(lambda: list(itertools.islice(it, n)), [])
 
 
+def get_lowest_unused_nums(used_nums: set[int]) -> Iterator[int]:
+    for num in itertools.count():
+        if num not in used_nums:
+            yield num
+    assert False, "unreachable"
```
r4victor: `itertools.count()` is infinite. why `assert False, "unreachable"`?

r4victor (changes requested):

src/dstack/_internal/utils/common.py
jvstme: It's an artifact of my previous implementation that `return`ed a single `int` and had to show the type checker that the path outside the loop never returns

r4victor (approved):

src/dstack/_internal/server/migrations/versions/2026/07_27_0552_d35995d56249_add_gateway_scaling_fields.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("gateway_computes", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("scale_in", sa.Boolean(), server_default=sa.false(), nullable=False)
+        )
+
+    with op.batch_alter_table("gateways", schema=None) as batch_op:
```
r4victor, line 23: Altering multiple tables should be done in separate transactions (migrations). See https://github.com/dstackai/dstack/blob/2f9618f4d521140350efd1b344412d122c1e0322/contributing/MIGRATIONS.md?plain=1#L47

r4victor (approved):

PR #4023, opened 2026-07-11
[Feature] Supporting dstack-inside-dstack for runs

peterschmidt85:
## Problem

Runs may need to call dstack themselves, for example to inspect runs, submit
another run, or attach to it. This currently requires the dstack server to
accept direct connections from the run.

## UX

Add `dstack: true` to a run configuration.

```yaml
type: task
image: dstackai/dstack
dstack: true
env:
  - DSTACK_TOKEN
commands:
  - dstack ps
```

With `dstack: true`, dstack exposes the server managing the run through
`/run/dstack/server.sock` inside each job. It sets `DSTACK_SERVER_URL` to the
corresponding `http+unix` URL and `DSTACK_PROJECT` to the current project,
allowing the CLI and API clients to use that connection.

Authentication remains explicit. dstack does not inject `DSTACK_TOKEN`; it can
be passed through the existing environment or secret mechanisms.

The CLI also accepts `DSTACK_SERVER_URL`, `DSTACK_PROJECT`, and `DSTACK_TOKEN`
as a complete configuration without requiring `~/.dstack/config.yml`.

`dstack` cannot be combined with `inactivity_duration` on dev environments,
since the persistent server connection counts as activity.

## Implementation

Before starting commands, the server opens an SSH connection to each job and
creates a remote Unix-socket forward. Connections to
`/run/dstack/server.sock` inside the job are forwarded to the HTTP port of the
server managing the run. `DSTACK_SERVER_URL` points clients to this socket.

This uses the existing job SSH path, so the server does not need a public URL
or gateway. Health checks use the same socket path end to end; if the SSH
connection or socket becomes stale, the server closes and recreates the
tunnel. The tunnel is removed when the job terminates.

With multiple server replicas, a replica probes the socket and adopts an
existing healthy one instead of overwriting it, so jobs keep a single stable
server owner.


src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
```diff
@@ -474,6 +477,11 @@ async def _process_running_job(context: _ProcessContext) -> _ProcessResult:
             context=context, startup_context=startup_context, result=result
         )
     elif context.job_model.status == JobStatus.RUNNING:
+        if _server_access_enabled(context):
+            await job_server_connections_pool.ensure(
+                context.job_model,
+                context.job_submission.job_runtime_data,
+            )
```
r4victor, line 484: If the tunnel establishment fails, this records `_failure_started_at`, but `_failure_started_at` then cleared immediately by `await job_server_connections_pool.remove(context.job_model.id)`. The job will be stuck in provisioning if there is something wrong with a connection and `job_server_connections_pool.retry_timed_out()` won't take any effect.

src/dstack/_internal/core/services/api_client.py
```diff
@@ -1,11 +1,23 @@
...
+)
 from dstack._internal.core.errors import ConfigurationError
 from dstack.api.server import APIClient
 
 
 def get_api_client(project_name: Optional[str] = None) -> Tuple[APIClient, str]:
+    env_project_name = project_name or os.getenv(DSTACK_PROJECT_ENV)
+    server_url = os.getenv(DSTACK_SERVER_URL_ENV)
+    token = os.getenv(DSTACK_TOKEN_ENV)
+    if env_project_name is not None and server_url is not None and token is not None:
```
un-def: Is it intentional to ignore these variables unless all three are set? E.g., sometimes I need to override the auth token per CLI call.

mkdocs/docs/concepts/dev-environments.md
```diff
@@ -642,6 +642,26 @@ The `schedule` property can be combined with `max_duration` or `utilization_poli
...
+Set `server` to `true` when a dev environment needs to use the dstack CLI. dstack configures the
+server and current project automatically. To run authenticated commands, pass `DSTACK_TOKEN`
+explicitly.
+
+<div editor-title=".dstack.yml">
+
+```yaml
+type: dev-environment
+image: dstackai/dstack
+server: true
```
un-def: `server_socket: true`? `server` is too vague, IMO

peterschmidt85: Agree, please share variants!

How about dstack: true - similar to Docker in Docker (docker: true)

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
peterschmidt85: Fixed in 8d0dfd4 — `remove()` now runs only when the job leaves RUNNING, so the failure time survives and `retry_timed_out()` fires. Regression test added.

Verified live: broken tunnel → job fails with "Could not establish dstack server access" after ~2 min (before: retried forever, until the runner's own 5-min watchdog killed it with a misleading `CONTAINER_EXITED_WITH_ERROR`).

src/dstack/_internal/core/services/api_client.py
peterschmidt85: Fixed in 065c6ab:
- `DSTACK_TOKEN` alone now overrides the configured token
- `DSTACK_SERVER_URL` alone is still ignored — it doubles as a server setting, and applying it alone would send the configured token to a different server
- incomplete env config with no config.yml → explicit error instead of silent fallback

mkdocs/docs/concepts/dev-environments.md
peterschmidt85: Renamed to `dstack: true` in 2c29b3a (mirrors `docker: true`). Docs, PR description, and tests updated; no migration needed — the field only lives in run spec JSON.

src/dstack/_internal/server/services/jobs/server_connection.py
```diff
@@ -0,0 +1,221 @@
...
+            shutil.rmtree(CONNECTIONS_DIR / str(job_id), ignore_errors=True)
+
+    async def remove_all(self) -> None:
+        async with self._pool_lock:
+            job_ids = set(self._connections).union(self._failure_started_at)
+        await asyncio.gather(*(self.remove(job_id) for job_id in job_ids))
+
+    async def _get_lock(self, job_id: uuid.UUID) -> asyncio.Lock:
+        async with self._pool_lock:
+            return self._locks.setdefault(job_id, asyncio.Lock())
```
jvstme: (nit) Is `_pool_lock` necessary? It only covers these two expressions that can't race since they are synchronous (no `await`)

src/dstack/_internal/server/services/jobs/server_connection.py
```diff
@@ -0,0 +1,221 @@
...
+            ]
+            # Probe through the job socket itself: a socket path can remain after its listener
+            # becomes unreachable.
+            self._tunnel.forwarded_sockets = [
+                SocketPair(
+                    local=UnixSocket(path=self._probe_socket_path),
+                    remote=UnixSocket(path=_REMOTE_SOCKET_PATH),
+                )
+            ]
+            await self._tunnel.aopen()
```
jvstme: So if there are multiple `dstack` server replicas, each replica will overwrite `/run/dstack/server.sock` once (at `JobServerConnectionsPool.ensure`), and then the socket will point to the replica that overwrote it last?

I anticipate a few issues with that design:

1. Disproportionate distribution of requests among `dstack` server replicas. For example, the server replica that starts last (e.g., as a result of scaling or rolling update) will gradually overtake `/run/dstack/server.sock` in all existing jobs, and all requests from all jobs will be handled exclusively by that single replica.
2. Worse availability.
   1. If the server replica behind `/run/dstack/server.sock` stops or fails, the job will be unable to request the server for some period of time (potentially longer than [30s](https://github.com/dstackai/dstack/blob/74b87ed9db4357b4cb43bdcbb9caeb654389d75d/src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py#L142)  if the job was locked) until another server replica detects the missing or unhealthy `/run/dstack/server.sock` and takes over.
   2. The server will also be unavailable for jobs during the overwrite — requests issued between `rm -f {remote_socket}` and `self._tunnel.aopen()` will fail.

Issue 2ii can be fixed with the existing design by not overwriting the socket if it already exists and is healthy.

Issues 1 and 2i would require revisiting the design. One alternative design would be to have each server replica create its own Unix socket in the job container (similar to [gateways](https://github.com/dstackai/dstack/blob/74b87ed9db4357b4cb43bdcbb9caeb654389d75d/src/dstack/_internal/server/services/gateways/connection.py#L90)), and have a load balancer in front of the sockets. For example, the runner could act as a load balancer.

src/dstack/_internal/server/services/jobs/server_connection.py
jvstme, line 1: Doesn't this connection pool [break](https://github.com/dstackai/dstack/blob/74b87ed9db4357b4cb43bdcbb9caeb654389d75d/src/dstack/_internal/server/services/runner/ssh.py#L63) `inactivity_duration` on container-based backends?

src/dstack/_internal/core/models/configurations.py
jvstme, line 1: Is there a reason not to support this new property for services? For example, it could be useful for agentic services that use `dstack` to start sandboxes, or for monitoring / visualization services that use `dstack` as a data source

src/dstack/_internal/server/services/jobs/server_connection.py
peterschmidt85: Good catch — dropped in e45abbf. `_get_lock` is now synchronous and the pool lock is gone; the guarded expressions are await-free so `setdefault` is already atomic under the single-threaded loop.

src/dstack/_internal/server/services/jobs/server_connection.py
peterschmidt85: Good catch — yes it does. The persistent server-access connection lands on the runner sshd port, and the tracker treats any connection lasting >10s as user activity, so `inactivity_duration` would never fire.

For now I reject the combination at config time (e428da5e6) rather than let inactivity silently break. Lifting the restriction needs a runner-side change — giving the server-access tunnel its own sshd port that the connection tracker ignores — which I'd do as a follow-up.

src/dstack/_internal/server/services/jobs/server_connection.py
peterschmidt85: Good catch — addressed in 42c8f6e84. Instead of unconditionally overwriting `/run/dstack/server.sock` on every open, a replica now probes it first and only (re)creates the reverse forward when it's missing or unreachable; otherwise it adopts the existing one. That keeps a single stable owner, so replicas no longer steal the socket from each other (issue 1) and there's no overwrite gap (issue 2ii). If the owner replica dies, another takes over on its next health check (issue 2i) — bounded, not instant.

Verified with two replicas on a shared Postgres and a real SSH fleet, with a job probing its socket once a second:
- Unfixed: start replica B, kill B's server → probe breaks (`200 → FAIL`), because B had stolen the socket.
- Fixed: same steps → probe holds at `200` (B adopted; replica A stays owner).
- Fixed failover: kill the owning replica entirely → the other takes over in ~33s (bounded by the 30s health-check throttle + tick).

This deliberately keeps the single-socket design and doesn't spread a single job's load across replicas or make failover instant — the per-replica-socket + load-balancer approach you sketched would, at the cost of a runner-side change. Happy to file that as a follow-up if we want those properties.

jvstme: > That keeps a single stable owner, so replicas no longer steal the socket from each other (issue 1).

To be precise, issue 1 is about disproportionate distribution of requests, and I think it is still relevant. Depending on server replica startup timing and job startup timing, a single server replica can still occupy `/run/dstack/server.sock` in all or most jobs.

I don't mind filing issues 1 and 2i as a follow-up.

src/dstack/_internal/core/models/configurations.py
peterschmidt85: Added in e69acf8b3 — `dstack: true` is now supported for services, same as tasks and dev environments (and excluded from run specs sent to older servers, like the others).

Each replica is a separate job that gets its own `/run/dstack/server.sock`, so multiple replicas behave like multi-node job replicas — no shared state between them.

Verified end-to-end with a `replicas: 2` service: both replicas got the socket + injected `DSTACK_SERVER_URL`/`DSTACK_PROJECT`, `dstack ps` worked over the socket, and each replica submitted an inner run via `dstack apply` that ran to completion. A service without `dstack` has no socket and no `DSTACK_SERVER_URL`, as expected.

One separate observation while testing: interactive `dstack attach` from *inside* a container (to another run) failed at the SSH-into-container step in my test rig. It's not specific to services or to `dstack: true` (same `SSHAttach` code as tasks; the submit/API path works fine), and it doesn't affect normal attach from a client. I'll dig into that separately.

src/dstack/_internal/server/services/jobs/server_connection.py
peterschmidt85: You're right, and thanks for the precision — issue 1 (even request distribution) is only partially addressed. Adopt stops the churn and gives a stable owner, but since the owner is whichever replica opens the socket first, startup timing can still concentrate most or all jobs on a single replica. So balanced distribution (issue 1) and instant failover (2i) remain open; the per-replica-socket + load-balancer design is the way to get them.

We'll create follow-up issues for 1 and 2i.

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
```diff
@@ -959,6 +976,36 @@ async def _process_running_status(
...
+            job_model=context.job_model,
+            job_update_map=result.job_update_map,
+            termination_reason=JobTerminationReason.TERMINATED_BY_SERVER,
+            termination_reason_message="Could not establish dstack server access",
+        )
+    return False
+
+
+def _server_access_enabled(context: _ProcessContext) -> bool:
+    return bool(getattr(context.run.run_spec.configuration, "dstack", False))
```
un-def: I'd recommend type narrowing with direct attribute access:

```python
if isinstance(context.run.run_spec.configuration, ConfigurationWithDstackParams):
    # task | dev-environment
    return context.run.run_spec.configuration.dstack
# service
return False
```

src/dstack/_internal/core/compatibility/runs.py
```diff
@@ -89,6 +93,15 @@ def get_run_spec_excludes(run_spec: RunSpec) -> IncludeExcludeDictType:
     if run_spec.configuration.backend_options is None:
         configuration_excludes["backend_options"] = True
 
+    if (
+        isinstance(
+            run_spec.configuration,
+            (DevEnvironmentConfiguration, TaskConfiguration, ServiceConfiguration),
+        )
```
jvstme: (nit) Redundant check

src/dstack/_internal/core/models/configurations.py
```diff
@@ -693,6 +693,18 @@ def check_image_or_commands_present(cls, values):
...
+class ConfigurationWithDstackParams(CoreModel):
+    dstack: Annotated[
+        bool,
+        Field(
+            description=(
+                "Make the dstack server accessible inside the run. "
+                "No authentication credentials are provided"
+            )
+        ),
+    ] = False
```
jvstme: (nit) Redundant mixin, can add the property directly to `BaseRunConfiguration`

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
un-def: Oh, I've missed e69acf8b327ca526834748372efdbdb9aa32b379. No need for configuration type check then, the field is present in all run configuration types.

src/dstack/_internal/core/compatibility/runs.py
peterschmidt85: Fixed in 1d0df9b9b — the `isinstance` guard was always true (every run configuration has `dstack`), so it is now just `if not run_spec.configuration.dstack`.

src/dstack/_internal/core/models/configurations.py
peterschmidt85: Fixed in 1d0df9b9b — moved the `dstack` field into `BaseRunConfiguration` and dropped the `ConfigurationWithDstackParams` mixin.

src/dstack/_internal/server/background/pipeline_tasks/jobs_running.py
jvstme: (nit) I assume the initial recommendation about direct attribute access is still relevant though

```suggestion
    return context.run.run_spec.configuration.dstack
```

jvstme (approved):
The run configuration option and the connection pool implementation look good to me. Haven't looked at the rest

src/dstack/_internal/server/services/jobs/server_connection.py
jvstme: Thanks. I think one issue can describe both cases, since the solution for both is the same.

I'd suggest to create it before resolving this thread so that we don't forget about it. Or let me know if you'd prefer me to create it

peterschmidt85: Created #4033 to track both balanced request distribution and fast failover for in-run server access. Thanks!

PR #4025, opened 2026-07-13
Vast.ai: Support spot instances

trashhalo:
Vast.ai offers interruptible (spot) instances, but the `vastai` backend has never been able to provision them.

Currently, `get_offers_by_requirements()` strips every spot offer:

```python
# TODO(egor-s): spots currently not supported
extra_filter=lambda offer: not offer.instance.resources.spot,
```

and `create_instance()` never bids, so only on-demand instances are provisioned. gpuhunt already emits Vast spot offers (each on-demand offer is duplicated with `price` set to the machine's `min_bid` and `spot=True`), so these offers reach dstack today and are then discarded.

Vast's create endpoint is the same for both instance types. `PUT /asks/{id}/` creates an **interruptible** instance when the payload includes a `price` (per-machine bid in \$/hour), and an **on-demand** instance when it is omitted. This matches the official `vast-python` client, whose `create instance --bid_price` sends the same `price` field to the same endpoint.

The fix:

1. Drop the `extra_filter` that removed spot offers in `get_offers_by_requirements()`. Spot vs. on-demand selection is already driven by the run's `spot_policy` through the catalog query, so no offers are surfaced that the user did not ask for.
2. In `run_job()`, bid the spot offer's price (which gpuhunt already set to `min_bid`) and pass it to `create_instance()`; on-demand offers pass no bid.
3. Add the `price` field to the `create_instance()` payload (`None` for on-demand preserves the existing behavior).

Interruption handling requires no backend changes: the server infers a reclaimed spot instance from the lost runner connection and, because the offer is marked `spot`, terminates the job with `INTERRUPTED_BY_NO_CAPACITY`, which maps to the generic `interruption` retry event. This is the same path RunPod, OCI, and Nebius spot instances already use.

Tested against a live Vast.ai account: spot offers carry a `min_bid` below `dph_base` (e.g. an RTX 4090 at \$0.24/hr on-demand vs. \$0.12/hr `min_bid`), and the create endpoint accepts the `price` field. Unit tests cover that `run_job()` bids on spot offers and does not bid on on-demand offers.

Vast.ai API reference: https://docs.vast.ai/api/create-instance

AI assistance: This PR was written primarily by Claude Code (investigation, implementation, and tests) and reviewed by me before submitting.

### DIFF
--- a/src/dstack/_internal/core/backends/vastai/api_client.py
+++ b/src/dstack/_internal/core/backends/vastai/api_client.py
@@ -41,6 +41,7 @@ class VastAIAPIClient:
         onstart: str,
         disk_size: int,
         registry_auth: Optional[RegistryAuth] = None,
+        bid: Optional[float] = None,
     ) -> dict:
         """
         Args:
@@ -49,6 +50,8 @@ class VastAIAPIClient:
             image_name: docker image name
             onstart: commands to run on start
             registry_auth: registry auth credentials for private images
+            bid: per-machine bid price in $/hour. If set, an interruptible
+                (spot) instance is created; if None, an on-demand instance.
 
         Raises:
             NoCapacityError: if instance cannot be created
@@ -63,6 +66,7 @@ class VastAIAPIClient:
         payload = {
             "client_id": "me",
             "image": image_name,
+            "price": bid,
             "disk": disk_size,
             "label": instance_name,
             "env": {
--- a/src/dstack/_internal/core/backends/vastai/compute.py
+++ b/src/dstack/_internal/core/backends/vastai/compute.py
@@ -94,8 +94,6 @@ class VastAICompute(
             backend=BackendType.VASTAI,
             locations=self.config.regions or None,
             requirements=requirements,
-            # TODO(egor-s): spots currently not supported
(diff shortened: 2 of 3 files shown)


jvstme (commented):
Hi @trashhalo. Thank you for contributing this useful feature. Overall, the PR looks good to me.

One issue I found during testing is that if you try to create an instance with an insufficient bid (for example, if someone outbids you just before you create the instance), Vast.ai returns an error but still creates an instance in the stopped state, which then continues to accumulate storage charges. This case requires additional cleanup.

I've already fixed this and added a few related improvements in 5e01c9c on `dstackai:vastai-spot`. If you agree with these changes, could you push the commits from `dstackai:vastai-spot` to `Hanno-Labs:vastai-spot` so they appear in this PR?

For some reason, GitHub won't let me push them to your repo myself, even though you've allowed edits from maintainers. This may be related to [this known GitHub limitation](https://github.com/orgs/community/discussions/5634).

jvstme:
We'd like to include this feature in this week's release, so I'll merge as-is and add my fixes in a separate PR. Thanks again for contributing

PR #4027, opened 2026-07-13
Feat: custom security groups

james-boydell:
Custom security groups / NSGs / firewall rules

Adds support for using a pre-existing security group instead of the one dstack creates automatically, so instances can run in private VPCs without an internet-exposed rule.

New security_group profile property (project-level default via backend config, overridable per fleet/run).
Supported on AWS (security_group_name/security_group_ids), Azure (network_security_group_names), and OCI (network_security_group_ids), each with per-region/location mapping. GCP gets a create_firewall_rules: false opt-out instead, since GCP firewall rules are VPC-wide.
When a custom security group is set, dstack never adds, removes, or modifies its rules — you're fully responsible for ingress/egress.
Docs updated in mkdocs/docs/concepts/backends.md.


mkdocs/docs/concepts/backends.md
```diff
@@ -242,6 +242,47 @@ There are two ways to configure AWS: using an access key or using the default cr
     Using private subnets assumes that both the `dstack` server and users can access the configured VPC's private subnets.
     Additionally, private subnets must have outbound internet connectivity provided by NAT Gateway, Transit Gateway, or other mechanism.
 
+??? info "Custom security group"
+    By default, `dstack` creates and manages its own security group per project (opening SSH to `0.0.0.0/0`
+    and allowing all traffic within the group so multi-node clusters work out of the box).
```
r4victor: Please try to use the same wording across backends for consistency (with backend-specific terminology) . Something like this:

By default, `dstack` creates and manages its own security group per project that allows: SSH ingress (TCP port 22) from `0.0.0.0/0`, unrestricted egress, and all traffic within the group so that multi-node clusters work out of the box.

r4victor:
@james-boydell, by default `dstack` indeed creates security groups / firewalls that allows SSH ingress from any IP. The public access can be disabled if you provision instances without public IPs – so security groups are not necessary to limit the public access – just to be clear.

I understand that security groups are complimentary to "no public ip" and can be useful to limit the connections from entire VPCs/peered VPCs to specific IPs, control the blast radius, meet the compliance etc. So PR makes sense 100%.

Can you expand why custom security groups are needed for your particular use case?

mkdocs/docs/concepts/backends.md
james-boydell: Made an attempt to update the wording/phrasing. Feel free to modify as you see fit!

james-boydell:
> @james-boydell, by default `dstack` indeed creates security groups / firewalls that allows SSH ingress from any IP. The public access can be disabled if you provision instances without public IPs – so security groups are not necessary to limit the public access – just to be clear.
> 
> I understand that security groups are complimentary to "no public ip" and can be useful to limit the connections from entire VPCs/peered VPCs to specific IPs, control the blast radius, meet the compliance etc. So PR makes sense 100%.
> 
> Can you expand why custom security groups are needed for your particular use case?

The immediate need is to meet internal compliance. It's understood that using instances without public IPs limits public access which has allowed us to make an internal exception for the time being. However we'd like to clean this up and comply with principle of least privilege and minimize the mute lists for our internal scanners.

peterschmidt85:
Thank you @james-boydell, for the PR and the details. We're looking into it, and will get back to you on it tomorrow!

r4victor:
@james-boydell, there seems to be multiple issues with run/fleet/profile-level security_group property. E.g. a run-level security_group is silently ignored for idle instances; a security_group set only in .dstack/profiles.yml may be ignored (I can go into more details if you insist to proceed in that direction.)

But more importantly, having this property on a run/fleet level also seems like an anti-security features since regular non-admin users could override security policies that admins need to enforce.

**I suggest implementing custom security groups / firewalls only on the backend configuration level**. This simplifies the PR significantly, follows the admin/user separation, and you don't need to fix security_group not being applied in some complicated cases.

src/dstack/_internal/core/backends/aws/configurator.py
```diff
@@ -146,6 +147,19 @@ def _check_config_iam_instance_profile(
                 f"Failed to check IAM instance profile {config.iam_instance_profile}"
             )
 
+    def _check_config_security_group(self, config: AWSBackendConfigWithCreds):
```
r4victor, line 150: Worth checking security group existence in configurator (as opposed to provisioning-time only) so that users get early error feedback, similar to vpc/subnet checks.

src/dstack/_internal/core/backends/oci/resources.py
```diff
@@ -556,10 +556,47 @@ def get_or_create_vcn(
...
+            # security list attached. Since dstack owns this subnet
+            # (it is not user-supplied), it's safe to update it in place -
+            # detaching the security list so the NSG becomes the sole
+            # security boundary for every instance in it, matching newly
+            # created subnets. This is unrelated to dstack's "fully
+            # hands-off" contract for user-supplied *security groups*, which
+            # this does not touch.
+            subnet = client.update_subnet(
+                subnet.id, oci.core.models.UpdateSubnetDetails(security_list_ids=[])
+            ).data
```
r4victor, line 599: Removing security lists from the subnet breaks ssh/egress for existing OCI instances on backend config apply. Custom NSGs are created only in create_instance().

Overall, the OCI approach seems broken since it requires users to create custom security groups in dstack-managed VCN. Custom VCNs should precede custom NSGs.

**I suggest dropping OCI implementation from this PR.**

src/dstack/_internal/core/backends/aws/compute.py
```diff
@@ -329,12 +331,27 @@ def create_instance(
...
-                ec2_client=ec2_client,
-                region=instance_offer.region,
-                project_id=project_name,
-                vpc_id=vpc_id,
-            )
+            security_group_id = instance_config.security_group
+            if security_group_id is None and self.config.security_group_ids is not None:
+                security_group_id = self.config.security_group_ids.get(instance_offer.region)
+            if security_group_id is None and self.config.security_group_name is not None:
+                security_group_id = aws_resources.get_security_group_id_by_name(
```
r4victor, line 338: getting security group should be cached like self._create_security_group and other methods – otherwise every AZ attempt re-issues an identical describe_security_groups call.

src/dstack/_internal/core/backends/azure/configurator.py
```diff
@@ -282,6 +284,27 @@ def _check_config_vpc(
                     except BackendError as e:
                         raise ServerClientError(e.args[0])
 
+    def _check_config_network_security_groups(self, config: AzureBackendConfigWithCreds):
+        if not config.network_security_group_names:
+            return
+        # When `regions` is None, all regions are used and there is no feasible way to validate
```
r4victor, line 290: If regions is None, it defaults to `LOCATIONS`, so the check is feasible (AWS does the check against the default region list).

src/tests/_internal/core/backends/aws/test_compute.py
r4victor, line 1: I'd drop all these test_compute since these tests are very ad-hoc and won't be very useful after the PR lands.

PR #4043, opened 2026-07-17
CLI: add `--no-profile` option

un-def:
Disables default profile loading. Mutually exclusive with `--profile`.

Supported by the following commands:
* `dstack apply` (run configurations only)
* `dstack preset create|apply`
* `dstack offer`

### DIFF
--- a/src/dstack/_internal/cli/commands/endpoint.py
+++ b/src/dstack/_internal/cli/commands/endpoint.py
@@ -15,13 +15,16 @@ from dstack._internal.cli.services.endpoints.store import (
     EndpointPresetStore,
     load_endpoint_configuration,
 )
-from dstack._internal.cli.services.profile import apply_profile_args, register_profile_args
+from dstack._internal.cli.services.profile import (
+    apply_profile_args,
+    load_profile_from_args,
+    register_profile_args,
+)
 from dstack._internal.cli.utils.common import confirm_ask, console
 from dstack._internal.core.errors import CLIError
 from dstack._internal.core.models.profiles import ProfileParams
 from dstack._internal.core.services import is_valid_dstack_resource_name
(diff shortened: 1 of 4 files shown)


peterschmidt85 (approved):
Look good to me from UX point of view

But also think of supporting `NO_DSTACK_PROFILE=1` for consistency?

un-def:
> But also think of supporting NO_DSTACK_PROFILE=1 for consistency?

A bit tricky in terms of precedence, I landed on:

`--no-profile` > `--profile=name` > `DSTACK_NO_PROFILE=1` > `DSTACK_PROFILE=name`


That is, 

`dstack offer` → default profile loaded
`DSTACK_NO_PROFILE=1 dstack offer` →  no profile loaded
`DSTACK_NO_PROFILE=1 DSTACK_PROFILE=my-profile dstack offer` → no profile loaded
`DSTACK_NO_PROFILE=1 dstack offer --profile my-profile` →  `my-profile` loaded
`DSTACK_NO_PROFILE=1 DSTACK_PROFILE=other-profile dstack offer --profile my-profile` →  `my-profile` loaded
`DSTACK_PROFILE=my-profile dstack offer --no-profile` →  no profile loaded

peterschmidt85:
> A bit tricky in terms of precedence, I landed on:

Can be made mutually exclusive?

un-def:
> Can be made mutually exclusive?

`--profile` and `--no-profile` are mutually exclusive, `DSTACK_PROFILE` and `DSTACK_NO_PROFILE` are not – unlike the command line, it's possible that the process environment is not in user's full control, e.g., they can set but cannot unset a variable, so, I'd rather not make them mutually exclusive unless there is a strong reason to do so.

PR #4059, opened 2026-07-22
Delete dull job and instance models created on no capacity retry

### LINKED ISSUE #4054: [Bug]: Slow `/api/runs/list` due to the large number of job submissions
### Steps to reproduce

1. Submit a run with a long retry (e.g. `retry.duration: 7d`) that never provisions.
2. The run accumulates lots of job submissions for every retry attempt (pending->submitted transitions).
3. Calling `/api/runs/list` (via the Python/HTTP API) is slow due to the large number of job submissions. Moreover, the entire server can become slow since the response serialization blocks the event loop.

### Actual behaviour

Every run retry attempt creates a job submission. Retry is exponential but the backoff is capped at 10 minutes. So a run with retry configured for a long duration accumulates submissions at ~6/hour ≈ 144/day – a thousand of submissions over a week of retry.

The CLI / UI pass `job_submissions_limit=1` to `/api/runs/list`. but users using API directly may request all job submissions.

### Expected behaviour

Possible solutions:
* Enforce job_submissions_limit server-side, e.g.  job_submissions_limit <=10. This fixes `/api/runs/list` issues, but runs can still have lots of job submissions produced and stored in the DB.
* Avoid creating JobModel for every retry attempt – need to figure out how to do it without losing the useful information about retry attempt time and reason.

### dstack version

master

### Server logs

```shell

```

### Additional information

_No response_

r4victor:
Fixes #4054

Prevent the accumulation of large number of job and instance models when run retries on no capacity:

* Hard-delete placeholder instances when terminating jobs that FAILED_TO_START_DUE_TO_NO_CAPACITY.
* Hard-delete old FAILED_TO_START_DUE_TO_NO_CAPACITY job models when creating new job submissions (for pending runs and retrying failed replicas). Only the latest failed job is preserved to show retrying status correctly.
* ~~Add runs to events about potentially never-provisioned jobs so that events are displayed with within_runs filter.~~ Event targets now store entity_run_id so that within_runs filter works even for job event when job models are deleted. entity_run_id is backfilled for existing runs and jobs.

Note: the cleanup is for FAILED_TO_START_DUE_TO_NO_CAPACITY jobs that a) carry no useful information and b) created in large quantities on retry. Job models failed due to other reasons are not cleaned up.


src/dstack/_internal/server/migrations/versions/2026/07_22_1331_87d4312605e5_backfill_eventtargetmodel_entity_run_id.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    op.execute(
+        """
+        UPDATE event_targets SET entity_run_id = jobs.run_id
+        FROM jobs
+        WHERE jobs.id = event_targets.entity_id
+            AND event_targets.entity_type = 'job'
+            AND event_targets.entity_run_id IS NULL
+        """
+    )
```
jvstme, line 41: Enums like `entity_type` hold uppercase values

```shell
$ sqlite3 ~/.dstack/server/data/sqlite.db "SELECT DISTINCT entity_type FROM event_targets;"
FLEET
GATEWAY
INSTANCE
JOB
PROJECT
RUN
SECRET
USER
VOLUME
```

src/dstack/_internal/server/services/events.py
```diff
@@ -354,23 +362,7 @@ async def list_events(
             )
         )
```
jvstme, line 363: > Event targets now store entity_run_id so that within_runs filter works even for job event when job models are deleted

Should we also introduce `entity_fleet_id` to be able to see deleted placeholder instance events when using the `within_fleets` filter?

jvstme (approved):

src/dstack/_internal/server/services/events.py
r4victor: No, I think placeholder instance events are not relevant.

jvstme: Then should we stop emitting them? Right now there's a weird behavior where the `Instance created for job` event is emitted and shown in `within_fleets`. Then, when the instance is deleted, the event disappears from `within_fleets` (and doesn't disappear if you aren't using the filter).

Although if we stop emitting them, users will still see the placeholder instance in the fleet during provisioning attempts, but won't find any related events, which won't look expected either.

So maybe `entity_fleet_id` would be the most straightforward fix.

r4victor: @jvstme, Should jobs and runs reference `entity_fleet_id` or only fleets and instances should do that? Do you think we should tie `entity_fleet_id` semantics to the `within_fleets` filter specifically?

r4victor: Maybe if we'd named `entity_*` columns something filter-specific like `within_filter_*` (e.g. `within_filter_fleet_id`) the semantics of these columns were more clear?

jvstme: > Should jobs and runs reference `entity_fleet_id` or only fleets and instances should do that?

Only fleets and instances.

> Do you think we should tie `entity_fleet_id` semantics to the `within_fleets` filter specifically?

Well, now `entity_fleet_id` will only be used in the `within_fleets` implementation, but the general semantics is that `entity_fleet_id` connects the target to the fleet that target is part of.

> Maybe if we'd named `entity_*` columns something filter-specific like `within_filter_*` (e.g. `within_filter_fleet_id`) the semantics of these columns were more clear?

We may use them for purposes other than filtering in the future. For example, we could return `fleet_id` in the API, so that callers can see the instance's fleet ID directly in the `/api/events/list` response, without having to call `/api/project/{project_name}/instances/get`. Same thing for jobs and their `run_id`.

The semantics look rather clear to me: `entity_project_id` -> the target entity is part of that project (or is the project itself), `entity_fleet_id` -> the target entity is part of that fleet (or is the fleet itself). If there are any doubts, we can explain the semantics in docstrings.

Also note that regular users of the `events` module won't need to think about these fields at all, since setting them is incapsulated in `Target.from_model`.

PR #4060, opened 2026-07-22
Add `full_offers` option to `Compute.get_offers()`

un-def:
When set to `True`, instructs the Compute not to adjust offers by requirements. Ignored by all backends except for Kubernetes and Slurm, which return full allocatable node resources if `True` or a purely synthetic offer reflecting resources that would be actually requested if `False` (the old/current default behavior).

Exposed as `full_offers` request body argument in HTTP API and `--full-offers` flag in `dstack apply`/`dstack offer` commands.

### DIFF
--- a/src/dstack/_internal/core/backends/base/compute.py
+++ b/src/dstack/_internal/core/backends/base/compute.py
@@ -121,7 +121,7 @@ class Compute(ABC):
         if `full_offers` set to `True`, the method should not adjust offer's resources according to
         `requirements`. For most backends, this flag has no meaning, as they work with predefined
         provider offers (even configurable disk size reflects the actual disk created once the
-        instance is provisioned), but some backends such as Kubernetes and Slurm allocates flexible
+        instance is provisioned), but some backends such as Kubernetes and Slurm allocate flexible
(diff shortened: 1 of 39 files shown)


src/dstack/_internal/core/backends/base/compute.py
```diff
@@ -109,12 +109,23 @@ class Compute(ABC):
...
         """
         Returns offers with availability matching `requirements`.
         If the provider is added to gpuhunt, typically gets offers using
         `base.offers.get_catalog_offers()` and extends them with availability info.
         It is called from async code in executor. It can block on call but not between yields.
+
+        if `full_offers` set to `True`, the method should not adjust offer's resources according to
+        `requirements`. For most backends, this flag has no meaning, as they work with predefined
+        provider offers (even configurable disk size reflects the actual disk created once the
+        instance is provisioned), but some backends such as Kubernetes and Slurm allocates flexible
```
r4victor:
```suggestion
        instance is provisioned), but some backends such as Kubernetes and Slurm allocate flexible
```

r4victor (approved):

PR #4062, opened 2026-07-23
Implement within_fleets filter via entity_fleet_id

r4victor:
Show fleet events about deleted placeholder instances by using entity_fleet_id (see https://github.com/dstackai/dstack/pull/4059#discussion_r3635989107).

### DIFF
--- a/src/dstack/_internal/server/services/events.py
+++ b/src/dstack/_internal/server/services/events.py
@@ -77,6 +77,7 @@ class Target:
     id: uuid.UUID
     name: str
     run_id: Optional[uuid.UUID] = None
+    fleet_id: Optional[uuid.UUID] = None
 
     def __post_init__(self):
         if self.type == EventTargetType.USER and self.project_id is not None:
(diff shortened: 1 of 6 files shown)


src/dstack/_internal/server/services/events.py
```diff
@@ -119,11 +123,18 @@ def from_model(
                 name=model.name,
             )
         if isinstance(model, InstanceModel):
+            fleet_id = model.fleet_id
+            if fleet_id is None:
+                # Not-yet-flushed models may only have the fleet relationship set.
+                fleet = model.__dict__.get("fleet")
```
jvstme, line 129: (nit) Isn't this equivalent to `model.fleet`?

```python
>>> i = InstanceModel()
>>> i.__dict__.get("fleet") is None
True
>>> i.fleet is None
True
```

jvstme (approved):

src/dstack/_internal/server/services/events.py
r4victor: No, `i.fleet` may raise `sqlalchemy.exc.MissingGreenlet` if the attribute is not loaded. `i.__dict__.get("fleet")` would return `None`.

PR #4101, opened 2026-08-05
Distinguish router and non-router replicas

jvstme:
**Problem**: Non-router replicas in a service with
a router should not be registered on the gateway.
To determine whether a particular replica should
be registered, one has to parse its job spec and
run spec and see if it is a router replica or not.
This does not work well for the upcoming gateway
state sync mechanism, which needs to quickly fetch
all jobs to be registered from the database,
without parsing their specs.

**Solution**:
- Do **not** set `JobModel.registered` for
  non-router replicas in services with a router.
  That way, `JobModel.registered` can be reliably
  used to identify the jobs that should be
  registered on the gateway, without parsing the
  job spec.
- Since the router worker sync pipeline needs both
  router and non-router replicas with successful
  probes, introduce `JobModel.ready`, set for any
  replica once probes pass.

```shell
$ dstack event --within-run prefill-decode | grep -E "ready|register"
[2026-08-05 10:30:29] [job prefill-decode-0-0] Service replica ready to receive requests
[2026-08-05 10:30:29] [job prefill-decode-0-0] Service replica registered to receive requests
[2026-08-05 10:44:45] [job prefill-decode-0-1] Service replica ready to receive requests
[2026-08-05 10:45:00] [job prefill-decode-0-2] Service replica ready to receive requests
[2026-08-05 11:15:22] [job prefill-decode-0-0] Service replica unregistered from receiving requests
```

#3959


r4victor:
IIUC, the rolling deployment logic also needs to be updated to use "ready" instead of "registered" to mean if the replica is ready to receive service requests.

See https://github.com/dstackai/dstack/blob/c2484d694375a6afa652ff41838635a75e1e2458/src/dstack/_internal/server/services/runs/replicas.py#L93

and

https://github.com/dstackai/dstack/blob/c2484d694375a6afa652ff41838635a75e1e2458/src/dstack/_internal/server/services/runs/replicas.py#L80-L82

suspicious why the rolling deployment tests do not cover it.

src/dstack/_internal/server/services/proxy/repo.py
```diff
@@ -113,10 +113,6 @@ async def get_service(self, project_name: str, run_name: str) -> Optional[Servic
                 ssh_head_proxy = rci.ssh_proxy
                 ssh_head_proxy_private_key = get_or_error(rci.ssh_proxy_keys)[0].private
             job_spec = get_job_spec(job)
-            if router_group is not None and job_spec.replica_group != router_group.name:
-                # Strict router-only: when a router is configured, the proxy should only be aware
-                # of router replicas.
-                continue
```
r4victor, line 119: What about Dynamo workers? I suppose registered remains True for dynamo workers, which means you cannot rely on registered=True to filter router jobs. 

https://github.com/dstackai/dstack/pull/3868

jvstme: `registered` [doesn't depend](https://github.com/dstackai/dstack/pull/4101/changes#diff-f251c7b8a2e9febf4e59db0a393e1f37124f790afa33b138b69325a33f33f485R1199) on the router type, so Dynamo workers also have `registered=False` and are filtered out here

r4victor: I was looking at the migration that sets registered=False only for jobs that are in the "service_router_worker_sync" table

r4victor (approved):

PR #4119, opened 2026-08-10
Fix SSH connection directory race

### LINKED ISSUE #4095: [Bug]: SSHTunnelPool races mkdir against rmtree on a shared connection dir, flaking tests under -n auto
### Steps to reproduce

`SSHTunnelPool` creates and removes the same connection directory from different processes, and `mkdir(exist_ok=True)` is not atomic against a concurrent `rmtree`. Under `pytest -n auto` this fails `test_terminating_jobs.py::TestJobTerminatingWorker::test_terminates_job[sqlite]`.

The race is between two lines that operate on the same `conn_dir`:

- [`pool.py:317`](https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/server/services/runner/pool.py#L317) `conn_dir.mkdir(parents=True, exist_ok=True)` in `_resolve_conn_dir`
- [`pool.py:282`](https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/server/services/runner/pool.py#L282) `shutil.rmtree(self._real_conn_dir, ignore_errors=True)` in `close`

`exist_ok=True` does not make `mkdir` atomic. CPython swallows `FileExistsError` only if the path is still a directory when it re-checks:

```python
# pathlib.py
try:
    self._accessor.mkdir(self, mode)
except OSError:
    if not exist_ok or not self.is_dir():
        raise
```

So `mkdir` raises because the directory exists, `close()` removes it before `is_dir()` runs, and the error propagates.

To reproduce the race directly, run this against any directory:

```python
import shutil, threading
from pathlib import Path

target = Path("/tmp/conns/127.0.0.4:22,10999,10998")
target.parent.mkdir(parents=True, exist_ok=True)
stop, errors = threading.Event(), []

def maker():
    while not stop.is_set():
        try:
            target.mkdir(parents=True, exist_ok=True)
        except FileExistsError as e:
            errors.append(e); stop.set(); return

def remover():
    while not stop.is_set():
        shutil.rmtree(target, ignore_errors=True)

ts = [threading.Thread(target=maker) for _ in range(4)]
ts += [threading.Thread(target=remover) for _ in range(4)]
for t in ts: t.start()
stop.wait(20); stop.set()
for t in ts: t.join()
print(errors[0] if errors else "no error")
```

It fails within a second, every run.

Three conditions make the test suite hit it:

1. `CONNECTIONS_DIR` is [`SERVER_DIR_PATH / "instance-connections"`](https://github.com/dstackai/dstack/blob/master/src/dstack/_internal/server/services/runner/pool.py#L34), the real `~/.dstack/server/`, shared by every process on the machine rather than a tmp dir.
2. The directory name is `{hostname}:{port},{forwarded_ports}`, and tests use fixed values, so workers collide on the identical path.
3. CI runs [`uv run pytest -n auto src/tests --runui`](https://github.com/dstackai/dstack/blob/master/.github/workflows/build-artifacts.yml#L93), so those workers are concurrent processes.

### Actual behaviour

```
FAILED src/tests/_internal/server/background/pipeline_tasks/test_terminating_jobs.py::TestJobTerminatingWorker::test_terminates_job[sqlite]
  - FileExistsError: [Errno 17] File exists: '/Users/runner/.dstack/server/instance-connections/127.0.0.4:22,10999,10998'

src/dstack/_internal/server/background/pipeline_tasks/jobs_terminating.py:692: in _process_terminating_job
src/dstack/_internal/server/background/pipeline_tasks/jobs_terminating.py:885: in _stop_container
src/dstack/_internal/server/services/runner/ssh.py:89: in wrapper
src/dstack/_internal/server/services/runner/pool.py:113: in get_or_open
src/dstack/_internal/server/services/runner/pool.py:201: in __init__
src/dstack/_internal/server/services/runner/pool.py:317: in _resolve_conn_dir
pathlib.py:1175: FileExistsError
```

It is scheduling dependent, so it appears on one matrix job while the same commit passes on the others. Observed on `macos-latest, 3.10`; nothing is macOS specific.

Since the suite runs with `fail-fast`, one occurrence cancels the rest of the matrix, and the cancelled jobs are reported as failures.

### Expected behaviour

`_resolve_conn_dir` should tolerate the directory disappearing underneath it, and the tests should not share a real path under `~/.dstack`.

### dstack version

master (`752de2826`)

### Additional information

Seen in https://github.com/dstackai/dstack/actions/runs/30932775255/job/92071568686. Re-running the job passed with no change, which is consistent with a race rather than a stale-state problem.

navaneethkrishnansuresh:
## Summary

- Retry transient connection-directory creation races without hiding persistent file or symlink conflicts.
- Bound retries so continuous filesystem churn cannot hang the server.
- Give each pytest worker an isolated SSH connections directory.
- Add deterministic regression coverage for recovery, conflict preservation, retry limits, and test isolation.

Fixes #4095

## Testing

- uv run pytest src/tests/_internal/server/services/runner/test_pool.py -vv — 4 passed
- uv run pytest -n auto src/tests/_internal/server/background/pipeline_tasks/test_terminating_jobs.py -vv — 19 passed, 19 skipped
- uv run pytest -n auto src/tests --runui — 3359 passed, 1724 skipped
- uv run pre-commit run -a --show-diff-on-failure — passed
- uvx pyright --pythonpath .venv/bin/python -p . — 0 errors, 0 warnings

## AI assistance

Assisted moderately by Codex.

### DIFF
--- a/src/dstack/_internal/server/services/runner/pool.py
+++ b/src/dstack/_internal/server/services/runner/pool.py
@@ -36,6 +36,8 @@ CONNECTIONS_DIR = SERVER_DIR_PATH / "instance-connections"
 MIN_ALIVE_CHECK_INTERVAL = 30
 """How often (at most) `InstanceConnection.is_alive()` runs `ssh -O check`, in seconds."""
 
+_CONN_DIR_CREATE_ATTEMPTS = 3
+
 
 @dataclass(frozen=True)
 class InstanceConnectionKey:
@@ -299,6 +301,22 @@ class InstanceConnection:
         return port_map
 
     @staticmethod
+    def _create_conn_dir(conn_dir: Path) -> None:
+        for attempt in range(_CONN_DIR_CREATE_ATTEMPTS):
+            try:
+                conn_dir.mkdir(parents=True, exist_ok=True)
+                return
+            except FileExistsError:
+                # Even with exist_ok=True, pathlib raises if the directory disappears
+                # between mkdir reporting EEXIST and pathlib checking that it is a directory.
+                if conn_dir.is_dir():
+                    return
+                if conn_dir.exists() or conn_dir.is_symlink():
+                    raise
+                if attempt == _CONN_DIR_CREATE_ATTEMPTS - 1:
+                    raise
+
+    @staticmethod
     def _resolve_conn_dir(
         key: InstanceConnectionKey, ephemeral: bool
     ) -> tuple[TemporaryDirectory, Path, Path]:
@@ -314,7 +332,7 @@ class InstanceConnection:
             CONNECTIONS_DIR
             / f"{key.hostname}:{key.port},{','.join(map(str, key.ports_to_forward))}"
         )
-        conn_dir.mkdir(parents=True, exist_ok=True)
+        InstanceConnection._create_conn_dir(conn_dir)
         # Connection_dir can have a long path that won't be accepted by the ssh command,
         # so we create a short temporary symlink.
         # The symlink may be removed by age-based /tmp cleanup while the connection is still alive.
--- a/src/tests/_internal/server/services/runner/test_pool.py
+++ b/src/tests/_internal/server/services/runner/test_pool.py
@@ -1,0 +1,59 @@
+from pathlib import Path
+
+import pytest
+
+from dstack._internal.server.services.runner import pool as runner_pool
+from dstack._internal.server.services.runner.pool import InstanceConnection
+
+
+class TestInstanceConnectionCreateConnDir:
+    def test_retries_if_directory_disappears_during_creation(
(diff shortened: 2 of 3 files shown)


r4victor:
#4095 is a test suite bug, not a production path bug since InstanceConnectionPool is documented to be:

https://github.com/dstackai/dstack/blob/acb27972a4a239acd93e5233309beb091abfa8b7/src/dstack/_internal/server/services/runner/pool.py#L71-L72

so there is no need to handle access from concurrent processing.
 
As to the solution, the retry approach may catch and hide the error but does not prevent the directory from being deleted again. A Genuine multi-process safety would need the dir not to be shared (e.g. a per-process token in the name) or an OS-level lock.

PR #4185, opened 2026-08-21
[Docs] Service groups and Replica IP interpolation

Bihan:


mkdocs/docs/concepts/services.md
```diff
@@ -291,7 +291,8 @@ Setting the minimum number of replicas to `0` allows the service to scale down t
 <span id="replica-groups"></span>
 
 ??? info "Replica groups"
```
peterschmidt85: Why its collapsible section? Shouldn't we do it similar to Tasks - as a separate section?

mkdocs/docs/concepts/services.md
```diff
@@ -336,7 +337,21 @@ Setting the minimum number of replicas to `0` allows the service to scale down t
 
     </div>
 
-    > Properties such as `regions`, `port`, `image`, `env` and some other cannot be configured per replica group. This support is coming soon.
+    [`groups`](../reference/dstack.yml/service.md#groups) and top-level [`replicas`](../reference/dstack.yml/service.md#replicas) are mutually exclusive.
+
+    > Properties such as `regions`, `port`, `env` and some other cannot be configured per replica group. This support is coming soon.
+
+### Accessing replica IPs
```
peterschmidt85: Shouldn't it be under "Replica groups" section?
Also shall we make it collapsible ("???") section?

mkdocs/docs/concepts/services.md
peterschmidt85: Also, shouldn't we cross-link Replica groups and PD disaggregation sections?

peterschmidt85 (approved):
Left few minor comments

PR #4194, opened 2026-08-23
[Presets] Support for `dstack preset push` and `dstack preset pull`

peterschmidt85:
## Push and pull

A preset created on one machine can now be used on another. Push it to the registry:

```shell
$ dstack preset push qwen38-27b-mi300x main/qwen38-27b-mi300x
OK
```

Then, on any other machine, pull it from the registry and export it to a service configuration:

```shell
$ dstack preset pull main/qwen38-27b-mi300x
OK

$ dstack preset export main/qwen38-27b-mi300x -f qwen38.dstack.yml
OK

$ dstack apply -f qwen38.dstack.yml
```

`export` writes the preset's patch files next to the configuration, so the service is applied
exactly as it was verified.

Push shares the verified service, its benchmark, and the hardware it was verified on. The
configuration, prompt, and trials that produced it stay on the machine that created it.

A pulled preset is used like any other, and is named `<project>/<name>`, which can never clash
with a preset of your own:

```shell
$ dstack preset list -a
 NAME                      ID                                    BASE              CONSTRAINTS      BENCHMARK                             STATUS           SUBMITTED
 main/qwen38-27b-mi300x    e7c8b170-8ca6-41cc-bfa9-ed1ba71c8c88  Qwen/Qwen3.8-27B  io=10K/1.5K c=4  tps/user=138 ttft=852ms ctx=1M        pulled           11 mins ago
 qwen38-27b-mi300x-crack2  6900d9d7                              Qwen/Qwen3.8-27B  io=10K/1.5K c=4  tps/user=138 ttft=852ms ctx=1M  ████  verified (4/15)  1 week ago
```

Its status is `pulled`, in the table above and in `dstack preset get --json`.

## Names

A name is a moving reference, as a Docker tag is. Pushing one that already exists transfers it to
the new preset, and the superseded preset keeps existing without a name, reached as
`<project>/<id>`, which `pull` also accepts. A preset pulled that way lands untagged, so it never
shadows the one the name points at. Deleting a preset releases its name rather than handing it
back to an older one.

## Registry

The registry is part of the dstack server. On dstack Sky, that is https://sky.dstack.ai. Push and
pull require membership in the project. If the server has no registry, both say so.

The server comes from the project's entry in `~/.dstack/config.yml`. A project that is not there
falls back to dstack Sky; `DSTACK_NO_SKY_FALLBACK` turns the fallback off.

> [!NOTE]
> The fallback is exactly the one `dstack login` applies.

## UI

Presets are listed with filters for project, user, and base model. A preset opens on what it was
verified on, the benchmark and the constraints it holds under, and the commands to pull, export,
and apply it. The pages are Sky-only, gated as Billing is.

## Also here

- `dstack preset export` no longer fails with `IsADirectoryError` when the preset mounts a
  directory in `files`.
- Presets can be the target of events, and `list_events` filters by them, so a server with a
  registry can record who pushed what. The events table shows them and filters by them.
- Events moves down the sidebar to sit next to Projects.

## Breaking changes

- `dstack preset get --json` reports the served model repository as `repo` instead of `model`, so
  `base` (the model family), `repo` (what the service loads), and `service.model` (what the
  endpoint serves it as) are now distinct.

## Backward compatibility

- Presets created by earlier versions are upgraded when read.

> [!NOTE]
> The registry that `push` and `pull` talk to is implemented separately and is not part of this
> PR.

### DIFF
--- a/src/dstack/_internal/cli/services/presets/registry.py
+++ b/src/dstack/_internal/cli/services/presets/registry.py
@@ -144,23 +144,24 @@ def pull_preset_from_registry(store: PresetStore, registry_ref: str) -> None:
         validate_preset_spec_files(remote.spec)
     except ValueError as e:
         raise CLIError(f"Preset {registry_ref!r} cannot be pulled: {e}") from e
-    # The local name is the qualified ref. It can never collide with a locally
-    # created preset (local names cannot contain `/`), so the only possible
-    # holder is an earlier pull; the name silently moves to the fresh copy,
-    # Docker-style. A non-current preset (pulled by id after the name was
-    # repointed) must not take the name from the current one — like a Docker
-    # pull by digest, it lands untagged.
-    qualified_name = f"{project}/{remote.name}"
-    local_name = qualified_name if remote.is_current else None
-    holder = store.find_by_name(qualified_name)
-    if local_name is not None:
+    # A preset carries its name only while the name resolves to it, so one
+    # pulled by id after the name moved on arrives without one and lands
+    # untagged, like a Docker pull by digest.
+    qualified_name = f"{project}/{remote.name}" if remote.name is not None else None
+    local_name = qualified_name
+    if qualified_name is not None:
+        # The qualified name can never collide with a locally created preset
+        # (local names cannot contain `/`), so the only possible holder is an
+        # earlier pull; the name silently moves to the fresh copy.
+        holder = store.find_by_name(qualified_name)
         if holder is not None and holder.id != preset_id:
             store.release_name(qualified_name)
-    elif holder is not None and holder.id == preset_id:
-        # This copy holds the name from an earlier pull, when it was current.
-        # Re-pulling it by id must not strip the name off the local store
-        # entirely, or local refs to it would stop resolving.
-        local_name = qualified_name
+    else:
+        # Re-pulling by id a copy that already holds the name from an earlier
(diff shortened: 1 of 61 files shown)


src/dstack/_internal/cli/services/presets/registry.py
```diff
@@ -0,0 +1,305 @@
...
+        # CLIError text is escaped when printed, so backticks, not rich markup.
+        raise CLIError(f"Invalid registry reference {ref!r}: expected `<project>/<name|id>`")
+    return project, rest
+
+
+def resolve_registry_client(project: str) -> APIClient:
+    """The server hosting the registry a ref points at: the config entry with
+    the project's name, or
+    Sky with any configured Sky token (dstack tokens are user tokens, so one Sky
+    entry authenticates any Sky project the user is a member of)."""
```
jvstme, line 52: So if I have several Sky users, the CLI will select a random one?

This might be unexpected — that random user may have insufficient permissions, or it may be a user that I just wouldn't want to be used and recorded as actor in events.

Long-term, this could be solved by [redesigning CLI authorization](https://docs.google.com/document/d/12-JWIoQDtVP_D6n6Qnss3j-TQ90XHSU6Rj6QmpjWtu0/edit?usp=sharing). Short term — by not falling back to Sky if it's not clear which user to choose (if there are more than one)?

PR #4199, opened 2026-08-24
Fix preset verification rejecting a tool's synthetic dataset name

### LINKED ISSUE #4198: [Bug]: Preset creation fails with a dataset error when no `dataset` is set
### Steps to reproduce

1. Create a preset with `shared_prefix_tokens` set and no `dataset`:

```yaml
type: preset
name: qwen3-27b-agentic
base: Qwen/Qwen3.8-27B
min_context_length: 262144
max_ttft: 5000
trials: 4
concurrency: 4
input_tokens: 131072
output_tokens: 512
shared_prefix_tokens: 130048
fleets: [mi300x-fleet]
```

2. Wait for the agent to finish all trials and verify the final service.

Whether it fails depends on which framework the agent picks, so it may take more than one
attempt. I ran this config twice on the same MI300X fleet three hours apart: the session that
chose vLLM succeeded, the one that chose SGLang failed.

### Actual behaviour

After ~1.5h, all four trials done and the final service verified and answering requests, preset creation fails with:

```
Claude final benchmark dataset does not match the requested dataset
```

I never set `dataset`, so there is no dataset I could have gotten wrong, and the message doesn't say which two values didn't match. The session is discarded and every run is torn down, including the final service that had already passed its own verification. Nothing in the CLI output suggests what to change, and re-running costs another 1.5h and another AMD instance.

Underneath, the reported dataset name is compared to the default `random` as a string:

https://github.com/dstackai/dstack/blob/2f024c64935504b65dfd8b4ee869365e9fb405de/src/dstack/_internal/cli/services/presets/verify.py#L140-L141

The agent reported `generated-shared-prefix`, which is what the system prompt tells it to pass when `shared_prefix_tokens` is set and it benchmarks with SGLang. Framework choice is the agent's own, so this is reachable on any fleet:

https://github.com/dstackai/dstack/blob/2f024c64935504b65dfd8b4ee869365e9fb405de/src/dstack/_internal/cli/services/presets/resources/system_prompt.md#L337-L340

The report schema has no way to describe that benchmark. `PresetRandomWorkload` allows `shared_prefix_tokens` but pins `dataset` to `random`; `PresetWorkload` takes any `dataset` but forbids `shared_prefix_tokens`, and extra fields are rejected:

https://github.com/dstackai/dstack/blob/2f024c64935504b65dfd8b4ee869365e9fb405de/src/dstack/_internal/cli/models/presets.py#L25-L36

The agent's trial report stated both facts and so matched neither branch:

```json
{"api": "completions", "dataset": "generated-shared-prefix", "num_requests": 16,
 "input_tokens": 131072, "output_tokens": 512, "concurrency": 4, "shared_prefix_tokens": 130048}
```

For the final report it dropped `shared_prefix_tokens` and kept `dataset`, which is schema-valid and is the branch verification rejects. Dropping `dataset` instead would have passed. So the outcome turns on which of two true facts the agent discards, and one of the two schema-valid choices always fails.

Two consequences: with a shared prefix, any benchmark tool that doesn't happen to name its synthetic dataset `random` can fail verification, and vLLM passes only because its flag is spelled `--dataset-name random`. And when the generic branch is used, `shared_prefix_tokens` is absent from the stored benchmark record even though the config set it.

### Expected behaviour

A config that sets `shared_prefix_tokens` and no `dataset` passes verification whichever benchmark tool the agent uses, and a benchmark that is actually rejected produces an error naming the two values that didn't match.

r4victor:
Fixes #4198. A preset with `shared_prefix_tokens` and no `dataset` failed verification whenever the agent benchmarked with a tool that doesn't call its synthetic dataset `random`.

- Collapse `PresetRandomWorkload` into `PresetWorkload` so one workload can carry both the tool's dataset name and `shared_prefix_tokens`, which the old union made mutually exclusive.
- Compare the requested shared prefix instead of the dataset name for synthetic workloads, since `random` is dstack's own name for them while the report carries the benchmark tool's.
- Compare `concurrency` against the request, which nothing checked before; `input_tokens` and `output_tokens` stay uncompared because they are measured means rather than an echo of the request.
- Name both values in every workload mismatch error.
- Update the system prompt's synthetic-workload branch to match the schema.

AI assistance: written with Claude Code.

### DIFF
--- a/src/dstack/_internal/cli/models/presets.py
+++ b/src/dstack/_internal/cli/models/presets.py
@@ -1,6 +1,6 @@
 import re
 from datetime import datetime
-from typing import Annotated, Literal, Optional, Union
+from typing import Annotated, Literal, Optional
 
 from pydantic import (
     Field,
@@ -24,16 +24,19 @@ PRESET_EXCLUDED_FIELDS = ("name", "gateway", *ProfileParams.model_fields)
 
 class PresetWorkload(CoreModel):
     api: Literal["chat_completions", "completions"]
-    dataset: str
+    dataset: Optional[str] = None
+    """The benchmark tool's own name for the data it served. It is the requested
+    dataset when the configuration named one; for the synthetic `random` workload it
+    is whatever the tool calls the data it generates (`random`,
+    `generated-shared-prefix`, ...), recorded but never compared with dstack's own
+    `random`."""
     num_requests: PositiveInt
     input_tokens: PositiveInt
     output_tokens: Annotated[int, Field(ge=2)]
     concurrency: PositiveInt
-
-
-class PresetRandomWorkload(PresetWorkload):
-    dataset: Literal["random"] = "random"
     shared_prefix_tokens: Annotated[int, Field(ge=0)] = 0
+    """How many leading tokens every measured request shared. Only a synthetic
+    workload has one: a named dataset defines its own requests."""
 
 
 class PresetBenchmarkLatency(CoreModel):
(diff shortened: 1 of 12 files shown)


peterschmidt85:
Will review and resolve conflicts myself tomorrow!

peterschmidt85:
Closing in favor of #4201: same fix, cleaner contract (dataset = the requested dataset, absent for synthetic), no system prompt changes.

PR #4253, opened 2026-09-02
Fix target type filter on Events page

### LINKED ISSUE #4211: [Bug]: `Target types` filter does not work on the Events page in UI
### Steps to reproduce

1. Open the Events page.
2. Set the `Target types` filter to any allowed value.

### Actual behaviour

The event list is always empty.

The API returns error 429:

```
{"detail":[{"type":"enum","loc":["body","include_target_types",0],"msg":"Input should be 'project', 'user', 'fleet', 'instance', 'run', 'job', 'volume', 'gateway' or 'secret'","input":"Instance","ctx":{"expected":"'project', 'user', 'fleet', 'instance', 'run', 'job', 'volume', 'gateway' or 'secret'"}}]}
```

### Expected behaviour

The event list shows all events that have at least one target of the specified type.

### dstack version

0.21.2

### Server logs

```shell

```

### Additional information

_No response_

hhhhzzzj:
Fix target type filter on the Events page

Selecting a target type on the Events page currently sends its display label to the API. For example, selecting "Instance" sends include_target_types=["Instance"], while the API expects the lowercase enum value "instance".

This keeps the display label in the UI and uses the canonical value in the request. A focused regression test covers the label/value mapping.

The focused Events test passes (4 tests), and ESLint, Prettier, and git diff --check pass. The backend test could not run in this checkout because pytest_asyncio is missing. The full frontend run still has unrelated existing failures.

For context, nightcityblade previously said they planned to work on this issue. jvstme later asked hhhhzzzj to proceed.

Fixes #4211


jvstme (approved):
I've rewritten the PR to simplify it, but the overall solution looks good to me

PR #4271, opened 2026-09-08
Disable gateway response buffering for model services

### LINKED ISSUE #4269: [Bug]: Gateway buffers streamed LLM responses, inflating TTFT to 16 KB worth of tokens
### Steps to reproduce

1. Deploy a vLLM or SGLang service behind a gateway with `https` enabled and a `model` configured.
2. Send a streaming chat completion through the gateway and watch the output arrive:

   ```shell
   curl -N -sS https://<run>.<gateway-domain>/v1/chat/completions \
     -H "Authorization: Bearer $DSTACK_TOKEN" -H "Content-Type: application/json" \
     -d '{"model": "<model>", "stream": true, "max_tokens": 200, "messages": [{"role": "user", "content": "Write a long story"}]}'
   ```

3. Run `dstack attach <run>` and send the same request to the forwarded port on `localhost`.

### Actual behaviour

Through the gateway the response does not stream. Nothing arrives, not even the response headers, until about 16 KB of SSE has accumulated, roughly 65 OpenAI-format chunks. Shorter responses arrive all at once when generation finishes. Directly against the replica, chunks arrive as they are generated.

Measured with a mock OpenAI-compatible service that emits one 247-byte SSE chunk every 250 ms, first chunk after 0.5 s, each chunk stamped with its server-side send time:

- Direct: TTFT 0.5 s, one chunk per read.
- Gateway, 60 tokens: TTFT 16.5 s, the whole response in one read at the end.
- Gateway, 200 tokens: TTFT 17.5 s, then bursts of 64 to 66 chunks every 16.5 s.
- Gateway, same service sending `X-Accel-Buffering: no`: TTFT 0.5 s plus request overhead, one chunk per read.

Cause: the service `location @` block uses the nginx default `proxy_buffering on`. In that mode nginx releases data only when a 4 KB proxy buffer fills or the response ends, and over HTTPS the 16 KB `ssl_buffer_size` output buffer sits on top of that, so the effective threshold becomes 16 KB.

https://github.com/dstackai/dstack/blob/7823319abf3aacea7ca7af453e3c226f5e89c4ce/src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2#L76-L95

vLLM and SGLang do not send `X-Accel-Buffering: no` on streaming responses, so nothing tells nginx to bypass buffering. TGI does send it, added in huggingface/text-generation-inference#498 for exactly this reason.

https://github.com/vllm-project/vllm/blob/6fbb00b18874e27ba7d7adc0a3b8e93fee763ab1/vllm/entrypoints/openai/chat_completion/api_router.py#L77-L80

The model entrypoint (`gateway.<domain>/proxy/models/<project>/`) is affected as well. Its first hop is fine because the gateway app sets `X-Accel-Buffering: no` on streaming responses, but the app forwards requests to the service through the same nginx `location @`, so the second hop buffers.

https://github.com/dstackai/dstack/blob/7823319abf3aacea7ca7af453e3c226f5e89c4ce/src/dstack/_internal/proxy/lib/routers/model_proxy.py#L61-L65

### Expected behaviour

Streamed tokens reach the client as the replica emits them. TTFT through the gateway should differ from direct access only by the per-request overhead (TLS, auth subrequest, tunnel connection), not by the size of the response.

### dstack version

0.21.5

### Server logs

_No response_

### Additional information

Proposed fix: disable proxy buffering for services registered with a `model`, mirroring how `cors_enabled` is derived in the gateway registry. Add a gateway-only `proxy_buffering` flag to the `Service` record, set it from the model at registration, render `proxy_buffering off;` in `location @`, and migrate existing services on gateway startup so no redeploy is needed. No server or API changes are required since the gateway already receives the model.

`proxy_buffering` is per location and cannot depend on the response, so this also disables buffering for the service's non-streaming requests. For those the cost is that a large body to a slow client holds the replica connection for the download instead of being spooled by nginx. Generation is already complete by then, so no engine resources are held. Services without a `model` keep buffering. A per-service override can be added later if a batch-heavy deployment needs it.

Ideally, we'd like a different buffering behavior depending on the requests's `body.stream` flag. It's not possible with nginx and is yet another reason to use nginx only as an edge proxy, and have a custom LLM/AI-aware proxy implementation.

pentaoa:
## Summary

Fixes #4269.

- Persist a gateway-only `proxy_buffering` property on services and disable it when registering a model service (both OpenAI and TGI formats).
- Pass the property into the Nginx service configuration and render `proxy_buffering off;` in the named service location that forwards regular HTTP requests to replicas.
- On gateway startup, derive the property for previously saved services from their associated models, persist it, and regenerate their configurations. Existing services do not need to be redeployed. Services without models retain the default buffering behavior.

This follows the issue's proposed scope. It does not change the server registration API, request buffering, WebSocket handling, or read timeouts. Disabling response buffering also applies to non-streaming requests to model services: a slow client keeps the upstream connection occupied while downloading the response instead of allowing Nginx to spool it.

## Validation

Added tests that exercise service and replica registration through the gateway API and inspect the generated service location for OpenAI, TGI, and non-model services. Added startup coverage loading an old state file without the new field, checking project-scoped model association and persistence across two restarts.

Before the fix: **4 failed, 1 passed**; all four model-service cases failed because the generated configuration omitted `proxy_buffering off;`. The non-model control passed.

After the fix:

```text
uv run --no-sync pytest -q src/tests/_internal/proxy/gateway src/tests/_internal/proxy/lib
62 passed in 3.61s
```

Ruff 0.12.7 check and format checks passed for all five changed Python files; `git diff --check` passed.

These tests use the existing SSH/Nginx system mocks and execute the actual configuration renderer and state-file recovery. I did not run a real Nginx/TLS gateway or measure TTFT locally, and did not run GPU/model workloads or the full server suite.

AI-assisted implementation and tests; I reviewed the changed registration, rendering, and startup paths and ran the validation described above.

### DIFF
--- a/src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
+++ b/src/dstack/_internal/proxy/gateway/resources/nginx/service.jinja2
@@ -76,6 +76,9 @@ server {
     location @ {
         set $dstack_replica_hit 1;
         {% if replicas %}
+        {% if not proxy_buffering %}
+        proxy_buffering off;
+        {% endif %}
         {% if cors_enabled %}
         proxy_hide_header 'Access-Control-Allow-Origin';
         proxy_hide_header 'Access-Control-Allow-Methods';
--- a/src/dstack/_internal/proxy/gateway/services/nginx.py
+++ b/src/dstack/_internal/proxy/gateway/services/nginx.py
@@ -68,6 +68,7 @@ class ServiceConfig(SiteConfig):
     replicas: list[ReplicaConfig]
     has_router_replica: bool = False
     cors_enabled: bool = False
+    proxy_buffering: bool = True
 
 
 class ModelEntrypointConfig(SiteConfig):
--- a/src/dstack/_internal/proxy/gateway/services/registry.py
+++ b/src/dstack/_internal/proxy/gateway/services/registry.py
@@ -62,6 +62,7 @@ async def register_service(
         replicas=(),
         has_router_replica=has_router_replica,
         cors_enabled=cors_enabled,
+        proxy_buffering=model is None,
     )
 
     async with lock:
@@ -407,6 +408,7 @@ async def get_nginx_service_config(
         replicas=sorted(replicas, key=lambda r: r.id),  # sort for reproducible configs
         has_router_replica=service.has_router_replica,
         cors_enabled=service.cors_enabled,
+        proxy_buffering=service.proxy_buffering,
     )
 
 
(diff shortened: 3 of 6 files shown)


peterschmidt85:
@pentaoa we don't accept PRs if the implementation is not discussed first and the author that submits the PR is not actively using dstack.

pentaoa:
@peterschmidt85 Thanks for clarifying, and apologies for not discussing the implementation with you first. I’m not currently an active dstack user, so I understand these PRs don’t meet your contribution requirements. I’ll hold off on further submissions.

PR #4272, opened 2026-09-08
Collect DinD metrics from the container cgroup root

### LINKED ISSUE #4267: [Bug]: Run metrics with `docker: true` fail until the first nested container starts
### Steps to reproduce

1. Start a run with `docker: true` that does some work before starting a nested container, then keeps a nested container running:

    ```yaml
    type: task
    name: dind-metrics
    docker: true
    commands:
      - sleep 300
      - docker run -d --name stress polinux/stress stress --vm 1 --vm-bytes 4G --vm-keep
      - sleep infinity
    ```

2. Watch the `dstack server` logs during the first five minutes.
3. After the nested container is up, run `dstack metrics dind-metrics`.

### Actual behaviour

Until the first nested container starts, the server logs the traceback below on every metrics collection cycle and stores no metrics points for the job.

Once a nested container is running, collection resumes, but the reported CPU and memory cover only the runner, `dockerd` and processes started directly by the job's shell. Everything running in nested containers is excluded, which is where the workload of a `docker: true` run usually lives.

The mechanism: `start-dockerd` moves every process in the container's root cgroup, including `dstack-runner`, into a child cgroup `/sys/fs/cgroup/dind`. In cgroup v2, a child cgroup only gets `memory.*` files when the memory controller is enabled in the parent's `cgroup.subtree_control`, which `start-dockerd` does not do, so `/sys/fs/cgroup/dind/memory.current` does not exist after the move. `cpu.stat` exists on every non-root cgroup regardless of controllers, which is why the CPU read succeeds and the memory read is the first to fail.

https://github.com/dstackai/dstack/blob/aaeb2017ea95322e14686ae0073d26fd0b75dfb6/docker/dind/start-dockerd#L95-L99

When `runc` creates the cgroup for the first nested container under `/sys/fs/cgroup/docker/`, it enables the controllers in the ancestors' `cgroup.subtree_control`. From then on `/sys/fs/cgroup/dind/memory.current` exists, but it accounts for the `dind` subtree only. The nested containers live under the sibling `/sys/fs/cgroup/docker/`.

The runner resolves its cgroup from `/proc/self/cgroup` and reads from that directory with no fallback:

https://github.com/dstackai/dstack/blob/aaeb2017ea95322e14686ae0073d26fd0b75dfb6/runner/internal/runner/metrics/metrics.go#L41-L57

Non-dind images are unaffected because the runner stays in the container's root cgroup, which is the container's own cgroup on the host and accounts for everything in the container.

### Expected behaviour

Metrics for `docker: true` runs are collected from the start and include nested containers.

Suggested fix: make the runner prefer the cgroup mount root (`/sys/fs/cgroup`) whenever `memory.current` exists there, and fall back to the process cgroup otherwise. Inside a container with a cgroup namespace, the mount root is the container's own cgroup, so it covers the runner, `dockerd` and all nested containers. On a bare host or without a cgroup namespace the root has no `memory.current`, so the process cgroup is still used there.

Enabling the controllers in `start-dockerd` the way moby's `hack/dind` does would only remove the error window and would still exclude nested containers.

### dstack version

master (aaeb2017e)

### Server logs

```shell
ERROR    dstack._internal.server.background.scheduled_tasks.metrics:139 Failed to collect job dind-metrics-0-0 metrics
  ...
  File ".../dstack/_internal/server/services/runner/client.py", line 349, in _raise_for_status
    raise RunnerResponseStatusError(response)
dstack._internal.server.services.runner.client.RunnerResponseStatusError: GET /api/metrics: 500: could not read memory usage: open /sys/fs/cgroup/dind/memory.current: no such file or directory
```

### Additional information

Related: #3398 has the same runner error on Kubernetes with `privileged: true`, but for a different reason (no cgroup namespace, so `/sys/fs/cgroup` is the host root). The fallback described above would not fix that case.

pentaoa:
## Summary

Fixes #4267.

Prefer the cgroup mount root for runner CPU and memory accounting when its `memory.current` exists. In a container cgroup namespace this includes the whole run, including nested containers under the sibling `docker/` group after `start-dockerd` moves the runner into `dind/`.

When the mount exposes the host root, which has no `memory.current`, retain the existing `/proc/self/cgroup` resolution. Resolve this on each collection, preserving support for process cgroup changes. A failure to inspect the root other than a missing file is reported rather than silently selecting a different accounting scope.

The change follows the approach suggested in the issue. It does not enable controllers, modify the Docker startup script, change GPU collection, or restore cgroup v1 support. The namespace behavior addressed by #3402 is retained; the separate privileged Kubernetes limitation described in #4267 is not claimed as fixed.

## Validation

Filesystem-fixture regression cases cover:

- ordinary containers whose process cgroup is `/`;
- DinD before the first nested container, when child memory files are absent;
- DinD after child memory files appear, with deliberately different root and child counters;
- host-namespace fallback to the process cgroup;
- missing proc metadata errors when fallback is required.

An additional test exercises `GetSystemMetrics` using root CPU, memory usage, and cache counters and checks the resulting working set.

With the original path-selection behavior extracted unchanged, both DinD cases failed because they selected `/dind` instead of the mount root. After the fix:

```text
go test -race ./internal/runner/metrics -run '^(TestGetMetricsCgroupPath|TestGetProcessCgroup|TestGetSystemMetrics_ContainerRoot)' -count=1
ok github.com/dstackai/dstack/runner/internal/runner/metrics 2.199s

go vet ./internal/runner/metrics
# passed
```

Changed Go files are gofmt-formatted and `git diff --check` passes. The metrics test package also cross-compiles for Linux/amd64 (compile check only, not Linux execution).

Validation ran on macOS using temporary proc/cgroup fixtures. I did not launch nested Docker containers, run GPU workloads, execute the full runner suite (existing collector-constructor tests require Linux procfs), or run golangci-lint locally.

AI-assisted implementation and tests; I inspected the DinD startup code, previous cgroup-path fix, and current collector, and ran the focused validation above.

### DIFF
--- a/runner/internal/runner/metrics/cgroups.go
+++ b/runner/internal/runner/metrics/cgroups.go
@@ -6,10 +6,28 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/dstackai/dstack/runner/internal/common/log"
 )
+
+func getMetricsCgroupPath(ctx context.Context, mountPoint, procPidCgroupPath string) (string, error) {
+	// In a container cgroup namespace, the mount root accounts for the whole
+	// container, including nested containers in sibling groups of /dind.
+	// The host's root cgroup has no memory.current; keep using the process
+	// cgroup in that case.
+	if _, err := os.Stat(path.Join(mountPoint, "memory.current")); err == nil {
+		return mountPoint, nil
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return "", fmt.Errorf("stat cgroup root memory.current: %w", err)
+	}
+	cgroupPathname, err := getProcessCgroupPathname(ctx, procPidCgroupPath)
+	if err != nil {
+		return "", fmt.Errorf("get cgroup pathname: %w", err)
+	}
+	return path.Join(mountPoint, cgroupPathname), nil
+}
 
 func getProcessCgroupMountPoint(ctx context.Context, ProcPidMountsPath string) (string, error) {
 	// See proc_pid_mounts(5) for the ProcPidMountsPath file description
--- a/runner/internal/runner/metrics/cgroups_test.go
+++ b/runner/internal/runner/metrics/cgroups_test.go
@@ -15,6 +15,51 @@ const (
 	rootMountLine    = "/dev/nvme0n1p5 / ext4 rw,relatime 0 0"
 )
 
+func TestGetMetricsCgroupPath(t *testing.T) {
+	for _, tc := range []struct {
+		name         string
+		processGroup string
+		rootMemory   bool
+		childMemory  bool
+		wantGroup    string
+	}{
+		{name: "ordinary container", processGroup: "/", rootMemory: true, wantGroup: "/"},
+		{name: "dind before nested container", processGroup: "/dind", rootMemory: true, wantGroup: "/"},
+		{name: "dind with nested container", processGroup: "/dind", rootMemory: true, childMemory: true, wantGroup: "/"},
+		{name: "host cgroup namespace", processGroup: "/system.slice/docker.scope", childMemory: true, wantGroup: "/system.slice/docker.scope"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			mountPoint := t.TempDir()
+			procFile := createProcFile(t, "cgroup", "0::"+tc.processGroup)
+			if tc.rootMemory {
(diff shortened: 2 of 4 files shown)


peterschmidt85:
@pentaoa we don't accept PRs if the implementation is not discussed first and the author that submits the PR is not actively using dstack.

pentaoa:
Thanks for clarifying, and apologies for not discussing the implementation with you first. I’m not currently an active dstack user, so I understand these PRs don’t meet your contribution requirements. I’ll hold off on further submissions.

PR #3955, opened 2026-06-11
Allow configuring `subnet_ids` in Azure settings

jvstme:
Similarly to `vpc_ids`, allow selecting specific
subnets to be attached to dstack VMs.

```yaml
projects:
- name: main
  backends:
  - type: azure
    subscription_id: ...
    tenant_id: ...
    creds:
      type: client
      client_id: ...
      client_secret: ...
    subnet_ids:
      westeurope: my-resource-group/my-vpc/my-subnet
    regions: [westeurope]
```

Closes #3931

PR #4274, opened 2026-09-08
Allow configuring `subnetworks` in GCP settings

### LINKED ISSUE #4254: [Feature] GCP backend: allow specifying the subnetwork (currently always picks the first usable subnet in the VPC)
## Problem

The GCP backend lets you pin the VPC via `vpc_name`, but not the subnet. Subnet selection is:

```python
# src/dstack/_internal/core/backends/gcp/resources.py
def get_vpc_subnet_or_error(...):
    vpc_subnets = get_vpc_subnets(vpc_name, region, usable_subnets)
    if vpc_subnets:
        return vpc_subnets[0]   # <- first usable subnet, no way to influence
```

i.e. the first subnet returned by `ListUsableSubnetworks` for that VPC+region wins.

In a VPC with multiple subnets in the same region, subnets often serve different purposes (different sizes, routing, peering/VPN reachability, reserved ranges), and you may want dstack to provision instances and gateways in one subnet rather than another. Today there is no way to influence the choice — which subnet dstack lands in is effectively arbitrary. This is especially limiting for private gateways and fleets with `public_ip: false`, where clients can only reach the VMs if they land in a subnet the client network routes to.

## Proposal

Add an optional `subnetwork` field to the GCP backend config (alongside `vpc_name`), e.g.:

```yaml
- type: gcp
  project_id: my-project
  vpc_name: my-vpc
  subnetwork: my-subnet   # optional; per-region map also fine
  creds:
    type: default
```

When set, `get_vpc_subnet_or_error` would select that subnet (erroring if it doesn't exist in the region) instead of `vpc_subnets[0]`. Happy to contribute a PR if the approach sounds right.

**dstack version**: 0.21.3 (server, GCP backend; behavior unchanged on master)

jagretti:
Closes #4254

Adds an optional `subnetworks` field to the GCP backend config — a mapping from regions to subnet names in the VPC specified by `vpc_name`:

```yaml
projects:
- name: main
  backends:
  - type: gcp
    project_id: my-project
    regions: [europe-west4]
    vpc_name: my-vpc
    subnetworks:
      europe-west4: my-subnet
    creds:
      type: default
```

Previously, `dstack` always provisioned in the first usable subnet returned by `ListUsableSubnetworks` for the VPC and region, with no way to influence the choice. This is limiting when a VPC has multiple subnets in the same region serving different purposes — especially for private gateways and fleets with `public_ip: false`, where clients can only reach the VMs if they land in a subnet the client network routes to.

## Behavior

- In mapped regions, instances, gateways, and TPUs are provisioned in the specified subnet (matched by name against usable subnets of the VPC; shared VPCs supported).
- In unmapped regions, behavior is unchanged: any usable subnet of the VPC is used. With `subnetworks` unset, behavior is identical to before.
- Backend config validation errors early if `subnetworks` is set without `vpc_name`, if it's keyed by a region not in `regions`, or if a mapped subnet doesn't exist in its region (the error lists the available subnets).
- `extra_vpcs`/`roce_vpcs` subnet selection is unchanged (out of scope).

The field shape and behavior mirror the Azure `subnet_ids` setting (#3955).

## Testing

- New unit tests for subnet matching (`TestGetVpcSubnetOrError`) and configurator validation; `uv run pytest src/tests/_internal/core/backends/gcp/` passes.
- Docs updated in `concepts/backends.md`; the reference page picks up the field description from the model schema.

### DIFF
--- a/mkdocs/docs/concepts/backends.md
+++ b/mkdocs/docs/concepts/backends.md
@@ -615,6 +615,30 @@ gcloud projects list --format="json(projectId)"
         allowing all traffic within the VPC. This is needed for multi-node tasks to work.
         The default VPC already permits traffic within the VPC.
 
+        If the VPC has multiple subnets in a region, `dstack` uses any usable subnet.
+        Specify `subnetworks` to target a specific subnet in a region:
+
+        <div editor-title="~/.dstack/server/config.yml">
+
+        ```yaml
+        projects:
+        - name: main
+          backends:
+            - type: gcp
+              project_id: gcp-project-id
+              creds:
+                type: default
+
+              regions: [europe-west4]
+              vpc_name: my-custom-vpc
+              subnetworks:
+                europe-west4: my-custom-subnet
+        ```
+
+        </div>
+
+        Regions not mapped in `subnetworks` keep using any usable subnet of the VPC.
+
     === "Shared VPC"
     
         <div editor-title="~/.dstack/server/config.yml">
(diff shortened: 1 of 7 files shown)


jvstme (approved):
Thanks for the PR, LGTM!

PR #4277, opened 2026-09-09
fix: reset agent retry budget only on real progress

### LINKED ISSUE #4163: [Bug]: [Presets] A repeatedly dying agent retries forever instead of failing
### Steps to reproduce

1. Make the agent process die at startup without submitting a report, in a way that still prints one `assistant` line. For example, with subscription auth supplied only by user-scope settings while the native OAuth session is expired: dstack spawns `claude` with `--setting-sources project,local` (`agent.py:356-357`), which excludes the user scope, so the agent starts and dies immediately.
2. Start any preset creation: `dstack apply -f preset.dstack.yml -y`.

### Actual behaviour

```
Agent process exited without a report; resuming in 30s.
```

repeats every 30 seconds, indefinitely, before any instance is provisioned. Never 60s or 120s: `agent.py:551-552` sets `made_progress` on any `assistant` line without inspecting its content, and `agent.py:238-239` resets the retry budget (`agent.py:52`) before the pop at `agent.py:246`, so the budget never drains and the `while True` at `agent.py:216` has no attempt cap. The agent's own error is captured at `agent.py:555-556` but the loop message omits it, so the cause is only visible in `.agent-stdout.jsonl`. A report-less death that prints no `assistant` line does bound correctly and exits after 30/60/120s.

### Expected behaviour

The creation fails immediately, printing the agent's error:

```
Failed to authenticate: OAuth session expired and could not be refreshed
```

Retries stay only for a transient death after real progress, bounded, printing the error each attempt.

breken-ai:
## Root cause

The resume-delay budget (30/60/120s) resets whenever the dying agent printed ANY assistant line. A crash-looping agent (your expired-OAuth case: one assistant line then exit) makes identical "progress" every attempt, so the `made_progress` boolean can't distinguish "got further" from "got to the same place again" - the retry loop never terminates.

## Fix

Replace the boolean with an assistant-line count and reset the budget only when an attempt exceeds the best previous attempt. A dying agent drains the 30/60/120s retries and returns the error; a genuinely progressing agent still earns resets.

## Verification

End-to-end regression test with a fake claude subprocess that prints one assistant line then exits 1: unbounded on current HEAD (9 attempts and counting), exactly 4 attempts with the fix. Full presets agent suite: 64 passed, 1 pre-existing Windows-only env failure (fails on main too, unrelated).

Fixes #4163

> Built by breken, your AI support engineer - breken.ai - this one's on us.

### DIFF
--- a/src/dstack/_internal/cli/services/presets/agent.py
+++ b/src/dstack/_internal/cli/services/presets/agent.py
@@ -101,7 +101,10 @@ class PresetAgentProcessOutput:
     report_data: Optional[dict[str, Any]] = None
     error: Optional[str] = None
     session_id: Optional[str] = None
-    made_progress: bool = False
+    # Assistant lines seen this attempt. A count, not a flag: a crash-looping
+    # agent reaches the same point every attempt, and only an attempt that got
+    # FURTHER than any before it earns a fresh retry budget.
+    assistant_events: int = 0
 
 
 def _get_claude_version(auth: "ClaudeAuth") -> Optional[str]:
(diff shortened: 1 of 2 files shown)


peterschmidt85:
@breken-ai we don't accept PRs if the implementation is not discussed first and the author that submits the PR is not actively using dstack.

PR #4278, opened 2026-09-09
fix: resolve preset env before claiming the agent session

### LINKED ISSUE #4164: [Bug]: [Presets] Any failure during resume marks the creation permanently failed
### Steps to reproduce

1. Start a creation from a configuration with any `env` entry, e.g. `env: [MODEL_LABEL=some-value]`.
2. Stop it: `dstack preset stop <id> -y`. It correctly records `interrupted` and prints `Resume it with dstack preset resume <id>`.
3. Resume from a shell where that variable is not set: `dstack preset resume <id>`. It fails with `Environment variable MODEL_LABEL is not set` (#4165).
4. Resume again, with the variable set: `dstack preset resume <id>`.

### Actual behaviour

Step 4 refuses:

```
Preset <id> creation failed and cannot be resumed
```

Step 3 is what marked the session failed. `create_preset` wraps the whole creation in `except BaseException: _close_agent_session(session, "failed")` (`create.py:437-439`), so any exception is terminal: it writes `failed` and deletes the workspace (`create.py:754-756`, `workspace.py:120-130`). Step 3's exception is raised at `create.py:415`, before `_create_preset` is entered, so nothing had started and nothing had changed, yet the session became unresumable. `load_resumable_session` then refuses (`session.py:331-332`).

Not caused by reconcile: `_is_reconcilable` requires status `running`, an existing report, and a dead owner (`create.py:252-262`), so an interrupted session is never reconciled.

### Expected behaviour

Step 3 leaves the session `interrupted`, and step 4 resumes it. A failure that happens before the agent is spawned is not a failed creation.

breken-ai:
## Root cause

`create_preset` (src/dstack/_internal/cli/services/presets/create.py:444-446) wraps everything in `except BaseException -> _close_agent_session(session, "failed")`, and `_resolve_preset_env` ran INSIDE the try. A resume from a shell missing a passthrough variable raises ConfigurationError before the agent is spawned and before anything changed - and the session is still written "failed", the workspace deleted, and load_resumable_session refuses it forever.

## Fix

Hoist `_resolve_preset_env` above the session claim and the try. The resume case stays "interrupted"/resumable; a fresh attempt leaves no dead record behind.

## Verification

Regression test resumes an interrupted session with a missing env var: on current HEAD the session flips to failed; with the fix it stays interrupted. Full create suite: 63 passed.

Fixes #4164

> Built by breken, your AI support engineer - breken.ai - this one's on us.

### DIFF
--- a/src/dstack/_internal/cli/services/presets/create.py
+++ b/src/dstack/_internal/cli/services/presets/create.py
@@ -411,12 +411,16 @@ def create_preset(
     allowed_fleets: Optional[tuple[str, ...]] = None,
     previous: Sequence[PresetSession] = (),
 ) -> PresetCreateResult:
+    # Resolve env before the session is claimed and before the try that marks
+    # it failed: a missing variable here means nothing was spawned and nothing
+    # changed, so a resumable session must stay interrupted - a failure this
+    # early is not a failed creation.
+    resolved_configuration = _resolve_preset_env(configuration)
     session = resume_session or create_preset_session(
         configuration,
         previous=tuple(session.preset_id for session in previous),
     )
     try:
-        resolved_configuration = _resolve_preset_env(configuration)
         # A creation session runs unattended for hours; an idling machine would
         # freeze the agent and the process supervising it alike.
         with prevent_idle_sleep():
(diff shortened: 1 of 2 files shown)


peterschmidt85:
@breken-ai we don't accept PRs if the implementation is not discussed first and the author that submits the PR is not actively using dstack.

Issue #2348, opened 2025-02-24
[Bug]: fabricmanager-dependent GPUs do not work in AWS, GCP, Azure, OCI


jvstme:
### Steps to reproduce

Try running this configuration on a fabricmanager-dependent instance (e.g. 8:A100 AWS, 8:H100 AWS, 8:A100 OCI, 8:H100 OCI) using `dstack`'s default OS images.
```yaml
type: task
commands:
- pip install numpy torch
- python -c 'import torch; print("CUDA available:", torch.cuda.is_available())'

resources:
  shm_size: 24GB
```

### Actual behaviour

```
/opt/conda/envs/workflow/lib/python3.9/site-packages/torch/cuda/__init__.py:129: UserWarning: CUDA initialization: Unexpected error from cudaGetDeviceCount(). Did you run some cuda functions before calling NumCudaDevices() that might have already set an error? Error 802: system not yet initialized (Triggered internally at /pytorch/c10/cuda/CUDAFunctions.cpp:109.)
  return torch._C._cuda_getDeviceCount() > 0
CUDA available: False
```

### Expected behaviour

 ```
CUDA available: True
```

### dstack version

0.18.42

### Server logs

```shell

```

### Additional information

```
> ssh proud-dragonfly-1-host sudo journalctl -u nvidia-fabricmanager
Warning: Permanently added '54.161.199.97' (ED25519) to the list of known hosts.
Feb 24 09:28:18 ip-172-31-92-157 systemd[1]: Starting NVIDIA fabric manager service...
Feb 24 09:28:28 ip-172-31-92-157 nv-fabricmanager[1507]: fabric manager NVIDIA GPU driver interface version 535.183.01 don't match with driver version 535.183.06. Please update with matching NVIDIA driver package.
Feb 24 09:28:31 ip-172-31-92-157 nv-fabricmanager[1507]: fabric manager NVIDIA GPU driver interface version 535.183.01 don't match with driver version 535.183.06. Please update with matching NVIDIA driver package.
Feb 24 09:28:31 ip-172-31-92-157 systemd[1]: nvidia-fabricmanager.service: Control process exited, code=exited, status=1/FAILURE
Feb 24 09:28:31 ip-172-31-92-157 systemd[1]: nvidia-fabricmanager.service: Failed with result 'exit-code'.
Feb 24 09:28:31 ip-172-31-92-157 systemd[1]: Failed to start NVIDIA fabric manager service.
```

jvstme:
[...]

Issue #2343, opened 2025-02-21
[Feature]: Support H100 in OCI


jvstme:
### Problem

Oracle Cloud provides on-demand bare metal 8xH100 instances ( `BM.GPU.H100.8`). Currently, these instances cannot be provisioned with dstack.

### Solution

Support the `BM.GPU.H100.8` shape.

### Known challenges

1. #2348
1. At the time of testing, the OCI API was reporting that `BM.GPU.H100.8` is out of capacity.
   ```
    #  BACKEND  REGION          INSTANCE       RESOURCES                                      SPOT  PRICE                
    1  oci      eu-frankfurt-1  BM.GPU.H100.8  256xCPU, 2048GB, 8xH100 (80GB), 50.0GB (disk)  no    $80    not available
   ```
   But in fact it was possible to start `BM.GPU.H100.8` via both the web console and the API. If this will be reproduced later, we'll need to research how accurate the OCI capacity reports are and reconsider using them.

### Would you like to help us implement this feature by sending a PR?

Yes

r4victor:
@jvstme, I suggest we close it since it's been a non-priority for a long time and re-open if it becomes relevant.

PR #4302, opened 2026-09-18
OCI: offer the H100, H200, L40S, A100-v2, B200, B300 and RTX PRO shapes

### LINKED ISSUE #4301: OCI: H100/H200/L40S/A100-v2/B200/B300/RTX PRO shapes are filtered out of offers
### Steps to reproduce

Configure the `oci` backend and run `dstack apply` with a run that requests an H100 (or H200, L40S, A100-v2, B200, B300, RTX PRO 6000) in a region and tenancy that has quota for it.

### Actual behaviour

No OCI offer is ever selected for those GPUs. `SUPPORTED_SHAPE_FAMILIES` in `src/dstack/_internal/core/backends/oci/compute.py` still lists only the shapes that existed when Bare Metal support landed (#1325): `BM.GPU3.`, `BM.GPU4.`, `BM.GPU.A10.` plus the VM families. Every newer shape that gpuhunt already catalogs (`BM.GPU.H100.8`, `BM.GPU.H200.8`, `BM.GPU.L40S.4`, `BM.GPU.A100-v2.8`, `BM.GPU.B200.8`) is filtered out by `_supported_instances`. #2343 asked for H100 and was closed as non-priority; its blocker (Fabric Manager, #2348) was fixed in #2355, so the images now install open kernel modules and Fabric Manager.

Separately, the gpuhunt OCI provider dropped OCI's AMD shapes and the newest NVIDIA shapes, and published GB200 under its display name (fix in https://github.com/dstackai/gpuhunt/pull/262).

### Expected behaviour

Runs that request these GPUs get OCI offers when the tenancy has quota. We run OCI GPU capacity (A10 and A100 today, H100-class shapes on the roadmap) and want to schedule it through dstack rather than by hand.

### dstack version

master (branch off `68e701d0`'s parent)

### Additional information

I have a change ready that adds the x86-64 NVIDIA families to the allowlist and records why AMD (no ROCm image) and Grace GB200/GB300 (arm64 hosts, x86 images only) stay out, plus a note on why `SUPPORTED_REGIONS` is fixed (the Marketplace listings carrying the dstack images are regional). PR to follow.

fede-kamel:
Closes #4301.

`SUPPORTED_SHAPE_FAMILIES` in the OCI backend still lists only the GPU shapes that existed when Bare Metal support landed (#1325): GPU3, GPU4, A10. Every newer NVIDIA shape from the gpuhunt catalog was filtered out by `_supported_instances`, so `dstack apply` could never select an OCI H100, H200, L40S, A100-v2, B200, B300 or RTX PRO 6000 even with quota. #2343 asked for H100; its blocker (Fabric Manager, #2348) was fixed in #2355 and the images now install open kernel modules and Fabric Manager.

This adds those families (all x86-64 hosts with NVIDIA GPUs, which is what the published `dstack-cuda-*` Marketplace images run). Deliberately still excluded, with the reason recorded next to the list and in tests: AMD Instinct shapes (no ROCm image is published, and `create_instance` always picks the CUDA image) and Grace GB200/GB300 shapes (arm64 hosts; the images are x86-64 only). It also documents why `SUPPORTED_REGIONS` is fixed: the Marketplace listings carrying the images are regional (verified: `dstack-cuda-0.14` present in us-chicago-1, absent in ca-toronto-1), so widening requires publishing images first.

The catalog side is in dstackai/gpuhunt (https://github.com/dstackai/gpuhunt/pull/262). Without it, this change already affects H100, H200, L40S and A100-v2, which the catalog carries today; B300 and RTX PRO 6000 appear once the gpuhunt change is merged and the hourly catalog republishes. Checked with dstack's pinned gpuhunt 0.1.30 that the new catalog rows convert through `catalog_item_to_offer` without error.

Tests: new `src/tests/_internal/core/backends/oci/test_compute.py` (13 supported / 10 unsupported shapes); OCI backend suite 43 passed, whole backends suite 473 passed; ruff clean.


jvstme:
@fede-kamel, thanks for both of the PRs, we appreciate the effort!

Have you had a chance to test each of these new shapes end-to-end against real hardware? If not, would you have a chance to do so?

The testing sequence I would use is:

- Provision each new shape through dstack and run a dstack dev environment.
- In the dev environment:
  - Verify that the GPU is detected and has the same specs as advertised in the dstack offer (use `nvidia-smi`).
  - Verify that the CPU and RAM specs match those advertised in the dstack offer.
  - Verify that the GPU works:
     ```
     pip install numpy torch
     python -c "import torch; print('CUDA available:', torch.cuda.is_available() and torch.cuda.get_device_name(0))"
     python -c "import torch; print('Random tensor:', torch.rand(10, device=torch.device('cuda')))"
     ```

We'd love to test them ourselves as well, just to make sure they'll work as expected for our users, but we don't currently have the right quotas.

fede-kamel:
Thanks for the review, and for being specific about what you wanted tested - it turned out to be the right question.

**Short answer: no, not for the shapes these PRs add.** I do not have quota for H100, H200, L40S, A100-v2, B200, B300 or RTX PRO, and they are not even offered as shapes in my tenancy, so there is nothing to draw down. The A100 shapes I do hold quota for were `OUT_OF_HOST_CAPACITY` in both availability domains while I was testing. I would rather say that plainly than imply coverage I do not have.

**What I could test, I did.** I ran your sequence on A10, on both a VM and a bare metal host in `us-phoenix-1`:

| | dstack offer | `VM.GPU.A10.1` | `BM.GPU.A10.4` |
|---|---|---|---|
| arch | `x86` | `x86_64` | `x86_64` |
| vCPUs | 30 / 128 | 30 | 128 |
| memory | 240 / 1024 GB | 235 GiB usable | 1007 GiB usable |
| GPU | `A10:24GB:1` / `:4` | 1 x A10, 23028 MiB | 4 x A10, 23028 MiB |
| torch | - | `device_count()==1`, CUDA tensor OK | `device_count()==4`, CUDA tensor OK |

Everything advertised matched the machine. The row I care about is `arch`: the Cost Estimator reports `processorType: "arm"` for these shapes and is wrong, and gpuhunt's `is_arm_cpu()` override is what keeps the catalog right - that is now confirmed on hardware rather than argued from a comment. Driver on the published image is `580.167.08`.

That covers the bare metal path the new families use, but it is not evidence about the new families themselves.

**So I have scoped this PR down.** Dropped `BM.GPU.B200.*`, `BM.GPU.B300.*` and `BM.GPU.RTXPRO.*` and retitled it. What is left is H100, H200, L40S and A100-v2 - the newer NVIDIA shapes the catalog already carries. R580 supports Blackwell, so I think those three would work, but "I think" is not what you asked for and they can come back when someone can boot one.

**Two bugs turned up before I could test anything**, both of which stopped the backend from being usable at all against my account. Filed separately since they are independent of the shape list:

* #4308 - profiles created by `oci session authenticate` cannot authenticate; clients are built without a `SecurityTokenSigner`, so `creds_valid()` returns `False` for a profile the OCI CLI accepts.
* #4309 - availability domains are resolved from the credentials' tenancy, so a `compartment_id` in another tenancy makes every `list_shapes` return 404 and no offers are ever produced.

Both have repro steps, unit tests, and 453 passing in the backends suite.

**On your quota problem:** I work at Oracle, on OCI. If it would help, I can try to get dstack access to the shapes you are missing so you can test these yourselves rather than take my word for anything - or run a sequence you specify on capacity here and post the raw output. Tell me which you would prefer, or if you would rather just hold these until you can verify independently, that is a reasonable call too.

peterschmidt85:
> On your quota problem: I work at Oracle, on OCI. If it would help, I can try to get dstack access to the shapes you are missing so you can test these yourselves rather than take my word for anything - or run a sequence you specify on capacity here and post the raw output. Tell me which you would prefer, or if you would rather just hold these until you can verify independently, that is a reasonable call too.

@fede-kamel Either or both options would be actually really cool:

1. See if you can get such a quota to test it on your side (if possible)

2. See if you could help us get such quota for our account so we can test it (if that can be arranged without involving sales from OCI)

fede-kamel:
Follow-up on the testing question, since I could not get hardware for these four.

I cross-checked what the catalog produces for them against Oracle's [Compute Shapes documentation](https://docs.oracle.com/en-us/iaas/Content/Compute/References/computeshapes.htm), which is an OCI source independent of the Cost Estimator the gpuhunt provider reads. Every field matches:

| shape | catalog | documentation |
|---|---|---|
| `BM.GPU.H100.8` | 224 vCPU, 2048 GB, 8 x H100 80 GB | 112 OCPU, 2048 GB, 640 GB total |
| `BM.GPU.H200.8` | 224 vCPU, 3072 GB, 8 x H200 141 GB | 112 OCPU, 3072 GB, 1128 GB total |
| `BM.GPU.L40S.4` | 224 vCPU, 1024 GB, 4 x L40S 48 GB | 112 OCPU, 1024 GB, 192 GB total |
| `BM.GPU.A100-v2.8` | 256 vCPU, 2048 GB, 8 x A100 80 GB | 128 OCPU, 2048 GB, 640 GB total |

The documentation also confirms all four shape names exactly as the allowlist matches them, which was the other thing that could silently be wrong - the Cost Estimator does decorate some names, `BM.GPU.GB200.4 (NVL72)` being the one that caught me out in the catalog PR.

This is not a substitute for running a job on one, and I am not presenting it as one. What it does rule out is the failure mode where the offer advertises specs that do not exist. Whether that is enough to land this, or whether you would rather wait for an actual run, is your call - I am happy either way, and the quota conversation is in your court too.

fede-kamel:
One more thing that I should have made explicit earlier, because it bears on how much the missing hardware test actually costs us here.

**The backend does not branch on GPU model.** Tracing what this PR changes:

* `_supported_instances()` is a prefix match on the shape name. No GPU logic at all.
* `create_instance()` has exactly one GPU-dependent branch - `requires_nvidia_proprietary_kernel_modules(gpu_name)`, which picks between the open-kernel-modules image and the proprietary one.
* Everything after that - cloud-init, security group, subnet, `launch_instance` - takes `shape` as an opaque string.

And that one branch resolves identically for the shape I tested and for the four this PR adds:

```
A10  -> False      H100 -> False
A100 -> False      H200 -> False
                   L40S -> False
V100, P100, P40, M60 -> True   (pre-Turing, different image)
```

So the `BM.GPU.A10.4` run went through the same code path, the same `dstack-cuda-*` image and the same launch call that a `BM.GPU.H100.8` run would. The interfaces are identical; the shape name and the catalog metadata are the only things that vary.

That narrows what is genuinely untested for these four shapes to two things, and both have now been checked against Oracle's published documentation:

1. **the shape name string** - confirmed, all four appear exactly as the allowlist matches them
2. **the catalog metadata** - confirmed, vCPU / RAM / per-GPU memory all match

What remains unverified is whether OCI's Hopper hosts have some host-level quirk the A10 hosts do not. That is a real residual, and it is why I am not claiming this is equivalent to a run. But it is a much smaller gap than "untested", and it does not grow with each shape added, since they all share one path.

The same reasoning applies to B200/B300/RTX PRO, which also resolve to `False` on that branch. I dropped them from this PR anyway, since Blackwell driver support is newer than Hopper's and I would rather be conservative there. Happy to restore them if you would rather the argument be applied consistently.

